PR #8, opened 2019-03-22
Merge apex_launchtest_ros functionality into launch_testing_ros

hidmic:
This pull request merges functionality from `apex_launchtest_ros` package into `launch_testing_ros`. Some renaming and style corrections were applied as well.

Connected to https://github.com/ros2/launch/issues/208.


hidmic:
CI (full run):

* Linux [Build Status](https://ci.ros2.org/job/ci_linux/6507/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2856/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5331/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6305/)

hidmic:
I'll rebase after fixing this on Windows (which I suspect wasn't supported to begin with).

hidmic:
Alright, all tests are passing but for a few unrelated, known failures!

hidmic:
Alright, all tests are passing but for a few unrelated, known failures!

launch_testing_ros/launch_testing_ros/tools/output.py
```diff
@@ -0,0 +1,63 @@
+# Copyright 2018 Open Source Robotics Foundation, Inc.
```
ivanpauno: Is it not 2019?
Or this came from another file?

ivanpauno:
Also, one of the tests is failing to me locally. So, check having green CI again.

launch_testing_ros/launch_testing_ros/tools/output.py
```diff
@@ -0,0 +1,63 @@
...
+    starting with the process ID (`b'pid'`) and return code (`b'rc'`) will be ignored.
+    :param filtered_patterns: A list of byte strings representing regexes that will cause
+    output lines to be ignored if they match one of the regexes.
+    :param filtered_rmw_implementation: RMW implementation for which the output will be
+    ignored in addition to the `filtered_prefixes`/`filtered_patterns`.
+    """
+    from launch_testing.tools.output import basic_output_filter as base_output_filter
+    from launch_testing.tools.output import get_default_filtered_prefixes
+    from launch_testing.tools.output import get_default_filtered_patterns
+    filtered_prefixes = filtered_prefixes or get_default_filtered_prefixes()
```
jacobperron: Is the empty list a valid for ignoring nothing? If so, I don't think this will work as it will still prefer the default.
Consider rewriting as:

```python
filtered_prefixes = filtered_prefixes if filtered_prefixes is not None else get_default_filtered_prefixes()
```

same for `filtered_patterns`.

launch_testing_ros/launch_testing_ros/tools/output.py
hidmic: It did came from somewhere else but 2019 is still correct.

PR #11, opened 2019-03-28
use yaml.safe_load instead of deprecated yaml.load

mikaelarguedas:
**DO NOT MERGE**

This PR removes deprecation warning similarly to colcon/colcon-metadata#14.

With this change the tests are currently failing due to the use of python tuples converted to yaml in the tests.
2 options to fix it:
- use the unsafe loader: As this is only test code and the fixtures are controlled, using the unsaf loader does not pose security concerns.
- use the safe loader: this allows to use the same loader throughout the codebase, but I'm not familiar with the code here to say if the change to use only standard YAML tags is trivial or not.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_node.py
@@ -160,7 +160,7 @@ class TestNode(unittest.TestCase):
         expanded_parameter_files = node_action._Node__expanded_parameter_files
         assert len(expanded_parameter_files) == 1
         with open(expanded_parameter_files[0], 'r') as h:
-            expanded_parameters_dict = yaml.load(h)
+            expanded_parameters_dict = yaml.safe_load(h)
             assert expanded_parameters_dict == {
                 '/my_ns': {
                     'my_node': {


dirk-thomas:
We need to resolve this deprecation warning. Therefore removing the backlog label.

mikaelarguedas:
@dirk-thomas Do you have a preference on the approach. If option 1 I can update the PR accordingly. If option 2, I likely won't have time to look into it and will discard this in favor of a PR implementing the correct change

dirk-thomas:
I don't have a preference how this gets resolved. I will leave it up to someone else to express a preference. Maybe @hidmic?

hidmic:
@mikaelarguedas option 2, hands down. Would you mind providing us with links to failing tests in CI?

mikaelarguedas:
This was a year ago so failure modes may have changed since.
This change never got ran on CI so no links to the fail tests exist. Applying this change on top of master should be enough to reproduce the failures.

IIRC some tests failed because some tuples with custom data structures were constructed and loaded during the tests.

jacobperron:
I believe the warnings were resolved in #143.

mikaelarguedas:
So we can merge this then and stop using the unsafe loader?

PR #15, opened 2019-04-16
Unit tests for composition actions

sloretz:
This is the start of unit tests for composition. It depends on #8, so if you'd like to run locally make a branch that merges this branch and #8 .

Issues to resolve:

1. How to make `launch_test test_launch_ros/test/rostest/composition.test.py` part of the test suite in a pure python package?
2. Add log_level to `ComposableNodeDescription`
3. `LaunchComposableNodes` returns actions to be executed when the container emits the event `OnProcessStart`, but that means they won't trigger if the container process is already started. At the moment composable nodes launching only works via the `composable_node_descriptions` on the container action
4. How to handle failures to launch nodes? Should emit an event so other launch actions can do something if a node fails to launch
5. How to unload nodes? Should emit an event with the node unique_id so it's possible to make an unload action


hidmic:
@sloretz #8 just got merged, mind to check if this is still working and put it on review (or not)?

sloretz:
@hidmic The test passes running with:

```
launch_test src/ros2/launch_ros/test_launch_ros/test/rostest/composition.test.py
```

This isn't ready to be merged as the test is not run with `colcon test --packages-select test_launch_ros`. I'm not sure how to make a launch test part of the test suite for a pure python package.  Is there a pytest plugin for launch-test? If not, a quick alternative would be a test file that called `launch_test` using `subprocess`.

hidmic:
That's a good point, we don't have a tool for that atm.

hidmic:
@sloretz with https://github.com/ros2/launch/pull/236 now we can call `launch_test` from Python code to some extent.

hidmic:
@sloretz once https://github.com/ros2/launch/pull/312 goes in, we can have launch tests in pure Python packages.

jacobperron:
A few issues that may be relevant:
- https://github.com/ros2/launch_ros/issues/107
- https://github.com/ros2/launch_ros/issues/108

wjwwood:
@sloretz I rebased this and I'm going to try and resurrect it in general, hope you don't mind.

hidmic:
@wjwwood friendly ping

wjwwood:
I do not remember what I was going to do to "restructure" these tests...

I'll re-review it, but I'm afraid after dropping the ball on this I've lost context.

wjwwood:
I'm guessing it needs to be updated to use the "launch tests from python packages" updates and/or changed to not use `ready_fn` in favor of pytest.

PR #16, opened 2019-04-17
Move OnProcessStart from LoadComposableNodes to ComposableNodeContainer

sloretz:
Currently `LoadComposableNodes.execute()` returns an action that loads the nodes when the container process starts. However, if the container process is already started then these loads will never be loaded because the event has already been triggered. This changes `LaunchComposableNodes.execute` to launch the nodes right away, and makes  `ComposableNodeContainer` return an action that triggers `LaunchComposableNodes` on the event `OnProcessStart`.

The result is `LoadComposableNodes` can be triggered by some event long after the container process starts.

I think this can be merged as is, though #15 has a test for this case. To run that you'll need to make local a branch with this + #8 + #15.

### DIFF
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -18,6 +18,8 @@ from typing import List
 from typing import Optional
 
 from launch.action import Action
+from launch.actions import RegisterEventHandler
+from launch.event_handlers.on_process_start import OnProcessStart
 from launch.launch_context import LaunchContext
 from launch.some_substitutions_type import SomeSubstitutionsType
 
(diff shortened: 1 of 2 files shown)


sloretz:
CI (testing launch_ros and test_launch_ros)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6719/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3033/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5522/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6549/)

sloretz:
@hidmic Documented in fd4ece6

CI linux (just one platform for linter tests since change is just in a docstring)
[Build Status](https://ci.ros2.org/job/ci_linux/6729/)

sloretz:
I'll assume @hidmic's approval extends to the documentation added in fd4ece6 and merge this

PR #23, opened 2019-05-10
Add frontend parsing methods for Node, ExecutableInPackage and FindPackage substitution

ivanpauno:
Add parsing method for Node and ExecutableInPackage.
Add test example of launching a node from xml.

Blocked by https://github.com/ros2/launch/pull/226
Blocked by https://github.com/ros2/launch_ros/pull/33


launch_ros/launch_ros/actions/node.py
```diff
@@ -178,6 +185,108 @@ def __init__(
...
+                    parser.parse_substitution(remap.get_attr('to'))
+                ) for remap in remappings
+            ]
+        parameters = entity.get_attr('param', types='list[Entity]', optional=True)
+        if parameters is not None:
+            kwargs['parameters'] = Node.parse_nested_parameters(parameters)
+        if_cond = entity.get_attr('if_cond', optional=True)
+        unless_cond = entity.get_attr('unless_cond', optional=True)
+        if if_cond is not None and unless_cond is not None:
+            raise RuntimeError("if and unless conditions can't be usede simultaneously")
```
hidmic: @ivanpauno consider:
```suggestion
            raise RuntimeError("if and unless conditions can't be used simultaneously")
```

launch_ros/launch_ros/substitutions/executable_in_package.py
```diff
@@ -48,6 +52,15 @@ def __init__(self, executable: SomeSubstitutionsType, package: SomeSubstitutions
...
         self.__package = normalize_to_list_of_substitutions(package)
 
+    @staticmethod
+    def parse(data: Iterable[SomeSubstitutionsType]):
+        """Parse a ExecutableInPackage substitution."""
+        if not data or len(data) != 2:
+            raise AttributeError('exec-in-package substitution expects 2 arguments')
+        executable = data[0]
+        package = data[1]
+        return ExecutableInPackage(executable, package)
```
hidmic: @ivanpauno nit: you can also do: 
```py
executable, package = data
```
or even

```py
return ExecutableInPackage(*data)
```

test_launch_ros/test/test_launch_ros/xml_frontend/node.xml
```diff
@@ -0,0 +1,12 @@
+<launch>
+    <node package="demo_nodes_py" executable="talker_qos" output="screen" name="my_node" namespace="my_ns" args="--number_of_cycles 1">
```
hidmic: @ivanpauno hmm I'm not sure if we should be using demo nodes for testing of core packages. Is that the case already? Otherwise, consider rolling out a dummy one just for this purpose.

test_launch_ros/test/test_launch_ros/xml_frontend/test_node_xml.py
```diff
@@ -0,0 +1,34 @@
...
+from pathlib import Path
+
+from launch import LaunchService
+
+from launch_frontend import Parser
+
+
+def test_node_xml_frontend():
+    """Parse node xml example."""
+    root_entity, parser = Parser.load(str(Path(__file__).parent / 'node.xml'))
```
hidmic: @ivanpauno here and below, can't we collapse both tests into one and just parameterize it on the path to the launch description file?

test_launch_ros/test/test_launch_ros/xml_frontend/node.xml
ivanpauno: We're already doing that.
https://github.com/ros2/launch_ros/blob/67d0c886e0245b5b4d7bd0e58564f07bdc13075f/test_launch_ros/package.xml#L1-L23
https://github.com/ros2/launch_ros/blob/67d0c886e0245b5b4d7bd0e58564f07bdc13075f/test_launch_ros/test/test_launch_ros/actions/test_node.py#L43-L52

hidmic: Hmm, ok. I wonder still. But OK to keep it as is now.

ivanpauno:
Rebased again with master

ivanpauno:
@hidmic https://github.com/ros2/launch_ros/pull/23/commits/218d879acf6acd6031895dd00710285924d96b31 and https://github.com/ros2/launch/pull/226/commits/cd4ff85e2886938c834400ab241f657030b01a45 handles parameters correctly after #31.
If you're ok with that change, this is ready for be reviewed by someone else.

ivanpauno:
> @hidmic [218d879](https://github.com/ros2/launch_ros/commit/218d879acf6acd6031895dd00710285924d96b31) and [ros2/launch@cd4ff85](https://github.com/ros2/launch/commit/cd4ff85e2886938c834400ab241f657030b01a45) handles parameters correctly after #31.
> If you're ok with that change, this is ready for be reviewed by someone else.

Reverted both commits after discussing with @hidmic.
See https://github.com/ros2/launch_ros/pull/23/commits/86bd241c5fde46ecddac34523387b7a25d121217 for checking how it looks now.

launch_ros/launch_ros/actions/node.py
```diff
@@ -177,6 +183,78 @@ def __init__(
...
+        def get_nested_dictionary_from_nested_key_value_pairs(params):
+            """Convert nested params in a nested dictionary."""
+            # TODO(ivanpauno): If our schema checking is not powerfull enough
+            # this could easily end in an infinite loop.
+            # In that case, we should do some extra formal checking before processing.
+            param_dict = {}
+            for param in params:
+                name = param.get_attr('name')
+                value = param.get_attr('value', types='guess', optional=True)
+                if value is not None:
```
wjwwood: This logic is a bit confusing, but I don't have a suggestion on how to make it clearer off-hand. Like if 'guess` returns `None` then implicitly it is a list of entities?

test_launch_ros/test/test_launch_ros/launch_frontend/test_node_frontend.py
```diff
@@ -0,0 +1,112 @@
...
+@pytest.mark.parametrize('file', (xml_file, yaml_file))
+def test_node_frontend(file):
+    """Parse node xml example."""
+    root_entity, parser = Parser.load(io.StringIO(file))
+    ld = parser.parse_description(root_entity)
+    ls = LaunchService()
+    ls.include_launch_description(ld)
+    assert(0 == ls.run())
+    # TODO(ivanpauno): Test this in a cleaner way.
+    lc = ls._LaunchService__context
```
wjwwood: I think it would be ok to add a method to the `LaunchService` so that you can access the context.

launch_ros/launch_ros/actions/node.py
ivanpauno: Yes this is a little confusing.
You have two mutually exclusive options:
- parameter specifying `name` and `value`.
- parameter specifying `name`, and having some other nested parameter entities.

Only having a `name` attribute and nothing more is an error (it's not clear in the code above, I will update).

The first options looks in an xml like:
```xml
<param name="asd" value="1"/>
```

and the second as:
```xml
<param name="asd">
    <param name="bsd" value="1"/>
</param>
```

The same rule is applied recursively for nested parameter entities.

It's also a third option, that can only be applied in a "root" param tag. It looks like:
```xml
<param from="path/to/file"/>
```
This option can not have other entities as children.

I haven't added tests for that option. I will do that before merging.

test_launch_ros/test/test_launch_ros/launch_frontend/test_node_frontend.py
```diff
@@ -0,0 +1,116 @@
...
+
+from launch import LaunchService
+from launch.launch_frontend import Parser
+
+from launch_ros.utilities import evaluate_parameters
+
+import pytest
+
+# Scaping the quote is needed in 'a_string' launch configuration, becuase of how the parser works.
+# TODO(ivanpauno): Check if it's possible to avoid that.
```
hidmic: @ivanpauno is all this quoting for `'a_string'` due to substitution grammar parsing? Is that bad for YAML markup?

test_launch_ros/test/test_launch_ros/launch_frontend/test_node_frontend.py
```diff
@@ -0,0 +1,116 @@
...
+def test_node_frontend(file):
+    """Parse node xml example."""
+    root_entity, parser = Parser.load(io.StringIO(file))
+    ld = parser.parse_description(root_entity)
+    ls = LaunchService()
+    ls.include_launch_description(ld)
+    assert(0 == ls.run())
+    lc = ls.context
+    evaluated_parameters = evaluate_parameters(
+        lc,
```
hidmic: @ivanpauno nit: using `ls.context` instead of `lc` is not that much longer and makes code much more readable.

test_launch_ros/test/test_launch_ros/launch_frontend/test_node_frontend.py
ivanpauno: Yeah, the substitution grammar needs the quotes escaped. In YAML markup you have the same problem, but it's not particularly "bad" for it.
The yaml example is in the same file (below the xml example).

ivanpauno:
@hidmic I rebased, because I needed https://github.com/ros2/launch_ros/pull/33.
The new commits since your last review are: https://github.com/ros2/launch_ros/pull/23/commits/a3ca07f38cd1cfd6ea8d1f6e8c7e84bb52da0f9e, https://github.com/ros2/launch_ros/pull/23/commits/07f162dba5e8d58323e7905984d82a6adcd3537b, https://github.com/ros2/launch_ros/pull/23/commits/c1c1d6accc481b469009f9160c4120700a7ec96b, https://github.com/ros2/launch_ros/pull/23/commits/3a95e38085c8272aceeda19109859d90e4afa5ce.

launch_ros/launch_ros/actions/node.py
```diff
@@ -172,6 +175,91 @@ def __init__(
...
+        # See parse method of `ExecuteProcess`
+        _, kwargs = super().parse(entity, parser, 'args')
+        kwargs['arguments'] = kwargs['args']
+        del kwargs['args']
+        kwargs['node_name'] = kwargs['name']
+        del kwargs['name']
+        package = parser.parse_substitution(entity.get_attr('package'))
+        kwargs['package'] = package
+        executable = parser.parse_substitution(entity.get_attr('executable'))
+        kwargs['node_executable'] = executable
```
hidmic: @ivanpauno just curious, why the two step assignment?

ivanpauno: Just a bad habit.

PR #28, opened 2019-05-17
fix example test logic

dirk-thomas:
Fixes ros2/build_cop#197.

Without the patch the publisher publishes a single message and waits for the subscriber to receive that message. Since the matching between both might not have happened the `assertWaitFor` times out after 10s and calls `assert` failing the test.

This patch relies on ros2/launch#243 and publishes 10 messages - one every second - until the listener received it.

Before: [Build Status](https://ci.ros2.org/job/ci_linux/7143/)
After: [Build Status](https://ci.ros2.org/job/ci_linux/7145/)

### DIFF
--- a/launch_testing_ros/examples/talker_listener.test.py
+++ b/launch_testing_ros/examples/talker_listener.test.py
@@ -160,8 +160,7 @@ class TestTalkerListenerLink(unittest.TestCase):
             )
             if success:
                 break
-        else:
-            assert False, 'Waiting for output timed out'
+        assert success, 'Waiting for output timed out'
 
     def test_fuzzy_data(self, listener):
         # This test shows how to insert a node in between the talker and the listener to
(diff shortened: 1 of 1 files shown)


launch_testing_ros/examples/talker_listener.test.py
```diff
@@ -146,17 +146,22 @@ def test_listener_receives(self, listener):
...
-            self.proc_output.assertWaitFor(
+            success = self.proc_output.waitFor(
                 expected_output=msg.data,
-                process=listener
+                process=listener,
+                timeout=1.0,
             )
+            if success:
+                break
+        else:
```
jacobperron: nit: the loop-else looks odd and might confuse future readers. Maybe just `assert success` instead?

dirk-thomas: Good point: done in d1724b73db122f0ebdf6dcce198c04c528e0bab2.

Issue #30, opened 2019-05-20
launch file should interpret LaunchConfiguration values as yaml


Karsten1987:
## Bug report

**Required Info:**

- Operating System:
  - OSX, but should apply to all
- Installation type:
  - source
- Version or commit hash:
  - HEAD

#### Steps to reproduce issue

It is currently not trivial to set a boolean parameter via a launch file.
```
    return LaunchDescription([
        DeclareLaunchArgument(
            'use_sim_time',
            default_value='True',
            description='Use simulation (Gazebo) clock if true'),

        Node(
            package='robot_state_publisher',
            node_executable='robot_state_publisher',
            node_name='robot_state_publisher',
            output='screen',
            parameters=[{'use_sim_time': LaunchConfiguration('use_sim_time')}],
            arguments=[urdf])
    ])
```

#### Expected behavior
Robot state publisher sets the boolean parameter `use_sim_time` to true

#### Actual behavior
Robot state publisher complains that parameter `use_sim_time` is a string 
 
#### Additional information

A workaround for this to work is to use `IfConditions` and `UnlessConditions`.
```
    return LaunchDescription([
        DeclareLaunchArgument(
            'use_sim_time',
            default_value='True',
            description='Use simulation (Gazebo) clock if true'),

        Node(
            package='robot_state_publisher',
            node_executable='robot_state_publisher',
            node_name='robot_state_publisher',
            output='screen',
            parameters=[{'use_sim_time': True}],
            arguments=[urdf],
            condition=IfCondition(LaunchConfiguration('use_sim_time'))),

        Node(
            package='robot_state_publisher',
            node_executable='robot_state_publisher',
            node_name='robot_state_publisher',
            output='screen',
            parameters=[{'use_sim_time': False}],
            arguments=[urdf],
            condition=UnlessCondition(LaunchConfiguration('use_sim_time')))
    ])
```

PR #31, opened 2019-05-20
Parameter values get evaluated by yaml rules

sloretz:
Resolves ros2/launch_ros#30

This PR evaluates parameter values as `yaml` would.

I'm not set up to run mypy at the moment, so I'm not confident the type annotations are correct.

### DIFF
--- a/launch_ros/launch_ros/utilities/evaluate_parameters.py
+++ b/launch_ros/launch_ros/utilities/evaluate_parameters.py
@@ -43,12 +43,12 @@ def evaluate_parameter_dict(
 ) -> Dict[str, EvaluatedParameterValue]:
     if not isinstance(parameters, Mapping):
         raise TypeError('expected dict')
-    output_dict = {}  # type: Dict[str, EvaluatedParameterValue]
+    output_dict: Dict[str, EvaluatedParameterValue] = {}
     for name, value in parameters.items():
         if not isinstance(name, tuple):
             raise TypeError('Expecting tuple of substitutions got {}'.format(repr(name)))
-        evaluated_name = perform_substitutions(context, list(name))  # type: str
-        evaluated_value = None  # type: Optional[EvaluatedParameterValue]
+        evaluated_name: str = perform_substitutions(context, list(name))
+        evaluated_value: Optional[EvaluatedParameterValue] = None
 
         if isinstance(value, tuple) and len(value):
             if isinstance(value[0], Substitution):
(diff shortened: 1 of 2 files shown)


launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -52,12 +54,29 @@ def evaluate_parameter_dict(
...
                 # Value is an array of a list of substitutions
                 output_subvalue = []  # List[str]
                 for subvalue in value:
-                    output_subvalue.append(perform_substitutions(context, list(subvalue)))
+                    value = perform_substitutions(context, list(subvalue))
+                    output_subvalue.append(value)
                     evaluated_value = tuple(output_subvalue)
+                # All values in a list must have the same type.
+                # If they don't then assume it is a list of strings
+                yaml_evaluated_value = []  # Union[List[str], List[int], List[float], List[bool]]
```
wjwwood: nitpick: now that we don't have to support python3.5 we can use the normal syntax for this:

```python
yaml_evaluated_value: Union[List[str], List[int], List[float], List[bool]] = []
```

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -52,12 +54,29 @@ def evaluate_parameter_dict(
...
+                type_ = None
+                dissimilar_types = False
+                for subvalue in evaluated_value:
+                    yaml_subvalue = yaml.safe_load(subvalue)
+                    subtype = type(yaml_subvalue)
+                    if type_ is not None and type_ != subtype:
```
wjwwood: When is `type_` `None`?

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -52,12 +54,29 @@ def evaluate_parameter_dict(
...
+                        dissimilar_types = True
+                        break
+                    yaml_evaluated_value.append(yaml_subvalue)
+                    type_ = subtype
+                if not dissimilar_types:
```
wjwwood: What should happen (if anything) in the `else` case for if statement?

launch_ros/launch_ros/utilities/evaluate_parameters.py
sloretz: Replaced comments with new syntax in this function in 379e43f

launch_ros/launch_ros/utilities/evaluate_parameters.py
sloretz: `type_` is `None` when looking at the first element of the list. A better name would be `last_subtype`

sloretz: `type_` -> `last_subtype` in 032ae59

launch_ros/launch_ros/utilities/evaluate_parameters.py
sloretz: Nothing should happen. If the values in a sequence after yaml'ing are not all the same type then `evaluated_value` stays a sequence of strings.

sloretz:
CI (testing launch_ros and test_launch_ros)
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7184/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3395/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5891/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7021/)

Karsten1987:
Oh, too late for the party. But anyway, this works great! :)

One can now set boolean parameters when launching a node:
```
        Node(
            package='robot_state_publisher',
            node_executable='robot_state_publisher',
            node_name='robot_state_publisher',
            output='screen',
            parameters=[{'use_sim_time': LaunchConfiguration('use_sim_time')}],
            arguments=[urdf])
```

PR #33, opened 2019-06-03
Restrict yaml loading in evaluate_parameters

ivanpauno:
Follow up of https://github.com/ros2/launch_ros/pull/31.

The idea of this PR is to restrict what conversions from yaml are allowed, and what conversions are precluded. Also, substitutions that are later loaded as a list (from yaml), are converted to tuples (as it was done in the other cases).

I didn't want to raise an error in the conversions that aren't allowed, but I think it would be a good idea.


sloretz:
Mind adding a test here to show what this is correcting?

https://github.com/ros2/launch_ros/blob/master/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py

ivanpauno:
> Mind adding a test here to show what this is correcting?
> 
> https://github.com/ros2/launch_ros/blob/master/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py

I will do it tomorrow.

hidmic:
I second @sloretz request. Just for the record though, code seems to be precluding nested parameter arrays and verifying that arrays evaluated from substitutions have a single element type.

wjwwood:
Same, a test would be nice, it would also help convey what's being changed here.

ivanpauno:
Sorry for the unclear PR yesterday. I updated the PR with a clearer description.
Tests were added in https://github.com/ros2/launch_ros/pull/33/commits/1456aca92b837a4970e986abd2651f6316024404.

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -54,7 +54,23 @@ def evaluate_parameter_dict(
...
+                    last_subtype = None
+                    dissimilar_types = False
+                    for subvalue in yaml_evaluated_value:
+                        subtype = type(subvalue)
+                        if last_subtype is not None and last_subtype != subtype:
+                            dissimilar_types = True
+                            break
+                        last_subtype = subtype
+                    if not dissimilar_types and subtype in (bool, int, float, str):
+                        evaluated_value = tuple(yaml_evaluated_value)
```
wjwwood: You mentioned that you were implicitly making this a tuple as it was in other places, but does anyone know if this is the right thing to do? I guess it might, but why can you not just do `evaluated_value = yaml_evaluated_value` or `evaluated_value = copy.copy(yaml_evaluated_value)`?

wjwwood: I guess you're copying the code from the `elif isinstance(value[0], Sequence):` block.

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
```diff
@@ -230,3 +230,55 @@ def test_mixed_path_dicts():
...
+    norm = normalize_parameters(orig)
+    expected = ({'foo': 1, 'fiz': '[(2, 4, 5), (2, 4, 5), (2, 4, 5)]'},)
+    assert evaluate_parameters(LaunchContext(), norm) == expected
+
+
+def test_unallowed_yaml_types_in_substitutions():
+    orig = [{'foo': 1, 'fiz': TextSubstitution(text="{'asd': 3}")}]
+    norm = normalize_parameters(orig)
+    expected = ({'foo': 1, 'fiz': "{'asd': 3}"},)
+    assert evaluate_parameters(LaunchContext(), norm) == expected
```
wjwwood: I guess this is the kind of input you'd like to have an exception for but didn't add?

ivanpauno: I would like to raise an error in all this cases:
https://github.com/ros2/launch_ros/blob/1456aca92b837a4970e986abd2651f6316024404/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py#L251-L284

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -54,28 +66,38 @@ def evaluate_parameter_dict(
...
-                    if last_subtype is not None and last_subtype != subtype:
-                        dissimilar_types = True
-                        break
-                    yaml_evaluated_value.append(yaml_subvalue)
-                    last_subtype = subtype
-                if not dissimilar_types:
+                yaml_evaluated_value: Union[List[str], List[int], List[float], List[bool]] = [
+                    yaml.safe_load(item) for item in evaluated_value
+                ]
+                if check_sequence_type_is_allowed(yaml_evaluated_value):
```
ivanpauno:
```suggestion
                if not check_sequence_type_is_allowed(yaml_evaluated_value):
                    raise TypeError(
                        'Expected a non-empty sequence, with items of uniform type. '
                        'Allowed sequence item types are bool, int, float, str.'
                    )
                evaluated_value = tuple(yaml_evaluated_value)
```

That change will also raise an error in:
https://github.com/ros2/launch_ros/blob/4efbfe5e0933a3747954f78b1c765348711375e9/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py#L214-L239
https://github.com/ros2/launch_ros/blob/4efbfe5e0933a3747954f78b1c765348711375e9/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py#L283-L293

I can add complicated logic for allowing interpreting some dissimilar arrays as arrays of strings, and some others not.
But, what's that useful for?

I think that in all the cases in `test_dictionary_with_dissimilar_array`, the user could write easily the parameters in other wat for getting an array of strings (if they want that). I wouldn't be so flexible as we are being now.

If we want that change, I could later simplify `normalize_parameters` logic in a follow-up PR.

launch_ros/launch_ros/utilities/evaluate_parameters.py
ivanpauno: I think that doing `copy.copy` or converting it to a `tuple` is not needed.

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -54,7 +54,23 @@ def evaluate_parameter_dict(
             if isinstance(value[0], Substitution):
                 # Value is a list of substitutions, so perform them to make a string
                 evaluated_value = perform_substitutions(context, list(value))
-                evaluated_value = yaml.safe_load(evaluated_value)
+                yaml_evaluated_value = yaml.safe_load(evaluated_value)
+                evaluated_type = type(yaml_evaluated_value)
+                if evaluated_type in (bool, int, float, str):
+                    evaluated_value = yaml_evaluated_value
+                # TODO(ivanpauno): Maybe, add better checking of other sequence types.
```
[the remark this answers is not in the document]
ivanpauno: Used collections.Sequence in https://github.com/ros2/launch_ros/commit/3f328d7d0c3c6a3cafc526cd9e0841867880c065.

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -54,7 +54,23 @@ def evaluate_parameter_dict(
...
+                if isinstance(yaml_evaluated_value, list):
+                    # If it is a list with dissimilar types, let the string without converting.
```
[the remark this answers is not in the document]
ivanpauno: Corrected in https://github.com/ros2/launch_ros/commit/3f328d7d0c3c6a3cafc526cd9e0841867880c065

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
ivanpauno: Started raising some exceptions in https://github.com/ros2/launch_ros/commit/4efbfe5e0933a3747954f78b1c765348711375e9.

launch_ros/launch_ros/utilities/evaluate_parameters.py
sloretz: When I wrote this I made it return tuples because I wanted the output of `normalize_parameters` to be an immutable thing on the composable node description. However, that didn't work out because they include dictionaries, and there isn't a built in immutable version of that. I never undid evaluating sequences as `tuple`. It probably could be anything as long as the output type is consistent.

wjwwood: Any is fine by me, I just wanted to bring it up for discussion.

launch_ros/launch_ros/utilities/evaluate_parameters.py
wjwwood: That sounds fine to me. It's better to be strict and relax it later if someone makes a case for it, than to make it stricter later.

wjwwood: To be clear, +1 to ratcheting up the exception more if all the cases don't throw yet, whether or not that's just this change or something additional.

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
```diff
@@ -212,31 +212,35 @@ def test_dictionary_with_bytes():
 
 
 def test_dictionary_with_dissimilar_array():
-    orig = [{'foo': 1, 'fiz': [True, 2.0, 3]}]
-    norm = normalize_parameters(orig)
-    expected = ({'foo': 1, 'fiz': ('True', '2.0', '3')},)
-    # assert evaluate_parameters(LaunchContext(), norm) == expected
+    with pytest.raises(TypeError) as exc:
+        orig = [{'foo': 1, 'fiz': [True, 2.0, 3]}]
```
sloretz: What happens if `orig` is `[{'foo': 1, 'fiz': ['True', '2.0', '3']}]` ? Does that get correcly interpretted as a list of strings, or does it raise `TypeError`?

ivanpauno: It raises `TypeError`. For avoiding it, the user should write `["'True'", "'2.0'", "'3'"]`.
I added test cases in: https://github.com/ros2/launch_ros/pull/33/commits/23421c96d9760338e7e68eee9bb37b8f4e4ea826.

I don't know if this is the best way to go or not. But I think that behavior is consistent with the other test cases I have added (in https://github.com/ros2/launch_ros/pull/33/commits/23421c96d9760338e7e68eee9bb37b8f4e4ea826).

ivanpauno:
@sloretz do you think this is ready? (after having green CI)

test_launch_ros/test/test_launch_ros/test_normalize_parameters.py
hidmic: How so? Are plain dictionary values evaluated as YAML too? [It doesn't look like it](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/utilities/evaluate_parameters.py#L88).

ivanpauno: Strings are converted to `TextSubstitution` in `normalize_parameters` before.
I don't have a better solution in mind, for avoiding the extra quotes.

ivanpauno: > How so? Are plain dictionary values evaluated as YAML too? [It doesn't look like it](https://github.com/ros2/launch_ros/blob/master/launch_ros/launch_ros/utilities/evaluate_parameters.py#L88).

Thinking about this again, I stopped converting raw strings (`str`) and `TextSubstitution` like yaml.
The second one is needed, as `str` are converted to `TextSubstitution` in `normalize_parameters`.

I'm happy with how the test looks now, but I have to recognize that both `normalize_parameters` and `evaluate_parameters` need a refactor. But I think that's ok for a follow-up.

ivanpauno:
Refactored in https://github.com/ros2/launch_ros/pull/33/commits/db6a00f4cccf0f04bac1a36e60908138d08eb43e after being lightened by @hidmic.
In https://github.com/ros2/launch_ros/pull/33/commits/821395a5f3264d55d87eb407d65e917146de23dc, I have corrected an old error I found in `normalize_parameters` while reading the code. Added test case for it.

launch_ros/launch_ros/utilities/normalize_parameters.py
```diff
@@ -74,7 +76,7 @@ def _normalize_parameter_array_value(value: SomeParameterValue) -> ParameterValu
         # all were floats or ints, so return floats
         make_mypy_happy_float = cast(List[Union[int, float]], value)
         return tuple(float(e) for e in make_mypy_happy_float)
-    elif Substitution in has_types and has_types.issubset({str, Substitution, tuple}):
+    elif Substitution in has_types and has_types.issubset({str, Substitution}):
```
hidmic: @ivanpauno why was `tuple` removed?

ivanpauno: It doesn't have sense, `normalize_to_list_of_substitutions` can't handle that and fails:
https://github.com/ros2/launch_ros/blob/78469a9122ad4ca8416bf3a6c34acecbfafb982f/launch_ros/launch_ros/utilities/normalize_parameters.py#L77-L79
I added a test case for that: https://github.com/ros2/launch_ros/blob/821395a5f3264d55d87eb407d65e917146de23dc/test_launch_ros/test/test_launch_ros/test_normalize_parameters.py#L95-L101.
Without removing `tuple`, it was failing.
The old test cases are passing with the removal.

ivanpauno:
Rebased with master before running CI.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7478/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3618/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6133/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7313/)

PR #40, opened 2019-07-12
Support xml and yaml files in ros2launch

ivanpauno:
What the title says.
The [first commit](https://github.com/ros2/launch_ros/commit/63daf079965157ba9fb798c3c54d652d56af08d5) doesn't depend on any PR. After the [second commit](https://github.com/ros2/launch_ros/commit/1ba59a7110807597fbdddb277b6ebfda1c9b6790), it depends on https://github.com/ros2/launch/pull/271, which allows better error handling a cleaner code.

TODO: Add tests. I've tested locally, and it works fine.

### DIFF
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -23,17 +23,14 @@ from typing import Tuple
 from ament_index_python.packages import get_package_share_directory
 from ament_index_python.packages import PackageNotFoundError
 import launch
-from launch.launch_description_sources import get_launch_description_from_python_launch_file
+from launch.launch_description_sources import AnyLaunchDescriptionSource
 from launch.launch_description_sources import InvalidPythonLaunchFileError
-from launch.launch_description_sources import load_python_launch_file_as_module
 import launch_ros
 import rclpy
 import rclpy.context
 
 # forward functions into this module's default namespace (useful for some autocompletion tools)
-get_launch_description_from_python_launch_file = get_launch_description_from_python_launch_file
 InvalidPythonLaunchFileError = InvalidPythonLaunchFileError
-load_python_launch_file_as_module = load_python_launch_file_as_module
 
 
 class MultipleLaunchFilesError(Exception):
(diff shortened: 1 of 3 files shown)


sloretz:
@hidmic [`ros2 launch` tab completion works with both](https://github.com/ros2/launch_ros/blob/5fac112b739a2c63222e3267a179a19fdfa6eef1/ros2launch/ros2launch/api/api.py#L82), though I was under the impression `_launch.py` was preferred too.

ros2launch/ros2launch/api/api.py
```diff
@@ -74,19 +67,19 @@ def get_share_file_path_from_package(*, package_name, file_name):
...
-    python_launch_file_paths = []
+def get_launch_file_paths(*, path):
+    """Return a list of paths to launch files within a given path."""
+    launch_file_paths = []
     for root, dirs, files in os.walk(path):
         for file_name in files:
-            if file_name.endswith('launch.py'):
-                python_launch_file_paths.append(os.path.join(root, file_name))
-    return python_launch_file_paths
+            if file_name.endswith(('.launch.py', '.launch.xml', '.launch.yaml')):
```
wjwwood: Yeah, adding the `.` to the `launch.py` isn't right. That was intentional so that anything ending with `launch.py` (including `foo_launch.py` and `foo.launch.py`) is acceptable. The same should (in my opinion) be applied to the new xml and yaml ones.

wjwwood:
Personally I still like `.launch.py` as I don't foresee my self importing it often and/or I don't mind importing it with `importlib`. For xml and yaml there's no reason to `_` instead unless we just want to use that everywhere. I don't remember where we left the conversion on what we recommend, but we didn't change any of the existing examples, e.g.:

https://github.com/ros2/demos/blob/dashing/demo_nodes_cpp/launch/topics/talker_listener.launch.py

ros2launch/ros2launch/api/api.py
ivanpauno: I agree.
Also, there's nothing stopping the user to run a file not following these naming conventions.
For example:
```bash
ros2 launch package filename_without_correct_ending
```

Will launch the file (regardless if it is an xml, yaml or python file).
But the tool will not autocomplete the name.

ros2launch/ros2launch/command/launch.py
```diff
@@ -109,33 +110,15 @@ def main(self, *, parser, args):
...
-Notice: ValueError can occur if your launch file ('{}') is a binary file and not a Python file.
-""".format(path), file=sys.stderr)
-            raise RuntimeError('ValueError: {}'.format(str(exc)))
-        except InvalidPythonLaunchFileError as exc:
-            # TODO(wjwwood): refactor this after we deprecate and then remove the old launch
-            print("""
-Notice: Your launch file may have been designed to be used with an older version of ROS 2.
-Or that the file you specified is Python code, but not a launch file.
-""", file=sys.stderr)
-            raise RuntimeError('InvalidPythonLaunchFileError: {}'.format(str(exc)))
```
wjwwood: Where did all of these special error handlers go?

wjwwood: I guess handled in part by https://github.com/ros2/launch/pull/271 and in part by only trying to interpret a launch file as a Python script if it ends in `.py`?

ivanpauno: After https://github.com/ros2/launch/pull/271, it works like this:
- Extension recognized: The "correct" loader (based on the extension) is tried first. If it fails, all the other loaders are also tried. If all of them fail, the exception of the "correct" loader is shown.
- Without extensions/not recognized extension: All loaders are tried. In case of failure a generic message is shown.

I should maybe move this special error handlers to `get_launch_description_from_python_launch_file`. I will do that.

The error raised by `get_launch_description_from_any_launch_file` is always `InvalidLaunchFileError`. The error information of the previously raised error is wrapped in it.

wjwwood: Sounds good to me, thanks for explaining.

ivanpauno:
CI (up to ros2launch launch launch_xml launch_yaml launch_ros test_launch_ros, only fastrtps):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7737/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3840/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6332/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7596/)

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7771/)

PR #42, opened 2019-07-18
Add PushRosNamespace action

ivanpauno:
Based on discussion https://github.com/ros2/design/pull/207#pullrequestreview-260917648, https://github.com/ros2/design/pull/207#issuecomment-510995374, and inspired on `Node` action [documentation](https://github.com/ros2/launch_ros/blob/5fac112b739a2c63222e3267a179a19fdfa6eef1/launch_ros/launch_ros/actions/node.py#L90-L97) (of a feature that was documented but not implemented).

This adds a way of scoping ROS entities in a namespace.

Example usage:
```python3
LaunchDescription([
    GroupAction([
        PushRosNamespace('my_ns'),
        IncludeLaunchDescription(PythonLaunchDescriptionSource('another_launch_file'),
    ])
])
```

I think we should consider in the future:
- Add `GroupAction` and `IncludeLaunchDescription` alternatives in `launch_ros`. Add to them an extra `ros_namespace` argument.
- Add `scoped` argument to `IncludeLaunchDescription` action (in `launch`), default to `True` (which will break api, but I think is more "normal").

### DIFF
--- a/launch_ros/launch_ros/actions/__init__.py
+++ b/launch_ros/launch_ros/actions/__init__.py
@@ -18,10 +18,12 @@ from .composable_node_container import ComposableNodeContainer
 from .lifecycle_node import LifecycleNode
 from .load_composable_nodes import LoadComposableNodes
 from .node import Node
+from .push_ros_namespace import PushRosNamespace
 
 __all__ = [
     'ComposableNodeContainer',
     'LifecycleNode',
     'LoadComposableNodes',
     'Node',
+    'PushRosNamespace',
 ]
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -166,7 +166,7 @@ class Node(ExecuteProcess):
         self.__arguments = arguments
 
         self.__expanded_node_name = '<node_name_unspecified>'
-        self.__expanded_node_namespace = '/'
+        self.__expanded_node_namespace = ''
         self.__final_node_name = None  # type: Optional[Text]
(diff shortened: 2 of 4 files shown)


wjwwood:
> Add GroupAction and IncludeLaunchDescription alternatives in launch_ros. Add to them an extra ros_namespace argument.

That is fine by me, though we have to address the namespacing in XML or instead of having new versions that "overload", find a way to extend the actions that are already in `launch` via `launch_ros`.

> Add scoped argument to IncludeLaunchDescription action (in launch), default to True (which will break api, but I think is more "normal").

I think it would be ok to do this at the XML level, but I would not do this at the action level. I intentionally left scoping to groups as to not complicate every action that might benefit from scoping, like the include action and the set launch configuration action.

wjwwood:
I would also not change the default behavior of scoping in the Python API, but again, I think that could be appropriate in the XML syntactic sugar.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7611/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3731/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6224/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7432/)

ivanpauno:
Double checking if linter is happy after https://github.com/ros2/launch_ros/pull/42/commits/5b64bd782e5d65ec3f7beaa67ae18ce86755fa14:

* Linux [![Build Status]* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7620/)

PR #44, opened 2019-07-25
Fix PushRosNamespace action

### LINKED ISSUE #43: Namespace not pushed by `PushRosNamespace`
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - Source
- Version or commit hash:
  - 68337f836b9dfbf4bed2b9c3b3ae03d21570cf13
- DDS implementation:
  -  Fast-RTPS

#### Steps to reproduce issue

Given the following `talker_launch.py`

``` python
import launch.actions
import launch_ros.actions

    return LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp',
            node_executable='talker',
            output='screen',
            node_name='talker')
    ])
```

And the following file with name `include_launch.py` that includes `talker_launch.py`:

```python
from launch.actions import IncludeLaunchDescription
from launch.launch_description_sources import PythonLaunchDescriptionSource
from launch.substitutions import ThisLaunchFileDir
from launch_ros.actions import PushRosNamespace


    return LaunchDescription([
        GroupAction([
            PushRosNamespace('talker_ns'),
            IncludeLaunchDescription(
                PythonLaunchDescriptionSource([ThisLaunchFileDir(), '/talker_launch.py']))
        ])
    ])

```

#### Expected behavior

We would expect the node name to be: `/talker_ns/talker`.

#### Actual behavior

However, it is: `/talker`

#### Additional information

But, if we provide `node_namespace=''` to the action in `talker_launch.py`, and launch `include_launch.py`, then the node name is `/talker_ns/talker`

ivanpauno:
Fixes #43.

Though the namespace is expanded correctly, it isn't added to the command when `node_namespace` argument is `None`:
https://github.com/ros2/launch_ros/blob/68337f836b9dfbf4bed2b9c3b3ae03d21570cf13/launch_ros/launch_ros/actions/node.py#L134-L137

I considered two possible fixes:
- Always add the namespace to the command.
- Change `node_namespace` default to `''`. When doing `node_namespace=None`, `__node:={}` argument won't be passed. If a relative namespace is passed, `ros_namespace` launch config will be used.
For absolute namespaces, `ros_namespace` will be ignored.

The second one is a little more flexible, so I applied that one.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -59,7 +59,7 @@ class Node(ExecuteProcess):
         package: SomeSubstitutionsType,
         node_executable: SomeSubstitutionsType,
         node_name: Optional[SomeSubstitutionsType] = None,
-        node_namespace: Optional[SomeSubstitutionsType] = '',
+        node_namespace: SomeSubstitutionsType = '',
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
         arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
@@ -131,10 +131,9 @@ class Node(ExecuteProcess):
             cmd += [LocalSubstitution(
                 'ros_specific_arguments[{}]'.format(ros_args_index), description='node name')]
             ros_args_index += 1
-        if node_namespace is not None:
-            cmd += [LocalSubstitution(
-                'ros_specific_arguments[{}]'.format(ros_args_index), description='node namespace')]
-            ros_args_index += 1
+        cmd += [LocalSubstitution(
+            'ros_specific_arguments[{}]'.format(ros_args_index), description='node namespace')]
+        ros_args_index += 1
         if parameters is not None:
             ensure_argument_type(parameters, (list), 'parameters', 'Node')
             # All elements in the list are paths to files with parameters (or substitutions that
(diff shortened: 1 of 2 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -284,13 +284,15 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
             if self.__node_namespace is not None:
```
wjwwood: Isn't this always true now? When would the user pass `None` here?

launch_ros/launch_ros/actions/node.py
```diff
@@ -59,7 +59,7 @@ def __init__(
         package: SomeSubstitutionsType,
         node_executable: SomeSubstitutionsType,
         node_name: Optional[SomeSubstitutionsType] = None,
-        node_namespace: Optional[SomeSubstitutionsType] = None,
+        node_namespace: Optional[SomeSubstitutionsType] = '',
```
wjwwood: I don't follow the distinction between `''` and `None` for this argument (below it still checks for if this `is not None`). Perhaps you could either update the doc block for this argument or remove the logic considering it to be `None`? When would the user pass `None` here?

ivanpauno: Yeah, I agree it was unclear.
After https://github.com/ros2/launch_ros/pull/44/commits/84420b413d6c0c177399d6847fabc241cb088361, the argument doesn't accept more `None`.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/7664/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3772/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6224/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7506/)

PR #52, opened 2019-08-05
Adapt to '--ros-args ... [--]'-based ROS args extraction

hidmic:
Precisely what the title says. Connected to https://github.com/ros2/rcl/pull/477.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_push_ros_namespace.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_push_ros_namespace.py
@@ -82,7 +82,5 @@ def test_push_ros_namespace(config):
         node_namespace=config.node_ns,
     )
     node._perform_substitutions(lc)
-    expected_cmd_len = 3 if config.expected_ns is not None else 1
-    assert expected_cmd_len == len(node.cmd)
     expected_ns = config.expected_ns if config.expected_ns is not None else ''
     assert expected_ns == node.expanded_node_namespace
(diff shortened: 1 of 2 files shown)


test_launch_ros/test/test_launch_ros/actions/test_push_ros_namespace.py
```diff
@@ -82,7 +82,5 @@ def test_push_ros_namespace(config):
         node_namespace=config.node_ns,
     )
     node._perform_substitutions(lc)
-    expected_cmd_len = 2 if config.expected_ns is not None else 1
-    assert expected_cmd_len == len(node.cmd)
```
ivanpauno: I think this should have been updated to:
```python3
    expected_cmd_len = 3 if config.expected_ns is not None else 2
    assert expected_cmd_len == len(node.cmd)
```

Checking that `__ns:=/` is not added when the expanded namespace is `''` is important.

hidmic: The thing, the original test was incorrect as `node.cmd` substitutions are not resolved by `node._perform_substitutions()`.

ivanpauno: The len of the list is the same before and after performing the substitutions, so it was ok correct.

hidmic: > The len of the list is the same before and after performing the substitutions

That is a result of the current implementation, and thus not guaranteed nor enforced by the API.

> Checking that __ns:=/ is not added when the expanded namespace is '' is important.

I agree with this, but the former test was not adequate for said purpose. We can change either `launch_ros.actions.Node` implementation or the test itself to query the fully resolved `cmd` instead of counting substitutions and expecting them to be right.

PR #57, opened 2019-08-28
Add substitution for finding package share directory

jacobperron:
Expose FindPackageShare as 'find-pkg' and FindPackage as 'find-pkg-root'.
The rationale is that getting the share directory is the more common use-case (e.g. including launch files).

This turns

    <include file="$(find-pkg foo)/share/foo/launch/foo.launch.xml">

into

    <include file="$(find-pkg foo)/launch/foo.launch.xml">

### DIFF
--- a/launch_ros/launch_ros/substitutions/find_package.py
+++ b/launch_ros/launch_ros/substitutions/find_package.py
@@ -30,21 +30,16 @@ from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 
 
-@expose_substitution('find-pkg-prefix')
 class FindPackage(Substitution):
     """
-    Substitution that tries to locate the package prefix of a ROS package.
+    Abstract base class for substitutions involving finding a package.
 
-    The ROS package is located using ament_index_python.
-
-    :raise: ament_index_python.packages.PackageNotFoundError when package is
-        not found during substitution.
+    Subclasses should implement the find() method.
     """
 
     def __init__(
         self,
         package: SomeSubstitutionsType,
-        *,
         find_package_func: Callable[[str], str] = get_package_prefix
     ) -> None:
         """Constructor."""
@@ -65,6 +60,17 @@ class FindPackage(Substitution):
         """Getter for package."""
         return self.__package
 
+    def find(self, package_name: Text) -> Text:
+        """
+        Find a directory for a package.
+
+        Called when the substitution is performed.
+
+        :param: package_name The name of the package.
+        :return: A directory related to the package.
+        """
+        raise NotImplementedError()
+
     def describe(self) -> Text:
         """Return a description of this substitution as a string."""
         pkg_str = ' + '.join([sub.describe() for sub in self.package])
@@ -73,8 +79,24 @@ class FindPackage(Substitution):
     def perform(self, context: LaunchContext) -> Text:
         """Perform the substitution by locating the package."""
         package = perform_substitutions(context, self.package)
-        result = self.__find_package_func(package)
+        result = self.find(package)
         return result
+
+
+@expose_substitution('find-pkg-prefix')
+class FindPackagePrefix(FindPackage):
+    """
+    Substitution that tries to locate the package prefix of a ROS package.
+
+    The ROS package is located using ament_index_python.
+
+    :raise: ament_index_python.packages.PackageNotFoundError when package is
+        not found during substitution.
+    """
+
+    def find(self, package_name: Text) -> Text:
+        """Find the package prefix."""
+        return get_package_prefix(package_name)
 
 
 @expose_substitution('find-pkg-share')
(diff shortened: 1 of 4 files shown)


jacobperron:
Design doc PR: https://github.com/ros2/design/pull/254/files

jacobperron:
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/7894/)

launch_ros/launch_ros/substitutions/find_package.py
```diff
@@ -39,10 +41,16 @@ class FindPackage(Substitution):
...
+    def __init__(
+        self,
+        package: SomeSubstitutionsType,
+        *,
+        find_package_func: Callable[[str], str] = get_package_prefix
+    ) -> None:
         """Constructor."""
         super().__init__()
         self.__package = normalize_to_list_of_substitutions(package)
+        self.__find_package_func = find_package_func
```
hidmic: @jacobperron out of curiosity, why pass in a function as opposed to regular method overriding by subclasses?

launch_ros/launch_ros/substitutions/find_package.py
```diff
@@ -39,10 +41,16 @@ class FindPackage(Substitution):
...
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
```
hidmic: @jacobperron nit: the message in the `AttributeError` this method may raise is always the same and does not match either of the frontend tags.

hidmic: @jacobperron same for the `describe()` method, it'll look the same for both substitutions.

launch_ros/launch_ros/substitutions/find_package.py
```diff
@@ -28,7 +30,7 @@
 from launch.utilities import perform_substitutions
 
 
-@expose_substitution('find-pkg')
+@expose_substitution('find-pkg-prefix')
 class FindPackage(Substitution):
```
hidmic: @jacobperron nit: should this be renamed to `FindPackagePrefix`?

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: I hadn't thought of that. I presume you mean defining a new method `find_package` that can be overridden by subclasses. Makes sense to me.

What do you think about turning `FindPackage` into an abstract class with a `find_package` method that needs to be overridden by subclasses? Then we can add a `FindPackagePrefix` (and leaves open the possibility for future subclasses like `FindPackageBin`).

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: See my comment above: https://github.com/ros2/launch_ros/pull/57#discussion_r321421929

launch_ros/launch_ros/substitutions/find_package.py
hidmic: Sure, that makes sense to me.

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: See 076b334

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: See 569ef58

launch_ros/launch_ros/substitutions/find_package.py
```diff
@@ -28,27 +30,28 @@
...
-    def __init__(self, package: SomeSubstitutionsType) -> None:
+    def __init__(
+        self,
+        package: SomeSubstitutionsType,
+        find_package_func: Callable[[str], str] = get_package_prefix
+    ) -> None:
         """Constructor."""
         super().__init__()
         self.__package = normalize_to_list_of_substitutions(package)
+        self.__find_package_func = find_package_func
```
hidmic: @jacobperron why keep the `self.__find_package_func` attribute?

launch_ros/launch_ros/substitutions/find_package.py
```diff
@@ -57,13 +60,56 @@ def package(self) -> List[Substitution]:
...
+        """Find the package prefix."""
+        return get_package_prefix(package_name)
+
+
+@expose_substitution('find-pkg-share')
+class FindPackageShare(FindPackage):
+    """
+    Substitution that tries to locate the share directory of a ROS package.
+
+    The directory located using ament_index_python.
```
hidmic: @jacobperron nit: 
```suggestion
    The directory is located using ament_index_python.
```

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8110/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4107/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6608/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/7991/)

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: Fixed 1cb949e

launch_ros/launch_ros/substitutions/find_package.py
jacobperron: Left-over by mistake. Removed: 1cb949e

Issue #55, opened 2019-08-20
yaml file passed to launch file doesn't take effect


xhuan28:
**Required Info:**

- Operating System:
  - Ubuntu18.04
- Installation type:
  - source
- Version or commit hash:
  - 84cdecd361a3ea274f77f0c1ec2f2c02097fe372
- DDS implementation:
  - fast-rtps
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

```
Pass the yaml file to launch.py, but it doesn't take effect.
```

#### Additional information

- launch file

```
import os
import launch
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode

def generate_launch_description():
    param_file_path = /my/file/path
    container = ComposableNodeContainer(
            node_name='my_container',
            node_namespace='',
            package='rclcpp_components',
            node_executable='component_container',
            composable_node_descriptions=[
                ComposableNode(
                    package='my_package',
                    node_plugin='namespace::Class',
                    node_name='my_node',
                    parameters=[param_file_path],),
            ],
            output='screen',
    )

    return launch.LaunchDescription([container])
```

- yaml file:
```
my_node:
  ros__parameters:
    enabled_param1: true
    enabled_param2: true
```

xhuan28:
More information:
But the parameters can be set successfully through dictionary one by one. Is there anything wrong with the namespace?
e.g.
```
                ComposableNode(
                    package='my_package',
                    node_plugin='namespace::Class',
                    node_name='my_node',
                    parameters=[param_file_path,                    # no effect
                                         {'enabled_param1': 'true'}],)   # successfully

```

xhuan28:
I have found out the root cause of this issue. It is caused by different naming rules between yaml file and direct dictionary parameter. A parameter from a yaml file will combine its name as a string like **"/ns.my_node.ros__parameters.my_param"**, while passing a parameter as a dictionary directly, the name should be "my_param". Parameter system takes these two parameters for different parameters, not the same one. So the parameter from a yaml file will not override the default value to take effect.
I think we need to make the naming rule consistent among different parameter sources. I submit a [PR#58](https://github.com/ros2/launch_ros/pull/58). Please have a review. Thanks.

mabelzhang:
Related issue https://github.com/ros2/rclcpp/issues/715

PR #58, opened 2019-08-29
Only keep base name of parameter name from yaml file

xhuan28:
A full name which includes namespace, node name and the string of "ros__parameters" from yaml file will cause an issue, See https://github.com/ros2/launch_ros/issues/55 for details.

Signed-off-by: Xiaojun <xiaojun.huang@intel.com>

### DIFF
--- a/launch_ros/launch_ros/utilities/normalize_parameters.py
+++ b/launch_ros/launch_ros/utilities/normalize_parameters.py
@@ -143,6 +143,12 @@ def normalize_parameter_dict(
         # Normalize the value next
         if isinstance(value, Mapping):
             # Flatten recursive dictionaries
+            pos = -1
+            for index, text_sub in enumerate(name):
+                if text_sub.text == 'ros__parameters':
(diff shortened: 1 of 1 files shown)


hidmic:
@xhuan28 right, you shouldn't have to specify it. 

Parameter YAML files as the ones you pass when running ROS aware processes on a terminal i.e. using `ros2 run` are meant to deal with the fact that many nodes may run in any given process. When you launch a composable node, you specify parameters only for that node. If put in a file, those parameters do not need to specify any node because you're already targeting one. 

Having said that, you may still feed in parameter YAML files that do specify nodes anyways (as described in https://github.com/ros2/rclcpp/issues/715#issuecomment-497392626) but I'm not positively sure that's what you'd want.

ivanpauno:
@xhuan28 Can this be closed?

hidmic:
@xhuan28 friendly ping

xhuan28:
It can be closed. Thanks.

PR #59, opened 2019-08-29
Use of -r/--remap flags where appropriate.

hidmic:
Precisely what the title says.

Connected to https://github.com/ros2/rcl/pull/491.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -131,7 +131,7 @@ class Node(ExecuteProcess):
         # The substitutions will get expanded when the action is executed.
         cmd += ['--ros-args']  # Prepend ros specific arguments with --ros-args flag
         if node_name is not None:
-            cmd += [LocalSubstitution(
+            cmd += ['-r', LocalSubstitution(
                 "ros_specific_arguments['name']", description='node name')]
         if parameters is not None:
             ensure_argument_type(parameters, (list), 'parameters', 'Node')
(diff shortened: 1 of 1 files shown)


ivanpauno:
The patch is not correct, you have to modify the local substitutions here:
https://github.com/ros2/launch_ros/blob/d10cb9ed675c6dbaf018dff632573376a5d22b26/launch_ros/launch_ros/actions/node.py#L348-L363

hidmic:
That was actually a tradeoff solution. Functionally it works the same way as if local substitutions had been updated, though it renders `ros_specific_arguments` incomplete, so much is true. Problem is, substitutions do not expand to `List[Text]` nor we have support for tokenizing `cmd` (https://github.com/ros2/launch/issues/263), and I purposefully wanted to keep these changes scoped.

Once https://github.com/ros2/launch/issues/263 is there, yes, we can change this.

ivanpauno:
> That was actually a tradeoff solution. Functionally it works the same way as if local substitutions had been updated, though it renders `ros_specific_arguments` incomplete, so much is true. Problem is, substitutions do not expand to `List[Text]` nor we have support for tokenizing `cmd` ([ros2/launch#263](https://github.com/ros2/launch/issues/263)), and I purposefully wanted to keep these changes scoped.

Ok, I understand.
I first thought that the following was possible:
- '-r' + local substitution added to the command.
- local substitution is replaced for an empty string, and then remapping rule parsing fails.

But after reading this for a second time, I think that that isn't possible.

PR #68, opened 2019-09-06
Promote special CLI rules to flags

hidmic:
This pull requests adapts `launch_ros` to the changes introduced by https://github.com/ros2/rcl/pull/495.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -139,7 +139,7 @@ class Node(ExecuteProcess):
             # evaluate to paths), or dictionaries of parameters (fields can be substitutions).
             i = 0
             for param in parameters:
-                cmd += ['-f', LocalSubstitution(
+                cmd += ['--params-file', LocalSubstitution(
                     "ros_specific_arguments['params'][{}]".format(i),
                     description='parameter {}'.format(i))]
                 i += 1
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -139,7 +139,7 @@ class Node(ExecuteProcess):
             # evaluate to paths), or dictionaries of parameters (fields can be substitutions).
             i = 0
             for param in parameters:
-                cmd += [LocalSubstitution(
+                cmd += ['-f', LocalSubstitution(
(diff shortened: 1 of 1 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -349,10 +349,7 @@ def execute(self, context: LaunchContext) -> Optional[List[Action]]:
         if self.__expanded_node_namespace != '':
             ros_specific_arguments['ns'] = '__ns:={}'.format(self.__expanded_node_namespace)
         if self.__expanded_parameter_files is not None:
-            ros_specific_arguments['params'] = []
-            param_arguments = cast(List[str], ros_specific_arguments['params'])
-            for param_file_path in self.__expanded_parameter_files:
-                param_arguments.append('__params:={}'.format(param_file_path))
+            ros_specific_arguments['params'] = self.__expanded_parameter_files
```
ivanpauno: IIUC, `-f` flag only accepts one parameter file, and `ros_specific_arguments['params']` (and `self.__expanded_parameter_files`) is a list.
And the limitation we talked in https://github.com/ros2/launch_ros/pull/59#issuecomment-526187711 isn't true.
Thought substitutions are usually performed to `Text` (and they should to please the type annotations), we were using `LocalSubstitutions` that were performed to `List[str]` or anything else (e.g.: `ros_specific_arguments['params']`).
See https://github.com/ros2/launch/blob/79975bf453af17fbc2848412d264dbcead623792/launch/launch/substitutions/local_substitution.py#L54.

hidmic: Hmm, a `LocalSubstitution` can evaluate to anything, that's true -- despite the fact we've type-annotated it as `Text`. But that's why I think we've been performing substitutions by indexing `ros_specific_arguments['params']`, and why I simply removed `__params:=` and prepended `-f` [above](https://github.com/ros2/launch_ros/pull/68/files/f6fdd8999f7506dfdc035705218ccfa84e4fec90#diff-d78062ac1f3a77c95190e470302b7e69R142).

ivanpauno: Sorry, I read the patch badly. I missed that we were doing `ros_specific_arguments['params'][i]` in the local substitution and not `ros_specific_arguments['params']`.

PR #69, opened 2019-09-11
Do not import rclpy nor launch_ros at module level.

hidmic:
Precisely what the title says. It does not solve the root cause of the problem (see https://github.com/ros2/build_cop/issues/235#issuecomment-529537182), but works around it.


hidmic:
CI run up to `launch_ros` and `test_launch_ros`:

- Windows: [Build Status](https://ci.ros2.org/job/ci_windows/7988/)

ivanpauno:
IMO, we should merge this PR asap, as is the easiest fix to https://github.com/ros2/build_cop/issues/235.

hidmic:
Alright, merging.

PR #72, opened 2019-09-19
Handle zero-width string parameters.

hidmic:
Precisely what the title says. Loading an empty string results in `None` when using the `yaml` module. 

Additional integration tests for zero-width parameters are added.

### DIFF
--- a/launch_ros/launch_ros/utilities/evaluate_parameters.py
+++ b/launch_ros/launch_ros/utilities/evaluate_parameters.py
@@ -67,7 +67,9 @@ def evaluate_parameter_dict(
             if isinstance(value[0], Substitution):
                 # Value is a list of substitutions, so perform them to make a string
                 evaluated_value = perform_substitutions(context, list(value))
-                yaml_evaluated_value = yaml.safe_load(evaluated_value)
+                yaml_evaluated_value = yaml.load(evaluated_value)
+                if yaml_evaluated_value is None:
+                    yaml_evaluated_value = ''
                 if type(yaml_evaluated_value) in (bool, int, float, str, bytes):
                     evaluated_value = yaml_evaluated_value
                 elif isinstance(yaml_evaluated_value, Sequence):
--- a/launch_ros/launch_ros/utilities/evaluate_parameters.py
+++ b/launch_ros/launch_ros/utilities/evaluate_parameters.py
@@ -67,7 +67,7 @@ def evaluate_parameter_dict(
             if isinstance(value[0], Substitution):
                 # Value is a list of substitutions, so perform them to make a string
                 evaluated_value = perform_substitutions(context, list(value))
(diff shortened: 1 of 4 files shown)


launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -67,7 +67,9 @@ def check_sequence_type_is_allowed(sequence):
             if isinstance(value[0], Substitution):
                 # Value is a list of substitutions, so perform them to make a string
                 evaluated_value = perform_substitutions(context, list(value))
-                yaml_evaluated_value = yaml.safe_load(evaluated_value)
+                yaml_evaluated_value = yaml.load(evaluated_value)
```
ivanpauno: why stop using safe load? I don't think that's a good idea.

hidmic: That's a mistake. A leftover of a previous attempt using a custom Loader for a different reason. Good catch!

hidmic:
> IMO, any hack around it is a bad idea.

That's connected to #74. 

Without this patch,

```xml
<param name="foo" value=""/>
```

results in `None` and `evaluate_parameters()` raises. You're then forced to do:

```xml
<param name="foo" value="''''"/>
```

for parameters (!) and 


```xml
<param name="foo" value="''"/>
```

for everything else.

ivanpauno:
> results in None and evaluate_parameters() raises. You're then forced to do:
> 
> `<param name="foo" value="''''"/>`
> for parameters (!) and
>
> `<param name="foo" value="''"/>`

I understand that's quite crappy, but:
- I don't know what `<param name="foo" value=""/>` mean in ROS 1.
    Does it set a parameter to an empty string, or does it use its default value?
- We should solve the underlying problem. Thus is, type inference from performed substitutions.
    Currently, each class is doing type conversions in a custom way.
- We should probably improve the type inference rules we're using in XML frontend.

hidmic:
I fully agree with that. In the meantime, we are not dealing with an empty string, and whether we fail or not depends on what that value ends up being used for. This patch doesn't make things better, but an actual solutoin for #74 won't be ready anytime soon.

ivanpauno:
> I fully agree with that. In the meantime, we are not dealing with an empty string, and whether we fail or not depends on what that value ends up being used for. This patch doesn't make things better, but an actual solutoin for #74 won't be ready anytime soon.

Ok. I don't like this much, but I'll approve after you address my review comment and confirm what ROS 1 is doing in this case.

hidmic:
@ivanpauno see https://github.com/ros2/launch_ros/pull/72/commits/f87c20e6fad4672c1ac21d31ac333afee5e4fabb. CI is in https://github.com/ros2/launch/pull/335.

PR #73, opened 2019-09-19
Refactor Node parse() function.

hidmic:
Precisely what the title says: ignore `cmd` from ExecuteProcess and parse `args` properly.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -232,9 +232,7 @@ class Node(ExecuteProcess):
         args = entity.get_attr('args', optional=True)
         if args is not None:
             kwargs['arguments'] = super()._parse_cmdline(args, parser)
-        node_name = entity.get_attr('node_name', optional=True)
-        if node_name is None and 'name' in kwargs:
-            node_name = kwargs['name']
+        node_name = entity.get_attr('node-name', optional=True)
         if node_name is not None:
             kwargs['node_name'] = node_name
         kwargs['package'] = parser.parse_substitution(entity.get_attr('pkg'))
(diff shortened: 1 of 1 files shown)


ivanpauno:
Depends on https://github.com/ros2/launch/pull/336

launch_ros/launch_ros/actions/node.py
```diff
@@ -228,11 +228,15 @@ def normalize_scalar_value(value):
...
-        del kwargs['args']
-        kwargs['node_name'] = kwargs['name']
-        del kwargs['name']
+        _, kwargs = super().parse(entity, parser, ignore=['cmd'])
+        args = entity.get_attr('args', optional=True)
+        if args is not None:
+            kwargs['arguments'] = super()._parse_cmdline(args, parser)
+        node_name = entity.get_attr('node_name', optional=True)
+        if node_name is None and 'name' in kwargs:
+            node_name = kwargs['name']
```
ivanpauno: This is quite strange. So, if you just pass `name` you're setting the `node_name`. But if you pass `node_name` and `name`, the first send the node name and the second the executable name.
What if someone wants to set the executable name without remapping the node name?

I originally avoided to expose the executable name, because I didn't like `<node ... node_name="..." .../>`.
But if we want to add it, I think it should look like: `<node ... name="..." exec-name="..." .../>`.

hidmic:
CI is in https://github.com/ros2/launch/pull/336.

launch_ros/launch_ros/actions/node.py
hidmic: Yeap, that change makes absolutely no sense. Thanks for pointing it out. But I'll simply stop doing creative renaming in the parsing function. It's certainly not where we'd want to be. See https://github.com/ros2/launch_ros/pull/73/commits/a4be9d08a85e00bf4cbc3977633d967dfb7f7fe1.

ivanpauno: Sounds good. We should probably revisit naming of attributes in other actions ...

ivanpauno:
@hidmic Can you update the tutorial and the design document that specifies the XML format?

Issue #75, opened 2019-07-17
secondary launch adding components into running containers


dirk-thomas:
For modularity the user might want to split a launch file into separate files which are started independently. One example would be a `base_launch.py` file which gets started once and a `dev_launch.py` file which contains stuff which the developer iterates on which is being started / stopped repeatedly.

For nodes running in separate processes these launch files can be started independently. What is not supported at the moment is for both launch files to contribute components into the same process. E.g. the second launch file want to inject a component into a process which was started by the first launch file.

The goal would be to support this use case within launch.

eric1221bday:
I've been dealing with this issue recently, and looking at the current implementation of `LoadComposableNodes`. It seems like some simple modifications should suffice to make this use case work. The class takes in a full `ComposableNodeContainer` but only really needs the `node_name` of that container.

Perhaps we can simply modify or add a new argument `target_container_name`, which, if filled, would be the one used to call the service call. That name can be a `str` or maybe even a `Substitution`

eric1221bday:
Oh right, so #77 which I authored should be able to deal with this problem

ivanpauno:
Fixed by https://github.com/ros2/launch_ros/pull/77.

PR #77, opened 2019-09-21
Allow separate launch composition

eric1221bday:
As mentioned in #75, this is my attempt to allow composing nodes in different launch files. This change allows this by making target_container also a substitution so users can pass in the container node name. That way the container node itself can be executed first in another launch file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,4 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .flake8
-.vscode
+
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .flake8
+.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -4,4 +4,3 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .flake8
-
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -25,8 +25,11 @@ from launch.action import Action
 from launch.launch_context import LaunchContext
 import launch.logging
 from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.some_substitutions_type import SomeSubstitutionsType_types_tuple
 from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
+from launch.utilities import ensure_argument_type
+from launch.utilities import is_a_subclass
 
 from .composable_node_container import ComposableNodeContainer
 
@@ -58,6 +61,12 @@ class LoadComposableNodes(Action):
         :param composable_node_descriptions: descriptions of composable nodes to be loaded
         :param target_container: the container to load the nodes into
         """
+        ensure_argument_type(
+            target_container,
+            list(SomeSubstitutionsType_types_tuple) + [ComposableNodeContainer],
+            'target_container',
+            'LoadComposableNodes'
+        )
         super().__init__(**kwargs)
         self.__composable_node_descriptions = composable_node_descriptions
         self.__target_container = target_container
@@ -160,11 +169,15 @@ class LoadComposableNodes(Action):
     ) -> Optional[List[Action]]:
         """Execute the action."""
         # resolve target container node name
-        if isinstance(self.__target_container, ComposableNodeContainer):
+
+        if is_a_subclass(self.__target_container, ComposableNodeContainer):
             self.__final_target_container_name = self.__target_container.node_name
-        else:  # TODO: maybe should do better type checking here
+        elif isinstance(self.__target_container, SomeSubstitutionsType_types_tuple):
             subs = normalize_to_list_of_substitutions(self.__target_container)
             self.__final_target_container_name = perform_substitutions(context, subs)
+        else:
+            self.__logger.error('target container is neither a ComposableNodeContainer nor a SubstitutionType')
+            return
 
         # Create a client to load nodes in the target container.
         self.__rclpy_load_node_client = context.locals.launch_ros_node.create_client(
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -16,13 +16,16 @@
 
 from typing import List
 from typing import Optional
+from typing import Text
+from typing import Union
 
 import composition_interfaces.srv
 
 from launch.action import Action
 from launch.launch_context import LaunchContext
 import launch.logging
-from launch.utilities import ensure_argument_type
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 
 from .composable_node_container import ComposableNodeContainer
@@ -36,11 +39,11 @@ class LoadComposableNodes(Action):
     """Action that loads composable ROS nodes into a running container."""
 
     def __init__(
-        self,
-        *,
-        composable_node_descriptions: List[ComposableNode],
-        target_container: ComposableNodeContainer,
-        **kwargs,
+            self,
+            *,
+            composable_node_descriptions: List[ComposableNode],
+            target_container: Union[SomeSubstitutionsType, ComposableNodeContainer],
+            **kwargs,
     ) -> None:
         """
         Construct a LoadComposableNodes action.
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -4,3 +4,4 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .flake8
+.vscode
```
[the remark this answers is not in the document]
mjcarroll: It would seem safer to add it to your global .gitignore if VSCode is your editor of choice.  I believe @dirk-thomas uses VSCode, is that the approach that you use?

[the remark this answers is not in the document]
dirk-thomas: Since there is an almost infinite amount of possible editors and custom config files, yes, I think those should be ignored by the user globally.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -55,15 +58,10 @@ def __init__(
         :param composable_node_descriptions: descriptions of composable nodes to be loaded
         :param target_container: the container to load the nodes into
         """
-        ensure_argument_type(
-            target_container,
-            ComposableNodeContainer,
-            'target_container',
-            'LoadComposableNodes'
-        )
```
hidmic: @eric1221bday why dropping the type check? `ensure_argument_type` can take more than one type.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -39,7 +42,7 @@ def __init__(
         self,
         *,
         composable_node_descriptions: List[ComposableNode],
-        target_container: ComposableNodeContainer,
+        target_container: Union[SomeSubstitutionsType, ComposableNodeContainer],
```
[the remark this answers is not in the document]
hidmic: Actually, I'd rather have a single argument. In the most general case a user would have to provide a fully qualified node name (FQNN), but in Python we offer the chance to pass in a `ComposableNodeContainer` action from which an FQNN can be fetched. The later and current behavior is a bit of a heuristic, as the underlying node may ignore all name remappings.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -161,10 +159,17 @@ def execute(
         context: LaunchContext
     ) -> Optional[List[Action]]:
         """Execute the action."""
+        # resolve target container node name
+        if isinstance(self.__target_container, ComposableNodeContainer):
+            self.__final_target_container_name = self.__target_container.node_name
+        else:  # TODO: maybe should do better type checking here
+            subs = normalize_to_list_of_substitutions(self.__target_container)
+            self.__final_target_container_name = perform_substitutions(context, subs)
```
hidmic: @eric1221bday why not resolving this TODO now?

launch_ros/launch_ros/actions/load_composable_nodes.py
eric1221bday: @hidmic It's because it has trouble with Union types, which is what SubstitutionType ultimately is. Alternative we can list out every single type a substitution can possibly be, but that felt very hacky to me.

launch_ros/launch_ros/actions/load_composable_nodes.py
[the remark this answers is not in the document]
eric1221bday: My personal thought when designing this was to reduce impact on current API, hence why a union type. I also started off thinking it should've been a separate argument. I agree that this makes the argument name vague with dual meaning, but I think this can be handled via better docstrings and docs in general.

launch_ros/launch_ros/actions/load_composable_nodes.py
eric1221bday: @hidmic So I meant to revisit this, the core of the issue is the same reason I removed the ensure_argument_type at the top, which is that Unions are tricky to type check. There's a stackoverflow [here](https://stackoverflow.com/a/49471187) with a suggested way of checking using both the `__origin__` and `__args__` attribute, but I'm unsure how advisable that is.

.gitignore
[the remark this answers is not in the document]
eric1221bday: removed

launch_ros/launch_ros/actions/load_composable_nodes.py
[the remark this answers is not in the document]
wjwwood: I like one argument that can either be a string (substitution) or a container node class. That way the two ways of specifying are mutually exclusive by design.

launch_ros/launch_ros/actions/load_composable_nodes.py
wjwwood: You can use `SomeSubstitutionsType_types_tuple`, e.g.:

https://github.com/ros2/launch/blob/8f1ed681973aad9b48ecc6712e54d2f66ee4dac3/launch/launch/actions/timer_action.py#L66

launch_ros/launch_ros/actions/load_composable_nodes.py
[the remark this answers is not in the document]
ivanpauno: ok, so one argument then.

launch_ros/launch_ros/actions/load_composable_nodes.py
wjwwood: As I said above you can use the types tuple we have, but if you choose to leave the TODO, please put your github handle into it (that's our convention): `# TODO(eric1221bday): ...`

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -161,10 +159,17 @@ def execute(
         context: LaunchContext
     ) -> Optional[List[Action]]:
         """Execute the action."""
+        # resolve target container node name
+        if isinstance(self.__target_container, ComposableNodeContainer):
```
wjwwood: Will this work if someone were to subclass `ComposableNodeContainer`? Ideally it would. I don't remember what the gotcha's are there, but I know we have this utility for some reason: https://github.com/ros2/launch/blob/a46aabdde1f58c7e2e5c056c0c4a1ea830faac5c/launch/launch/utilities/class_tools_impl.py#L34 which uses `issubclass()` internally.

eric1221bday: @wjwwood so at least according to [this](https://stackoverflow.com/a/1549854) it seems like it should, but I'm happy to use the created util function instead

launch_ros/launch_ros/actions/load_composable_nodes.py
eric1221bday: @wjwwood added it back in with suggested tuple, seems to work

launch_ros/launch_ros/actions/load_composable_nodes.py
eric1221bday: @wjwwood alright I ended up using `is_a_subclass` for the ComposableNodeContainer and `isinstance` with `SomeSubstitutionsType_types_tuple` to do the type checking. If the argument is neither I will log and error and return. Considering the type checking at the constructor this is a bit of an unlikely hit, but no harm really.

launch_ros/launch_ros/actions/load_composable_nodes.py
eric1221bday: As mentioned above, ended but using `is_a_subclass`

ivanpauno: @wjwwood I don't understand `is_a_subclass` implementation.
[isinstance](https://docs.python.org/2/library/functions.html#isinstance) already returns `True` if the object is instance of a subclass.
[issubclass](https://docs.python.org/2/library/functions.html#issubclass) returns `True` if a class is a subclass of another (and not when an object is instance of a subclass).
So I don't understand why `is_a_subclass` checks for both `isinstance` and `issubclass`.

IMO, using [is_a](https://github.com/ros2/launch/blob/a46aabdde1f58c7e2e5c056c0c4a1ea830faac5c/launch/launch/utilities/class_tools_impl.py#L25) util in launch or `isinstance` is ok.

wjwwood: @ivanpauno Yeah, I don't remember a rationale off-hand, so maybe your recommendation is sufficient.

eric1221bday:
@ivanpauno so is this good?

ivanpauno:
> @ivanpauno so is this good?

Yes, it's good.
We can follow-up about `is_a_subclass` usage later (https://github.com/ros2/launch_ros/pull/77#discussion_r328781278).

----

We're currently in a feature freeze, so this is not going to be merged into `master` until we create an Eloquent branch (probably in one/two weeks).

eric1221bday:
@ivanpauno gentle ping now that the feature freeze seems to be over.

.gitignore
```diff
@@ -4,3 +4,4 @@ __pycache__
 .mypy_cache
 .pytest_cache
 .flake8
+
```
ivanpauno: nit: delete unnecessary change.

ivanpauno:
@eric1221bday please address [flake8 failures](https://ci.ros2.org/job/ci_linux/8764/testReport/junit/launch_ros.test/test_flake8/test_flake8/).

eric1221bday:
@ivanpauno should be fixed

ivanpauno:
Sanity check (using rebased branch `ivanpauno/test-#77`):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8793/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4613/)
* MacOS [Build Status](http://ci.ros2.org/job/ci_osx/7207/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8713/)

ivanpauno:
Going in! Thanks for the contribution @eric1221bday !

jacobperron:
I'd like to see this change backported to Eloquent.

Since it looks like mainly bug fixes since the last release, maybe we can sync the eloquent branch with master?

ivanpauno:
> Since it looks like mainly bug fixes since the last release, maybe we can sync the eloquent branch with master?

Sounds reasonable to me.

PR #82, opened 2019-10-04
Add support for launching nodes not in a package

hidmic:
Precisely what the title says. Requiring nodes to be installed for `launch_ros.actions.Node` to be able to launch them is an unnecessary restriction. This way, packages with launch tests can have fixture node executables that do not get installed anywhere.

Follow-up PR after https://github.com/ros2/launch_ros/pull/80. Useful in upcoming PRs.

PR #83, opened 2019-10-08
Fix launch_ros.actions.Node parsing function

hidmic:
Follow-up after #82. Fixes https://github.com/ros2/build_cop/issues/244.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py
+++ b/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py
@@ -16,6 +16,7 @@
 
 import io
 import pathlib
+import sys
 import textwrap
 
 from launch import LaunchService
@@ -31,7 +32,7 @@ xml_file = \
     <launch>
         <let name="a_string" value="\'[2, 5, 8]\'"/>
         <let name="a_list" value="[2, 5, 8]"/>
-        <node pkg="demo_nodes_py" exec="talker_qos" output="screen" name="my_node" namespace="my_ns" args="--number_of_cycles 1">
+        <node pkg="demo_nodes_py" exec="talker_qos" output="screen" node-name="my_talker" namespace="my_ns" args="--number_of_cycles 1">
             <param name="param1" value="ads"/>
             <param name="param_group1">
                 <param name="param_group2">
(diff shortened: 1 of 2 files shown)


hidmic:
CI up to `launch_testing_ros` and `test_launch_ros`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/8307/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4258/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/6775/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/8210/)

hidmic:
So, test failures on Windows stem from path backslashes being escaped by launch frontend parsers. 

We, @ivanpauno and I, think that having to escape backslashes when pure Windows paths want to be used in an XML launch file is fine, but being forced to specify paths in a platform specific way isn't. It turns launch files platform-specific. Actions should handle POSIX paths. 

I'll open an issue to track this enhancement.

hidmic:
Alright, all green! @wjwwood @clalancette PTAL so I can merge.

test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py
```diff
@@ -26,12 +27,17 @@
 import pytest
 
 yaml_params = str(pathlib.Path(__file__).parent / 'params.yaml')
+
+# Escape backslashes if any to keep them after parsing takes place
+yaml_params = yaml_params.replace('\\', '\\\\')
```
wjwwood: Could you use `as_posix` instead of `str()` above to avoid this?

https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.as_posix

hidmic: Hmm, I don't have a Windows box at hand to try it out, but if we do that then that POSIX path will be forwarded to the underlying node. And `rcl` will use `fopen` to attempt to parse the file with that path. And I think `fopen` will not handle POSIX paths in Windows.

wjwwood: I see.

ivanpauno: > Hmm, I don't have a Windows box at hand to try it out, but if we do that then that POSIX path will be forwarded to the underlying node. And `rcl` will use `fopen` to attempt to parse the file with that path. And I think `fopen` will not handle POSIX paths in Windows.

I'm almost sure that most native Python functions that takes paths accept both posix and windows' style paths.
I remember I tested that recently and it worked.
Particularly, I think I tested `open`, but it's better to double check.

wjwwood:
@hidmic any reason to not merge this?

hidmic:
No reason. I didn't on Friday because of your last comment and the discussion it sparked.

Merging now!

PR #99, opened 2019-11-29
Maintain order of parameters regarding name and from tag

### LINKED ISSUE #95: parameter order in XML is not maintained
When a XML launch file declares parameters for a node the order of the `param` tags is not maintained if some use `name` and some use `from`: https://github.com/ros2/launch_ros/blob/025596b2ec2e4cbd844cfc0ef660d9b3240e3457/launch_ros/launch_ros/actions/node.py#L210-L228

As a consequence when multiple of the `param` tags with different attributes set the same parameter the actually set value is unexpected since it might not be the last one.

BMarchi:
Fixes #95 . As the title says, maintain the order of the parameters if it has the `from` or `name` tag.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -208,7 +208,6 @@ class Node(ExecuteProcess):
             return param_dict
 
         normalized_params = []
-        params_without_from = []
         for param in params:
             from_attr = param.get_attr('from', optional=True)
             name = param.get_attr('name', optional=True)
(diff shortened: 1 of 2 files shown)


BMarchi:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8762/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4590/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7161/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/8690/)

hidmic:
Alright, all green. Going in!

PR #101, opened 2019-12-03
Pass the node-name attribute through the substitution parser

mxgrey:
Without this, it's not possible to use substitutions in the `node-name` attribute.


ivanpauno:
Thanks for the fix @mxgrey !

Sanity check:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8782/)
- Arch [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4604/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7187/)
- Windows [Build Status](http://ci.ros2.org/job/ci_windows/8701/)

PR #106, opened 2019-12-12
Fix misleading deprecated warning when using launch arguments

ivanpauno:
Before this patch, running:
`ros2 launch any_package any_launch_file some_arg:=some_value`
ended up with a warning:
```
[WARN] [1576175943.679492205] [rcl]: Found remap rule 'asd:=bsd'. This syntax is deprecated. Use '--ros-args --remap asd:=bsd' instead.
```

That's misleading, because the only way of passing launch arguments is by adding `key:=value` at the end of the command.

To fix this, I did the following:
- Stop using the `LaunchContext`  arguments in the initialization of the `rclpy.Context`: https://github.com/ros2/launch_ros/blob/d6dd5c0e75fb6cdbfb9b65b8468b4b0b5b2172da/launch_ros/launch_ros/default_launch_description.py#L60
  Add a custom way of passing those arguments.
- Pass no arguments in the creation of the `rclpy.Context` in `ros2launch` https://github.com/ros2/launch_ros/blob/8fae305d9aa1f4604c55de02c6224b8827cf09ee/ros2launch/ros2launch/api/api.py#L144
  Before, we were passing the same key values configurations that `ros2launch` received, which can cause problems (e.g.: being used as remapping rules).
- Finally, deleted the `argv` option of `ros2launch` arguments. IIUC, it was always empty as the previous argument (launch configurations) was taking a variable number of options.

I'm not sure if this is the correct way to go, maybe we want to do something different.
e.g.: Filtering the launch configurations from the arguments, and use the others as the arguments of the `rclpy.Context`. IMO, that may be confusing.

### DIFF
--- a/launch_ros/launch_ros/default_launch_description.py
+++ b/launch_ros/launch_ros/default_launch_description.py
@@ -28,11 +28,12 @@ from rclpy.executors import SingleThreadedExecutor
 class ROSSpecificLaunchStartup(launch.actions.OpaqueFunction):
     """Does ROS specific launch startup."""
 
-    def __init__(self, rclpy_context=None):
+    def __init__(self, rclpy_context=None, argv=None):
         """Create a ROSSpecificLaunchStartup opaque function."""
         super().__init__(function=self._function)
         self.__shutting_down = False
         self.__rclpy_context = rclpy_context
+        self.__argv = None
 
     def _shutdown(self, event: launch.Event, context: launch.LaunchContext):
         self.__shutting_down = True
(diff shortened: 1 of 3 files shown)


ivanpauno:
@Karsten1987 FYI.

ivanpauno:
- Linux [Build Status](https://ci.ros2.org/job/ci_linux/8891/)
- AArch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/4680/)
- macOS [Build Status](http://ci.ros2.org/job/ci_osx/7280/)

ivanpauno:
> I don't know if we need to pass ROS arguments to the internal launch_ros node or not, but for now it's probably ok to not do so.

As far as I see, we don't need to pass remapping rules or parameters.
The other ROS arguments are logging options, for which we could provide a different way of configuring that.

ivanpauno:
https://github.com/ros2/launch_ros/issues/120 reintroduced this bug.

I'm working on a fix and a regression test.

PR #113, opened 2020-01-16
Fix LoadComposableNodes action so that loading happens asynchronously

### LINKED ISSUE #107: LoadComposableNodes action blocks previous actions
## Bug report

In particular, it would be really convenient to be able to start a node container just before some actions to load components into it. Maybe not a *bug*, but certainly a big usability improvement.

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  -  source
- Version or commit hash:
  - d6dd5c0e75fb6cdbfb9b65b8468b4b0b5b2172da
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue

Launch the following launch file:
```python
from launch_ros.actions import ComposableNodeContainer
from launch_ros.actions import LoadComposableNodes
from launch_ros.descriptions import ComposableNode


    return LaunchDescription([
        ComposableNodeContainer(
           package='rclcpp_components', node_executable='component_container',
           node_name='my_container', node_namespace=''
        ),  
        LoadComposableNodes(
                ComposableNode(
                    package='composition',
                    node_plugin='composition::Talker'
                )
            ],  
            target_container='my_container'
        ),  
    ])
```

#### Expected behavior

A container node starts and the composition talker component is loaded.

#### Actual behavior

The container does not start and so the `LoadComposableNodes` 

#### Additional information

I understand that the `LoadComposableNodes` action is blocking until the service is available. But my intuition says that since the ComposableNodeContainer action appears first, it should be started before the loading action starts.

After talking with @wjwwood, it sounds like we should update the implementation of `LoadComposableNodes` so it waits for the service in a separate thread so that it doesn't block the asyncio loop.

jacobperron:
Fixes #107.

I'm not sure if this is the best solution, but it gets the job done.

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -14,8 +14,6 @@
 
 """Module for the LoadComposableNodes action."""
 
-import threading
-
 from typing import List
 from typing import Optional
 from typing import Text
@@ -158,9 +156,7 @@ class LoadComposableNodes(Action):
         """
         next_composable_node_description = composable_node_descriptions[0]
         composable_node_descriptions = composable_node_descriptions[1:]
-        thread = threading.Thread(
-             target=self._load_node, args=(next_composable_node_description, context))
-        thread.start()
+        self._load_node(next_composable_node_description, context)
         if len(composable_node_descriptions) > 0:
             context.add_completion_future(
                 context.asyncio_loop.run_in_executor(
@@ -192,4 +188,9 @@ class LoadComposableNodes(Action):
                 self.__final_target_container_name
             )
         )
-        self._load_in_sequence(self.__composable_node_descriptions, context)
+
+        context.add_completion_future(
+            context.asyncio_loop.run_in_executor(
+                None, self._load_in_sequence, self.__composable_node_descriptions, context
+            )
+        )
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -14,6 +14,8 @@
 
 """Module for the LoadComposableNodes action."""
 
+import threading
+
 from typing import List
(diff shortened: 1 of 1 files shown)


launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -156,7 +158,9 @@ def _load_in_sequence(
         """
         next_composable_node_description = composable_node_descriptions[0]
         composable_node_descriptions = composable_node_descriptions[1:]
-        self._load_node(next_composable_node_description, context)
+        thread = threading.Thread(
+             target=self._load_node, args=(next_composable_node_description, context))
+        thread.start()
```
wjwwood: @jacobperron and I talked in person about this, but just to summarize, I think this thread needs to be joined at some point and we need to indicate to the launch service that this action has some asynchronous behavior, by overriding the `get_asyncio_future()` method: https://github.com/ros2/launch/blob/4333910ba41977f3ab9867082c86a8698ada3a16/launch/launch/launch_description_entity.py#L137

jacobperron: On second look, I realized that the code was already making asynchronous calls and providing futures to the launch service with [add_completion_future](https://github.com/ros2/launch_ros/blob/04ce971240eb39f43ca46cf3cc0b3cef4501f4a1/launch_ros/launch_ros/actions/load_composable_nodes.py#L161-L164). The bug was that the first call was not following the same pattern, which I've fixed in 04ce971

PTAL

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9186/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/4889/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7491/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9095/)

PR #111, opened 2019-12-31
Fix frontend topic remapping

jacobperron:
Pass data_type parameter to remap entity.

Without this change remap entities will not work, for example:

```xml
<launch>
  <node name="talker" pkg="demo_nodes_cpp" exec="talker">
    <remap from="chatter" to="foo" />
  </node>
</launch>
```

PR #104, opened 2019-12-10
check for shutdown while waiting for a service response to avoid hang during shutdown

wjwwood:
Should fix https://github.com/ros2/launch_ros/issues/102

PR #100, opened 2019-12-02
Fix push-ros-namespace in xml/yaml launch files

ivanpauno:
`expose_action` was missing, and there were no test for markup based launch files.

PR #98, opened 2019-11-25
Add pid to launch_ros node name as suffix

BMarchi:
Fixes #2 . As the title says, it adds the pid to the node's name as a suffix. Is there any need to create a test for this?

PR #115, opened 2020-01-17
Eloquent patch release 1

ivanpauno:
Backports:
- https://github.com/ros2/launch_ros/pull/111
- https://github.com/ros2/launch_ros/pull/104
- https://github.com/ros2/launch_ros/pull/106 (this breaks API, but of something completly unused and that was causing failures)
- https://github.com/ros2/launch_ros/pull/103 (just to avoid conflicts with other backported commits)
- https://github.com/ros2/launch_ros/pull/99
- https://github.com/ros2/launch_ros/pull/100
- https://github.com/ros2/launch_ros/pull/101
- https://github.com/ros2/launch_ros/pull/98

Other stuff that could be included:
- #77 (I didn't include it because it's an enhancement, not a fix).

### DIFF
--- a/launch_ros/examples/lifecycle_pub_sub_launch.py
+++ b/launch_ros/examples/lifecycle_pub_sub_launch.py
@@ -32,7 +32,7 @@ import lifecycle_msgs.msg
 
 
 def main(argv=sys.argv[1:]):
-    """Main."""
+    """Run lifecycle nodes via launch."""
     ld = launch.LaunchDescription()
 
     # Prepare the talker node.
--- a/launch_ros/examples/pub_sub_launch.py
+++ b/launch_ros/examples/pub_sub_launch.py
@@ -28,7 +28,7 @@ import launch_ros.actions
 
 
 def main(argv=sys.argv[1:]):
-    """Main."""
+    """Run demo nodes via launch."""
     ld = LaunchDescription([
         launch_ros.actions.Node(
             package='demo_nodes_cpp', node_executable='talker', output='screen',
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -15,6 +15,7 @@
 """Module for the LifecycleNode action."""
 
 import functools
+import threading
 from typing import cast
 from typing import List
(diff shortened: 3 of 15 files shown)


mjcarroll:
DCO is unhappy, at a minimum.

ivanpauno:
> DCO is unhappy, at a minimum.

The commit is signed. It seems to be a false-positive, but I can add my signature if that makes the checker happier.

ivanpauno:
> > DCO is unhappy, at a minimum.
> 
> The commit where the checker is failing is signed. It seems to be a false-positive, but I can add my signature if that makes the checker happier.

Sorry, there is actually one commit failing because of a missing DCO, and some others  failing because of strange reasons. I can add my signature in all of them to avoid the checker to fail, but I don't know if that's correct. Also, if the commit is already in master, IDK if I should add an extra signature.

dirk-thomas:
If you co-author a commit (e.g. a cherry pick or merge) you need to sign that commit too to pass DCO.

ivanpauno:
> If you co-author a commit (e.g. a cherry pick or merge) you need to sign that commit too to pass DCO.

That wasn't exactly the failure that the DCO checker was showing, but done.

mjcarroll:
Can you fix that linter error?

ivanpauno:
> Can you fix that linter error?

Are you talking about the errors here c#issuecomment-575773549?

Does are from `launch` repository, and not from `launch_ros`.
They are there just because I didn't backport https://github.com/ros2/launch/pull/362.
We usually don't backport those commits, but I can add it to the list if you want.

ivanpauno:
CI is the same as in: https://github.com/ros2/launch/pull/370#issuecomment-575773549

wjwwood:
I'm concerned about how #106 breaks behavior and API, but I think that probably no one is using it (or that probably no one would be affected by it) and it would make more sense to fix it. But we've been wrong about that in the past.

So, I'd tentatively vote to backport it and then make an announcement with the patch release letting people know. But I'll leave that up to @mjcarroll to decide how to proceed.

mjcarroll:
I'll add it to the announcement.  If it causes issues, we can revert, but I also believe it to be unlikely.

PR #122, opened 2020-02-06
Rename node-related parameters

jacobperron:
The 'node-' prefix seems redundant since the attribute is part of a "node" tag.
Add a deprecation warning in case the old name is used.

I couldn't find a reason why 'node-name' was preferred in https://github.com/ros2/launch_ros/pull/73.
I'm suggesting in this PR that we change it 'name', which seems more intuitive for me, especially coming from ROS 1.

### DIFF
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -35,8 +35,8 @@ class ComposableNodeContainer(Node):
     def __init__(
         self,
         *,
-        name: SomeSubstitutionsType = None,
-        namespace: SomeSubstitutionsType = None,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: Optional[SomeSubstitutionsType] = None,
         node_name: Optional[SomeSubstitutionsType] = None,
         node_namespace: Optional[SomeSubstitutionsType] = None,
         composable_node_descriptions: Optional[List[ComposableNode]] = None,
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -48,7 +48,8 @@ class ComposableNodeContainer(Node):
         Most arguments are forwarded to :class:`launch_ros.actions.Node`, so see the documentation
         of that class for further details.
 
-        .. warning:: Parameters `node_name` and `node_namespace` are deprecated.
+        .. deprecated:: Foxy
+           Parameters `node_name` and `node_namespace` are deprecated.
            Use `name` and `namespace` instead.
 
         :param: name the name of the node, mandatory for full container node name resolution
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -16,6 +16,7 @@
 
 from typing import List
 from typing import Optional
+import warnings
 
 from launch.action import Action
 from launch.actions import RegisterEventHandler
@@ -34,8 +35,10 @@ class ComposableNodeContainer(Node):
     def __init__(
         self,
         *,
-        node_name: SomeSubstitutionsType,
-        node_namespace: SomeSubstitutionsType,
+        name: SomeSubstitutionsType = None,
+        namespace: SomeSubstitutionsType = None,
+        node_name: Optional[SomeSubstitutionsType] = None,
+        node_namespace: Optional[SomeSubstitutionsType] = None,
         composable_node_descriptions: Optional[List[ComposableNode]] = None,
         **kwargs
     ) -> None:
@@ -45,12 +48,35 @@ class ComposableNodeContainer(Node):
         Most arguments are forwarded to :class:`launch_ros.actions.Node`, so see the documentation
         of that class for further details.
 
-        :param: node_name the name of the node, mandatory for full container node name resolution
-        :param: node_namespace the ros namespace for this Node, mandatory for full container node
-           name resolution
+        .. warning:: Parameters `node_name` and `node_namespace` are deprecated.
+           Use `name` and `namespace` instead.
+
+        :param: name the name of the node, mandatory for full container node name resolution
+        :param: namespace the ROS namespace for this Node, mandatory for full container node
+             name resolution
+        :param: node_name (DEPRECATED) the name of the node, mandatory for full container node
+            name resolution
+        :param: node_namespace (DEPRECATED) the ros namespace for this Node, mandatory for full
+            container node name resolution
         :param composable_node_descriptions: optional descriptions of composable nodes to be loaded
         """
-        super().__init__(node_name=node_name, node_namespace=node_namespace, **kwargs)
+        if node_name is not None:
+            warnings.warn("The parameter 'node_name' is deprecated, use 'name' instead")
+            if name is not None:
+                raise RuntimeError(
+                    "Passing both 'node_name' and 'name' parameters. Only use 'name'."
+                )
+            name = node_name
+        if node_namespace is not None:
+            warnings.warn("The parameter 'node_namespace' is deprecated, use 'namespace' instead")
+            if namespace is not None:
+                raise RuntimeError(
+                    "Passing both 'node_namespace' and 'namespace' parameters. "
+                    "Only use 'namespace'."
+                )
+            namespace = node_namespace
+
+        super().__init__(name=name, namespace=namespace, **kwargs)
         self.__composable_node_descriptions = composable_node_descriptions
 
     def execute(self, context: LaunchContext) -> Optional[List[Action]]:
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -60,7 +60,7 @@ class LifecycleNode(Node):
             one, or even all lifecycle nodes, and it requests the targeted nodes
             to change state, see its documentation for more details.
         """
-        super().__init__(node_name=node_name, **kwargs)
+        super().__init__(name=node_name, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
         self.__current_state = \
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -20,6 +20,7 @@ from typing import cast
 from typing import List
 from typing import Optional
 from typing import Text
+import warnings
 
 import launch
 from launch.action import Action
@@ -36,7 +37,7 @@ from ..events.lifecycle import StateTransition
 class LifecycleNode(Node):
     """Action that executes a ROS lifecycle node."""
 
-    def __init__(self, *, node_name: Text, **kwargs) -> None:
+    def __init__(self, *, name: Text = None, node_name: Optional[Text] = None, **kwargs) -> None:
         """
         Construct a LifecycleNode action.
 
@@ -49,7 +50,7 @@ class LifecycleNode(Node):
         - :class:`launch.events.lifecycle.StateTransition`:
 
             - this event is emitted when a message is published to the
-              "/<node_name>/transition_event" topic, indicating the lifecycle
+              "/<name>/transition_event" topic, indicating the lifecycle
               node represented by this action changed state
 
         This action also handles some events related to lifecycle:
@@ -59,8 +60,25 @@ class LifecycleNode(Node):
           - this event can be targeted to a single lifecycle node, or more than
             one, or even all lifecycle nodes, and it requests the targeted nodes
             to change state, see its documentation for more details.
+
+        .. warning:: The parameter `node_name` is deprecated, use `name` instead.
+
+        :param name: The name of the lifecycle node.
+          Although it defaults to None it is a required parameter and the default will be removed
+          in a future release.
+        :param node_name: (DEPRECATED) The name fo the lifecycle node.
         """
-        super().__init__(name=node_name, **kwargs)
+        if node_name is not None:
+            warnings.warn("The parameter 'node_name' is deprecated, use 'name' instead")
+            if name is not None:
+                raise RuntimeError(
+                    "Passing both 'node_name' and 'name' parameters. Only use 'name'."
+                )
+            name = node_name
+        # TODO(jacobperron): Remove default value and this check when deprecated API is removed
+        if name is None:
+            raise RuntimeError("'name' must not be None.'")
+        super().__init__(name=name, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
         self.__current_state = \
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -67,7 +67,8 @@ class LifecycleNode(Node):
             one, or even all lifecycle nodes, and it requests the targeted nodes
             to change state, see its documentation for more details.
 
-        .. warning:: The parameter `node_name` is deprecated, use `name` instead.
+        .. deprecated:: Foxy
+           The parameter `node_name` is deprecated, use `name` instead.
 
         :param name: The name of the lifecycle node.
           Although it defaults to None it is a required parameter and the default will be removed
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -37,7 +37,13 @@ from ..events.lifecycle import StateTransition
 class LifecycleNode(Node):
     """Action that executes a ROS lifecycle node."""
 
-    def __init__(self, *, name: Text = None, node_name: Optional[Text] = None, **kwargs) -> None:
+    def __init__(
+        self,
+        *,
+        name: Optional[Text] = None,
+        node_name: Optional[Text] = None,
+        **kwargs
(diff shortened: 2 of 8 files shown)


jacobperron:
I also just noticed that the [migration guide](https://index.ros.org/doc/ros2/Tutorials/Launch-files-migration-guide/#node) uses the attribute "name", which is currently inconsistent with the implementation.

launch_ros/launch_ros/actions/node.py
```diff
@@ -235,6 +237,11 @@ def parse(cls, entity: Entity, parser: Parser):
         if args is not None:
             kwargs['arguments'] = super()._parse_cmdline(args, parser)
         node_name = entity.get_attr('node-name', optional=True)
+        if node_name is not None:
+            warnings.warn(
+                "The attribute 'node-name' is deprecated. Use the attribute 'name' instead.")
+            kwargs['node_name'] = parser.parse_substitution(node_name)
+        node_name = entity.get_attr('name', optional=True)
```
hidmic: @jacobperron FYI `name` is also an attribute for plain processes. This change means it will no longer be possible to set both independently (it currently is possible in Python). Probably not the most relevant feature, though if node name is given a suffix to make it anon, that'll show in logs from now on.

hidmic:
> which is currently inconsistent with the implementation.

You're absolutely right, good catch.

launch_ros/launch_ros/actions/node.py
jacobperron: I'd overlooked that `name` was being used for ExecuteProcess. I've proposed a solution in 61503c9. We avoid forwarding the 'name' attribute to ExecuteProcess and expose a new attribute for node, 'exec-name', that keeps the feature of setting the process node.

launch_ros/launch_ros/actions/node.py
```diff
@@ -244,6 +244,13 @@ def parse(cls, entity: Entity, parser: Parser):
         node_name = entity.get_attr('name', optional=True)
         if node_name is not None:
             kwargs['node_name'] = parser.parse_substitution(node_name)
+            # Undo the parent action parsing 'name' as the process name
+            # For node actions, use 'exec-name' to rename the process
+            if 'name' in kwargs:
+                del kwargs['name']
+        exec_name = entity.get_attr('exec-name', optional=True)
+        if exec_name is not None:
+            kwargs['name'] = parser.parse_substitution(exec_name)
```
hidmic: @jacobperron hmm, while this is a fine patch, it grows the gap between Python and markup-based launch files. This is something we've discussed with @ivanpauno several times, how detrimental to standardization is the fact that launch frontend parsing support allows one to depart from the original action "signature". IMHO, we should apply this exact same change to the `Node` action constructor itself and then parsing function should just use it. WDYT?

ivanpauno: The problem with changing the `Node` signature is that it breaks API in a confusing way:
if somebody was using `name` to specify the executable name, they will have a surprise when that's now used as node name.
I think will should keep `XML` attributes as close as possible to `Python` constructor arguments.

hidmic: That is also true. But doing so in XML/YAML only but not in Python makes it worst IMO.

jacobperron: I agree that this change should happen in the Node constructor, keeping the signature the same as the front-end. I'll do this if there's a consensus to go ahead with this change.

> The problem with changing the Node signature is that it breaks API in a confusing way:
if somebody was using name to specify the executable name, they will have a surprise when that's now used as node name.

Yeah, this is a problem. IMO, it seems like a good thing to change long term. In the short-term, we can document the behavior change in the release notes and on whatever documentation exists for launch_ros (note, besides the API docstrings, I only found mention of the "name" attribute in the [migration guide](https://index.ros.org/doc/ros2/Tutorials/Launch-files-migration-guide/#node) and it is already assuming the behavior I expect).

I also understand if there's consensus to leave the implementation as-is in order to avoid breaking the existing behavior.

ivanpauno: I think that the API break is reasonable.
To avoid breaking user launch files, `node_name` can be deprecated with a warning.
The users that were using `name` will have to check the release notes to understand the problem (though I expect not much users using it).
We should also merge the argument name change together with a PR updating the release notes. If not, we will forget later.

If @wjwwood and @hidmic are also ok with breaking API, I think it's fine.

wjwwood: 👍 to breaking API to make it make sense, and 👍 for updating the documentation and release notes simultaneously.

jacobperron: Thanks for the input, all. I'm on it :sunglasses:

jacobperron:
@hidmic @ivanpauno I've reverted the 'name' overriding in the parser (45d4992) and added new Node parameters for `name` and `namespace` (827042b), deprecating the old parameters. I've also added a new Node param `exec_name` for setting the process name (e7e1325)

launch_ros/launch_ros/actions/node.py
```diff
@@ -61,8 +63,11 @@ def __init__(
         self, *,
         node_executable: SomeSubstitutionsType,
         package: Optional[SomeSubstitutionsType] = None,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: Optional[SomeSubstitutionsType] = '',
         node_name: Optional[SomeSubstitutionsType] = None,
         node_namespace: SomeSubstitutionsType = '',
```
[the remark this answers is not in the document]
ivanpauno: Maybe the note in the docstring it's just fine.

launch_ros/launch_ros/actions/node.py
```diff
@@ -115,9 +120,16 @@ def __init__(
...
+
         :param: node_executable the name of the executable to find if a package
             is provided or otherwise a path to the executable to run.
         :param: package the package in which the node executable can be found
+        :param: name the name of the node
+        :param: namespace the ROS namespace for this Node
+        :param: exec_name the label used to represent the process.
+            Defaults to the basename of node executable.
         :param: node_name the name of the node
         :param: node_namespace the ros namespace for this Node
```
ivanpauno: I would add a comment here indicating deprecation. So, it's more obvious for someone that jumped to the documentation of these arguments.

launch_ros/launch_ros/actions/node.py
```diff
@@ -135,8 +147,18 @@ def __init__(
         # The substitutions will get expanded when the action is executed.
         cmd += ['--ros-args']  # Prepend ros specific arguments with --ros-args flag
         if node_name is not None:
+            warnings.warn("The parameter 'node_name' is deprecated, use 'name' instead")
+            if name is None:
```
ivanpauno: Passing both `name` and `node_name` should be detected as an error IMO. I would add an else case.

launch_ros/launch_ros/actions/node.py
```diff
@@ -135,8 +147,18 @@ def __init__(
...
+                cmd += ['-r', LocalSubstitution(
+                    "ros_specific_arguments['name']", description='node name')]
+                name = node_name
+        if name is not None:
             cmd += ['-r', LocalSubstitution(
                 "ros_specific_arguments['name']", description='node name')]
+        if node_namespace:
+            warnings.warn("The parameter 'node_namespace' is deprecated, use 'namespace' instead")
+            if not namespace:
+                namespace = node_namespace
```
ivanpauno: same applies here

launch_ros/launch_ros/actions/node.py
```diff
@@ -237,13 +261,19 @@ def parse(cls, entity: Entity, parser: Parser):
         node_name = entity.get_attr('node-name', optional=True)
         if node_name is not None:
             kwargs['node_name'] = parser.parse_substitution(node_name)
+        node_name = entity.get_attr('name', optional=True)
+        if node_name is not None:
+            kwargs['name'] = parser.parse_substitution(node_name)
```
ivanpauno: If both `name` and `node_name` are specified, it should be detected as an error.

launch_ros/launch_ros/actions/node.py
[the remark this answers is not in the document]
jacobperron: I've added a note to the param docstrings, hopefully readers will notice one of the two notes, and/or the deprecation warning printed to the terminal. (e069970)

launch_ros/launch_ros/actions/node.py
jacobperron: Done e069970

launch_ros/launch_ros/actions/node.py
jacobperron: Raising an exception if this happens: fa4ab9f

launch_ros/launch_ros/actions/node.py
jacobperron: See fa4ab9f

launch_ros/launch_ros/actions/node.py
jacobperron: This is detected by fa4ab9f

jacobperron:
> It would be good to check if something has to be updated in demos/examples/tutorials.

Good idea, I'll open up PRs connecting to this for instances I find.

jacobperron:
`LifecylceNode` has a required keyword argument named `node_name`. I've just patched how it is passed to it's parent constructor (586e09c), but what do you think about also renaming it's argument to `name`? This would be a harder API break, since we would would have to introduce a new argument `name` that is required.

jacobperron:
Adding to Foxy release notes: https://github.com/ros2/ros2_documentation/pull/522

ivanpauno:
> LifecylceNode has a required keyword argument named node_name. I've just patched how it is passed to it's parent constructor (586e09c), but what do you think about also renaming it's argument to name? This would be a harder API break, since we would would have to introduce a new argument name that is required.

I think that if we're going to do the change, it makes sense to rename all the `node_*` of any action/description (`LifecycleNode`, `ComposableNodeContainer`, `ComposableNode`, ...).
To avoid that hard API break in that case, the following can be done:
```python3
  def __init__(self, *, name: Optional[Text] = None, node_name: Optional[Text] = None, **kwargs) -> None:
```

If `node_name` is passed, a warning is showed.
If both `name` and `node_name` are passed, an error is raised.
If only `name` is passed, we're happy.

In another release, `node_name` can be removed and `name` be required.

jacobperron:
> If node_name is passed, a warning is showed.
If both name and node_name are passed, an error is raised.
If only name is passed, we're happy.

Also, if both are `None`, an error raised. I don't think the `LifecycleNode` action will work otherwise.

I can do this.

jacobperron:
I've also deprecated the node_name parameter for LifecycleNode (09aa4c5) and updated the connected demos PR.

jacobperron:
And I've done the same for ComposableNodeContainer (4f19781).

jacobperron:
I've created a ticket (https://github.com/ros2/launch_ros/issues/123) so we don't forget to remove the deprecated API post-Foxy.

launch_ros/launch_ros/actions/node.py
```diff
@@ -61,8 +63,11 @@ def __init__(
         self, *,
         node_executable: SomeSubstitutionsType,
```
ivanpauno: Maybe, change this one to executable too?

jacobperron: Sure, I don't mind doing this, but it will trigger a lot more changes on dependent packages.

Also, I noticed that the front-end uses slight different names for this and for the package name: `exec` and `pkg`. I don't think anything needs to be done about it, but just thought I'd point it out since the discrepancies between Python and the front-end are what motivated the current changes I'm making.

jacobperron:
Updated ComposableNode (5849c05) and the demos PR.

launch_ros/launch_ros/actions/node.py
ivanpauno: > Sure, I don't mind doing this, but it will trigger a lot more changes on dependent packages.

> Also, I noticed that the front-end uses slight different names for this and for the package name: exec and pkg. I don't think anything needs to be done about it, but just thought I'd point it out since the discrepancies between Python and the front-end are what motivated the current changes I'm making.

Yeah, that's true. I think that have some arguments prefixed with `node_` and some others not, is strange. I would avoid the `node_` prefix everywhere, though it can be split in different PRs to avoid handling an annoing number of PRs at the same time.

About the difference between the frontend and python, I prefer the short versions for the frontend (`pkg`, `exec`), as the other ones are much more verbose. I also think that the current Python names are ok.

jacobperron: I'll open a follow-up set of PRs deprecating `node_executable`.

launch_ros/launch_ros/actions/composable_node_container.py
```diff
@@ -34,8 +35,10 @@ class ComposableNodeContainer(Node):
     def __init__(
         self,
         *,
-        node_name: SomeSubstitutionsType,
-        node_namespace: SomeSubstitutionsType,
+        name: SomeSubstitutionsType = None,
+        namespace: SomeSubstitutionsType = None,
```
hidmic: @jacobperron nit: if these have `None` as default, type annotation should read `Optional[SomeSubstitutionsType]`

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -36,7 +37,7 @@
 class LifecycleNode(Node):
     """Action that executes a ROS lifecycle node."""
 
-    def __init__(self, *, node_name: Text, **kwargs) -> None:
+    def __init__(self, *, name: Text = None, node_name: Optional[Text] = None, **kwargs) -> None:
```
hidmic: @jacobperron same about type annotation.

launch_ros/launch_ros/actions/composable_node_container.py
```diff
@@ -45,12 +48,35 @@ def __init__(
         Most arguments are forwarded to :class:`launch_ros.actions.Node`, so see the documentation
         of that class for further details.
 
-        :param: node_name the name of the node, mandatory for full container node name resolution
-        :param: node_namespace the ros namespace for this Node, mandatory for full container node
-           name resolution
+        .. warning:: Parameters `node_name` and `node_namespace` are deprecated.
```
hidmic: @jacobperron BTW, it turns out there's a [`deprecated` directive](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-deprecated) we can use.

launch_ros/launch_ros/actions/composable_node_container.py
jacobperron: Fixed 7c00297

launch_ros/launch_ros/actions/lifecycle_node.py
jacobperron: 7c00297

launch_ros/launch_ros/actions/composable_node_container.py
jacobperron: Using deprecated directive 50395ef
I wasn't sure what version number to put, so I used "Foxy".

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9356/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5040/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7663/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9277/)
* Windows-container [Build Status](http://ci.ros2.org/job/ci_windows-container/144/)

test_launch_ros/.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
hidmic: @jacobperron hmm do we need to keep `.eggs` around?

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -191,6 +192,13 @@ def test_create_node_with_invalid_parameters(self):
         )
         self._assert_launch_no_errors([node_action])
 
+    def test_deprecated_node_parameters(self):
+        node_action = launch_ros.actions.Node(
+            package='demo_nodes_py', node_executable='talker_qos', output='screen',
+            node_name='my_node', node_namespace='my_ns'
+        )
+        self._assert_launch_no_errors([node_action])
+
```
hidmic: @jacobperron should we also ensure that it does throw if both `name` and `node_name` are provided? Same for namespace.

test_launch_ros/.eggs/README.txt
jacobperron: Oops, removed (a836234)

test_launch_ros/test/test_launch_ros/actions/test_node.py
jacobperron: Added tests c19dc08

ivanpauno:
@jacobperron is this ready to be merged?

jacobperron:
I was hesitating due to the CI failures. I've retriggered CI.

nuclearsandwich:
@jacobperron I haven't completed the first CI test but in local research https://github.com/ros2/ci/pull/399 is a workaround for the installation paths which should allow the tests to run.

launch_ros/launch_ros/actions/node.py
jacobperron: I finally got around to doing this. Here's the PR: https://github.com/ros2/launch_ros/pull/140

Issue #120, opened 2020-01-30
Revisit injection of launch_ros description


jacobperron:
#### Feature description

This issue was raised as part of https://github.com/ros2/launch_ros/issues/97 and previously discussed in https://github.com/ros2/launch/issues/208.

Both `ros2 launch` and `ros2 test` [inject an included description](https://github.com/ros2/launch_ros/blob/c5e2aa2bf0eb60d61c66dbe0a6da365230b3387d/ros2launch/ros2launch/api/api.py#L146-L148) that creates an rclpy node. This node is often useful for testing, though not always needed, and necessary to launch certain actions (e.g. `LifecycleNode`). The problem is that it can cause confusion and user error.

Related to testing, we end up with two types of launch test files: those that can be executed with `launch_test` and those that require using `ros2 test`. As evidenced by https://github.com/ros2/launch_ros/issues/97, this can lead to seemingly broken launch files, which are actually just user error (executing with `launch_test` instead of `ros2 test`).

Related to using `ros2 launch`, if a user decides to include the default description themselves (perhaps a rare case), they will end up starting multiple nodes with the same name. This currently results in the following warning, but may be escalated to an error in the future:

```
[WARN] [rcl.logging_rosout]: Publisher already registered for provided node name. If this is due to multiple nodes with the same name then all logs for that logger name will go out over the existing publisher. As soon as any node with that name is destructed it will unregister the publisher, preventing any further logs for that name from being published on the rosout topic.
```

I've opened this ticket to discuss alternatives to relying on the CLI tools for including launch_ros description.

#### Implementation considerations

Proposed solutions:

1. Remove the injection
  - In this case, we rely on the users to provide the boiler plate for including the ROS-specific description if they need it. [Example](https://github.com/ros2/launch_ros/blob/2cedc8b798fe26d363be239c9e7312fe0154b749/launch_testing_ros/test/examples/talker_listener_launch_test.py#L54).
  - **Pros**:
    - rostests can be run with `launch_test` or `ros2 test`.
    - Helps avoid accidentally including the default description more than once.
  - **Cons**:
    - Extra boilerplate for the user.
    - Requires knowledge of which launch_ros actions require the inclusion of the default description.
2. Only inject when necessary
  - Somehow know when an action requires the extra node provided by the default description and conditionally include it. This would require knowing when the description is already included so it is not included multiple times.
  - **Pros**:
    - Same as (1).
  - **Cons**:
    - Added complexity to launch and/or launch_ros. 

[...]

PR #128, opened 2020-03-10
Enable implicit ROS startup by launch_ros actions. 

hidmic:
Addresses #120. This pull request drops the concept of a default launch description in lieu of a so called ROS adapter accessible from and managed by a launch context. Actions that make use of it can trigger implicit and idempotent instantiation of such adapter, though a user may do so explicitly.


launch_ros/launch_ros/actions/explicit_ros_startup.py
```diff
@@ -0,0 +1,60 @@
...
+from typing import List
+from typing import Optional
+
+import launch.actions
+import launch_ros.ros_adapters
+
+import rclpy
+
+
+class ExplicitROSStartup(launch.actions.OpaqueFunction):
```
jacobperron: I'm guessing that the purpose of this action is to allow users to initialize the ROS adapter. But, I don't think we should encourage users to use the launch ROS context (or node) directly. It is more of an implementation detail of launch_ros. It could also lead to misuse (e.g. I can imagine a user passing the node to their own executor).

I think it makes more sense for users to create their own nodes and manage their own contexts if they need them, and it doesn't involve too much overhead.

Maybe someone else can weigh in here.

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -158,14 +160,15 @@ def execute(self, context: launch.LaunchContext) -> Optional[List[Action]]:
         self._perform_substitutions(context)  # ensure self.node_name is expanded
         if '<node_name_unspecified>' in self.node_name:
             raise RuntimeError('node_name unexpectedly incomplete for lifecycle node')
+        node = get_ros_node(context)
         # Create a subscription to monitor the state changes of the subprocess.
-        self.__rclpy_subscription = context.locals.launch_ros_node.create_subscription(
+        self.__rclpy_subscription = node.create_subscription(
```
ivanpauno: What if `ExplicitROSStartup` wasn't executed before?

launch_ros/launch_ros/actions/explicit_ros_startup.py
wjwwood: 👍, keep the launch_ros context and node hidden, let the user create their own if they want to use ROS for something.

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+
+    def start(self):
+        """Start ROS adapter."""
+        if self.__is_running:
+            raise RuntimeError('Cannot start a ROS adapter that is already running')
+        try:
+            if self.__context is not None:
+                rclpy.init(args=self.__argv, context=self.__context)
+        except RuntimeError as exc:
+            if 'rcl_init called while already initialized' in str(exc):
```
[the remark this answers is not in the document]
hidmic: This logic predates this PR (see [here](https://github.com/ros2/launch_ros/blob/d10e4429563920fe101c4ce721f53eb3f1fe3b05/launch_ros/launch_ros/default_launch_description.py#L63)). I can only presume this was here to cope with multiple attempts to initialize the default context, but you have a great point about bad use of `pass` and so this likely wasn't the original author's intent. 

I'd be OK with removing this altogether. WDYT?

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+
+    If no adapter is found, one will be created.
+    """
+    if not hasattr(context.locals, 'ros_adapter'):
+        ros_adapter = ROSAdapter()
+        context.extend_globals({'ros_adapter': ros_adapter})
+        context.register_event_handler(launch.event_handlers.OnShutdown(
+            on_shutdown=lambda *args, **kwargs: ros_adapter.shutdown()
+        ))
+    return context.locals['ros_adapter']
```
wjwwood: Should we consider making this thread-safe? I'm just imaging some action creating a thread and trying to access this from the thread and another action accessing it from the asyncio run loop in a different thread, at the same time.

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+
+        if autostart:
+            self.start()
+
+    def start(self):
+        """Start ROS adapter."""
+        if self.__is_running:
+            raise RuntimeError('Cannot start a ROS adapter that is already running')
+        try:
+            if self.__context is not None:
```
wjwwood: I think that if context is `None`, then this action should create and manager its own context, rather than passing the `None` on to `create_node` so that the default singleton is used.

Imagine:

- user does rclpy.init()
- this function trys that but catches the error that it is already initialized
- this function creates node with `context=None` (use singleton context)
- user does rclpy.shutdown()
- this class or an action using this class trys to do something with the shutdown context

By managing its own context, nothing the user does will affect it.

launch_ros/launch_ros/actions/explicit_ros_startup.py
hidmic: I can see how this could lead to misuse, yeah. But on the other hand, we'd be disallowing any form of ROS context configuration, which was previously possible via [`ROSSpecificLaunchStartup`](https://github.com/ros2/launch_ros/blob/d10e4429563920fe101c4ce721f53eb3f1fe3b05/launch_ros/launch_ros/default_launch_description.py#L28). So I'd be personally inclined towards simply documenting this properly. 

One last thought. It may prove useful to allow users to create nodes and have launch servicing them and handling their lifetime.

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+        autostart: bool = True
+    ):
+        """
+        Construct adapter.
+
+        :param: argv List of global arguments for rclpy context initialization.
+        :param: context Provide a context other than the default rclpy context
+          to pass down to rclpy.init.
+          The context is expected to have already been initialized by the caller
+          using rclpy.init().
```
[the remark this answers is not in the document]
hidmic: This is something I wanted to discuss. We were already inconsistent about initializing / shutting down contexts. I think it'd be better to own these contexts and their lifetime, and always initialize them and shut them down unless explicitly stated otherwise. Thoughts?

launch_ros/launch_ros/actions/explicit_ros_startup.py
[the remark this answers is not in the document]
hidmic: Mimic'd [`ROSSpecificLaunchStartup`](https://github.com/ros2/launch_ros/blob/d10e4429563920fe101c4ce721f53eb3f1fe3b05/launch_ros/launch_ros/default_launch_description.py#L28) structure. Don't mind either way.

launch_ros/launch_ros/actions/lifecycle_node.py
hidmic: Then [`get_ros_node`](https://github.com/ros2/launch_ros/pull/128/files/2591ad1354db8313905c08b7fe4b8dade2f41f01#diff-81308dd39f1a305c2cd555c95657491dR163) will ensure `node` comes into existence.

launch_ros/launch_ros/ros_adapters.py
hidmic: Hmm, I had the impression `launch` wasn't thread-safe i.e. that using the same launch context across `asyncio` loops in separate threads wasn't supported. We can lock it otherwise.

launch_ros/launch_ros/ros_adapters.py
hidmic: That makes sense, and goes hand-in-hand with `rclpy` contexts used by `launch` should have their lifetime handled by `launch`.

launch_ros/launch_ros/ros_adapters.py
wjwwood: Parts of it are, if we think a function may be called outside of the asyncio loop (like from a user thread) then it should be thread-safe. I'm not talking about calling the run loop for asyncio in multiple threads concurrently, that is definitely not supported.

launch_ros/launch_ros/actions/explicit_ros_startup.py
wjwwood: > One last thought. It may prove useful to allow users to create nodes and have launch servicing them and handling their lifetime.

We could provide that, but I think that's a different use case than this class. This class is useful (imo) for other actions to build on. If the we wanted to share something with the user and manage the lifetime for them, then we need a way to know when they are done with it and a way to give them access to the node and context we created. Though there is a lot of overlap, it seems like a different use case to me.

wjwwood: Having the ability to configure the context seems compelling to me, but I wonder if there are other ways to allow that without exposing this to the user.

launch_ros/launch_ros/ros_adapters.py
hidmic: Thinking about this again, do we even need to allow for a context to be passed along? Without an `ExplicitROSStartup` action, users can't access this class' instance. Arguments can already be provided to initialize it and context is accessible through a property. So if you don't disagree, I'll just drop that.

launch_ros/launch_ros/ros_adapters.py
hidmic: I see. The thing is that something like:

> I'm just imaging some action creating a thread and trying to access this from the thread and another action accessing it from the asyncio run loop in a different thread, at the same time.

won't work properly because `launch.Context` itself isn't thread-safe e.g. one cannot emit an event or affect context globals or locals in a thread-safe manner. So I'd rather push for that in a follow-up PR i.e. make `launch.Context` thread-safe and then make this function thread-safe. WDYT?

launch_ros/launch_ros/actions/explicit_ros_startup.py
hidmic: Following @wjwwood rationale, if users won't be making use of this class for their own purposes then there's no need for the context to be configurable.

ros2launch/ros2launch/api/api.py
```diff
@@ -26,8 +26,7 @@
 from launch.frontend import Parser
 from launch.launch_description_sources import get_launch_description_from_any_launch_file
 import launch_ros
-import rclpy
-import rclpy.context
+import launch_ros.actions
```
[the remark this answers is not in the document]
hidmic: Good catch. Dropped in 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/ros_adapters.py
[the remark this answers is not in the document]
hidmic: Dropped the logic in 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/ros_adapters.py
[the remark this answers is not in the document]
hidmic: A context of its own is now instantiated for the adapter in 577d77bfc5fa2101616ce7856b6d55745370692b. Its lifetime is now entirely handled by the adapter.

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+        *,
+        argv: Optional[List[str]] = None,
+        context: Optional[rclpy.Context] = None,
+        autostart: bool = True
+    ):
+        """
+        Construct adapter.
+
+        :param: argv List of global arguments for rclpy context initialization.
+        :param: context Provide a context other than the default rclpy context
```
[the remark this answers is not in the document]
hidmic: I like `ros_context` :). See 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/ros_adapters.py
```diff
@@ -0,0 +1,138 @@
...
+        self.__node.destroy_node()
+        rclpy.shutdown(context=self.__context)
+        self.__is_running = False
+
+    @property
+    def argv(self):
+        return self.__argv
+
+    @property
+    def context(self):
```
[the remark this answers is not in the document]
hidmic: Done in 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/actions/explicit_ros_startup.py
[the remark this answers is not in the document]
hidmic: Action dropped in 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/ros_adapters.py
hidmic: Done in 577d77bfc5fa2101616ce7856b6d55745370692b.

launch_ros/launch_ros/ros_adapters.py
wjwwood: The context is not thread-safe, but things that use it are, e.g.: https://github.com/ros2/launch/blob/0d040023306843d71b6316b160fab7b219f41892/launch/launch/launch_service.py#L124-L144

I'd say I agree with you if the user or other actions were accessing the context directly, but they're not, they are calling this function which happens to use the context as an implementation detail.

I think it should still be thread-safe to be honest.

launch_ros/launch_ros/ros_adapters.py
wjwwood: I don't think we need to, but having the ability to doesn't cost anything, imo.

launch_ros/launch_ros/ros_adapters.py
hidmic: Hmm, I think `launch.LaunchService` case is different, as that's API for `launch` client code as opposed to API used within `launch.Action` subclasses, where pretty much everything is not thread-safe. 

But forget all that, how would you make these functions thread-safe without doing the same for `launch.LaunchContext`? I don't think a global lock for these functions is a proper solution, because (a) two threads using two different `launch.LaunchContext` instances may contend with each other for no reason, and (b) `launch.LaunchContext` global dict itself isn't thread-safe (which may be solved with such a lock, yes, but if and only if no other API that could potentially run concurrently attempts to do the same).

launch_ros/launch_ros/ros_adapters.py
hidmic: Rationale for dropping it was that it'd require additional logic to decide what should handle that context's lifetime, and for no reason since we've agreed these adapters are for actions to build on and not for users to play with.

hidmic:
CI up to `test_launch_ros` and `test_communication` to put this to test:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9618/) (canary build)

launch_testing_ros/launch_testing_ros/test_runner.py
```diff
@@ -19,19 +19,4 @@
 
 
 class LaunchTestRunner(launch_testing.test_runner.LaunchTestRunner):
```
ivanpauno: We should consider deprecating launch_testing_ros

hidmic: I thought about it, but I chose not to do that now both to keep this PR small and to not take that leap and remove a package we may need again afterwards.

launch_ros/launch_ros/ros_adapters.py
wjwwood: Why not just a lock in this function? Either a static (global) one or a one owned by this class?

I'm not trying to thread-safe the context, but I do expect other actions using the node or context to access it through this class's methods.

launch_ros/launch_ros/ros_adapters.py
wjwwood: That's fine with me to leave it as-is.

launch_ros/launch_ros/ros_adapters.py
hidmic: > Why not just a lock in this function? Either a static (global) one or a one owned by this class?

Because of the reasons exposed above.

hidmic:
Had to rebase to solve conflicts, sorry folks!

launch_ros/launch_ros/ros_adapters.py
wjwwood: Sorry I'll try to address the two points directly.

> (a) two threads using two different launch.LaunchContext instances may contend with each other for no reason

Is that actually a problem? The scope and time bounds on this function are straightforward, so I don't think there's a deadlock problem. Are you concerned about performance due to lock contention?

> (b) launch.LaunchContext global dict itself isn't thread-safe (which may be solved with such a lock, yes, but if and only if no other API that could potentially run concurrently attempts to do the same).

Of course if someone trys to access this fields in the context directly it may be an issue, but I thought the idea was to always access this field of the context (`context.locals.ros_adapter`) via this method right?

Do you mean issues when someone asynchronously calls this function while someone else concurrently accesses a _different_ field in the context?

---

I'm happy to drop this if you don't think it's a concern, my concern is definitely academic for the moment.

hidmic: > Are you concerned about performance due to lock contention?

A bit of that, a bit of OCD :)

> Do you mean issues when someone asynchronously calls this function while someone else concurrently accesses a different field in the context?

Precisely. In particular, `launch.LaunchContext.extend_locals()` and `launch.LaunchContext.extend_globals()` do not seem safe to use concurrently as per this [Python 3.8 FAQ](https://docs.python.org/3.8/faq/library.html#what-kinds-of-global-value-mutation-are-thread-safe) (see disclaimer at the bottom, due to `__del__` being called as a side effect of `dict.update`). I give you that the chance of actually hitting a problem is (likely very) low though.

----

If you think there's no risk of this backfiring in the future, I'll add that lock.

wjwwood: We can leave it out. It would be nice to mention it explicitly in the docstring, as I feel this class is more likely to be used concurrently (like in the callback of a subscription or something) than other context functions.

hidmic: Done in 63ff30f88337fe90acbb114063bd9ed835203b7f.

hidmic:
Alright, going in! Thank you all for the reviews!

launch_testing_ros/launch_testing_ros/test_runner.py
```diff
@@ -19,19 +19,4 @@
...
-
-    def generate_preamble(self):
-        import launch_ros  # Import on first use to avoid races at module level
-        return [launch.actions.IncludeLaunchDescription(
-            launch_description_source=launch.LaunchDescriptionSource(
-                launch_description=launch_ros.get_default_launch_description(
-                    rclpy_context=self._rclpy_context,
-                )
-            )
-        )]
```
dirk-thomas: This removed led to a new flake8 warning: https://ci.ros2.org/view/nightly/job/nightly_linux_debug/lastCompletedBuild/testReport/launch_testing_ros.test/test_flake8/test_flake8/

> `.\launch_testing_ros\test_runner.py:17:1: F401 'launch' imported but unused`

hidmic: Argh, I'll PR a fix ASAP.

Issue #136, opened 2020-02-05
Command substitution: special characters inside XML comments causes launch to fail


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - from source 
- Version or commit hash:
  - 4333910ba41977f3ab9867082c86a8698ada3a16 

Specifically, I've noticed that colons (`:`) and tab characters are not working.

#### Steps to reproduce issue

I ran into this issue trying to pass a URDF from xacro to robot_state_publisher, like the [example here](https://github.com/ros/robot_state_publisher/blob/f877cbf4399d8c497b946376664bd76b0c192920/launch/rsp-launch-xacro-command-subst.xml). For some reason, if the URDF/xacro file contains a colon (`:`) or a tab character inside of an XML comment, then launch fails.

Here's my example URDF file `foo.urdf`:

```xml
<?xml version="1.0" ?>
<robot>
  <!--: -->
</robot>
```

And my launch file:

```xml
<launch>
  <node pkg="demo_nodes_cpp" exec="talker" output="screen">
    <param name="foo" value="$(command 'cat foo.urdf')" />
  </node>
</launch>
```

#### Expected behavior

The launch file succeeds and we see the talker node chatting.

#### Actual behavior

Error:

```
[ERROR] [launch]: Caught exception in launch (see debug for traceback): mapping values are not allowed here
  in "<unicode string>", line 3, column 7:
      <!--: -->
          ^
```

#### Additional information

The error seems to specifically occur if there is a colon followed by a space. E.g. making the following change works fine:

```diff
-      <!--: -->
+      <!--:-->
```

Replacing the colcon with a tab character also triggers the same error:

```
[ERROR] [launch]: Caught exception in launch (see debug for traceback): while scanning for the next token
found character '\t' that cannot start any token
  in "<unicode string>", line 3, column 7:
      <!--	-->
          ^
```

ivanpauno:
This can be fixed now by avoiding using `yaml` conversions:

```xml
<launch>
  <node pkg="demo_nodes_cpp" exec="talker" output="screen">
    <param name="foo" value="$(command 'cat foo.urdf')" type="str"/>
  </node>
</launch>
```

The fix was introduced in https://github.com/ros2/launch_ros/pull/137.

karl-schulz:
Is there also a fix available for python launch files? Having special characters like `:` or `#` still breaks the parameter parsing in that case...

Example
```
Node(
[...]

Issue #74, opened 2019-09-19
Bad interaction between launch frontend and YAML parameters


hidmic:
Both launch XML and YAML frontends currently coexist with the YAML parsing of substitution text in `Node` parameters. This has caused us issues in the past (e.g. https://github.com/ros2/launch/pull/226#discussion_r300031818), forcing additional quotes as escape mechanisms. At the time we deemed this as a situation we'd rarely find in the field.

-------

Well,  as you can see in https://github.com/RobotWebTools/rosbridge_suite/pull/433, specifically on commit https://github.com/RobotWebTools/rosbridge_suite/pull/433/commits/d466c89475a068acd3e90a286d633c75d611204c, it's not as rare as we thought. 

There, **I'm forced to surround a `*` with two `'` on each side**.

- **Why a `*` needs this?** Because it's later used as a `Node` parameter and parameters are parsed as YAML, that parses that `*` as a malformed alias (see [PyYAML documentation](https://pyyaml.org/wiki/PyYAMLDocumentation) about aliases). 

- **Why two `'`?** The inner pair prevents `*` from being parsed as a YAML alias, and the outer pair is removed by the frontend for consistency -- otherwise one has to remember that some values given in e.g. XML are parsed using YAML under the hood and some are not.  

IMHO the deeper, unresolved issue that https://github.com/ros2/launch_ros/pull/31 partially addressed is that Substitutions in `launch` are always text and it is up to the receiving end to do any additional parsing, opening the door to things like this.

dirk-thomas:
Since this affects how you have to write a launch file I would think this needs to be fixes asap before more packages start using this.

ivanpauno:
A think we should have a custom class for a parameter description, instead of using a native type to represent them. There are a bunch of reasons:
- Conditional inclusion of parameters isn't now possible.
  As I commented [here](https://github.com/ros2/launch/issues/290#issuecomment-520823768) it would be much easier if we have a custom class for that.
- If we would have a class that could be constructed like:
  `Parameter('name', VALUE_HERE, type=int)`
  We could avoid all this problem.

Also, we currently don't provide a way of coercing the result of a substitution to another thing that's not an string. So each class is doing a custom thing for doing this conversions (e.g.: for converting from `str` to `bool`). I did a comment about "typed substitutions" in https://github.com/ros2/ros2_documentation/pull/302#discussion_r306325733, though that's not the only solution.

ivanpauno:
[...]

PR #137, opened 2020-03-24
Fix problems when parsing a `Command` `Substitution` as a parameter value

ivanpauno:
Trying to address https://github.com/ros2/launch_ros/issues/136.
Related with https://github.com/ros2/launch_ros/issues/74.

Since https://github.com/ros2/launch_ros/pull/31, substitutions in parameter values are loaded like `yaml`. This is an issue with the `Command` substitution, because it is not possible to skip the `yaml.load()` done by the `Node` action when parsing the input parameters.

My proposal here is to add a `Parameter` class, which allows to specify the type of the parameter.
One of the options is to convert the result of the substitution like `yaml`.

Other things I would like to revisit after this:
- Add a type attribute in launch frontend parameters, map them to the new API.
- Maybe stop supporting to pass a dictionary in `parameters` Node argument. Or maybe, don't allow substitutions in them (How to tic-toc?).
- Consolidate `type_utils` here with `type_utils` used in `launch.frontend`.
- Drop support of union of types in `launch.frontend.type_utils`. Update parsing functions if necessary.
- Revisit lists in XML frontend. Drop `*-sep` and support list written like in `yaml`?


test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -134,10 +138,56 @@ def test_launch_node_with_parameter_files(self):
...
+        node_action = self._create_node(
+            parameters=[
+                Parameter(
+                    name='param1',
+                    value=EnvironmentVariable(name='PARAM1_VALUE'),
+                    value_type=str,
+                ),
+                Parameter(
+                    name=EnvironmentVariable(name='PARAM2'),
+                    value=[[EnvironmentVariable(name='PARAM2')],'_value'],
```
wjwwood:
```suggestion
                    value=[[EnvironmentVariable(name='PARAM2')], '_value'],
```

launch_ros/launch_ros/parameter.py
```diff
@@ -0,0 +1,163 @@
...
+
+class Parameter:
+    """Describes a ROS Parameter."""
+
+    def __init__(
+        self,
+        *,
+        name: SomeSubstitutionsType,
+        value: 'SomeParameterValue',
+        value_type: Any = None
```
wjwwood: Seems there are very well defined set of types it can be below, so is `Any` the right thing to use here?

wjwwood:
> Maybe stop supporting to pass a dictionary in parameters Node argument. Or maybe, don't allow substitutions in them (How to tic-toc?).

My instinct is to leave them, but I don't know for sure.

test_launch_ros/test/test_launch_ros/utilities/test_type_utils.py
```diff
@@ -0,0 +1,149 @@
...
+    assert coerce_to_type('off') is False
+    assert coerce_to_type('True') is True
+
+    assert coerce_to_type('[2, 1, 1]') == [2, 1, 1]
+    assert coerce_to_type('[.2, .1, .1]') == [.2, .1, .1]
+    assert coerce_to_type('[asd, bsd, csd]') == ['asd', 'bsd', 'csd']
+    assert coerce_to_type('[on, false, no]') == [True, False, False]
+
+
+def test_coercions_given_specific_type():
```
hidmic, line 51: @ivanpauno consider checking that coercion to a type different from the one that results of using `data_type=None` works e.g. ` assert coerce_to_type("1", data_type=str) == '1'`.

ivanpauno:
Rebased with master, I'm working in addressing comments.

ivanpauno:
I've created https://github.com/ros2/launch/pull/438, to consolidate all `type_utils` in `launch` repo.
I still have to polish things up, but it's ready for a review.

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -160,10 +211,12 @@ def test_launch_node_with_parameter_dict(self):
...
-        expanded_parameter_files = node_action._Node__expanded_parameter_files
-        assert len(expanded_parameter_files) == 1
-        with open(expanded_parameter_files[0], 'r') as h:
-            expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
+        expanded_parameter_arguments = node_action._Node__expanded_parameter_arguments
+        assert len(expanded_parameter_arguments) == 1
+        file_path, is_file = expanded_parameter_arguments[0]
+        assert is_file
+        with open(file_path, 'r') as h:
+            expanded_parameters_dict = yaml.load(h)
```
hidmic, line 219: @ivanpauno why remove the `Loader=yaml.FullLoader` kwarg?

ivanpauno: 1. I have deleted that by mistake.
2. We should not been using it ...
  The rcl yaml parameter parser cannot handle `!!python/tuple` AFAIR.
  We were using this loader because we're dumping parameters arrays as tuples and not as lists. We should be doing the later.

I will fix `1.`. We should fix `2.` in a follow up (I have to confirm this is actually a problem).

launch_ros/launch_ros/actions/node.py
```diff
@@ -225,27 +231,28 @@ def __init__(
...
+
         def get_nested_dictionary_from_nested_key_value_pairs(params):
             """Convert nested params in a nested dictionary."""
             param_dict = {}
             for param in params:
                 name = tuple(parser.parse_substitution(param.get_attr('name')))
-                value = param.get_attr('value', data_type=None, optional=True)
+                type_identifier = param.get_attr('type', data_type=None, optional=True)
+                data_type = None
+                if data_type is not None:
```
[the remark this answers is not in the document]
ivanpauno: Yeap, thanks for catching it.
See https://github.com/ros2/launch_ros/pull/137/commits/27633f82c03ec59eb59398015c3b052a93f5a874.

test_launch_ros/test/test_launch_ros/actions/test_node.py
ivanpauno: Done in https://github.com/ros2/launch_ros/pull/137/commits/ef5209c49295b281b3039af9639734d1a1023d64.

hidmic:
I have to review this, and we need to decide what to do with https://github.com/ros2/launch_ros/blob/1f0ea9f6cb4193e1e2618f9dc8f0c730f547fb44/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py#L50

ivanpauno:
> I have to review this, and we need to decide what to do with
> 
> https://github.com/ros2/launch_ros/blob/1f0ea9f6cb4193e1e2618f9dc8f0c730f547fb44/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py#L50

Those "qoute" can be avoided by explicitly passing the type.

hidmic:
> Those "qoute" can be avoided by explicitly passing the type.

Yes, that's what I thought. Though I wonder if we should keep that test case to make sure XML special characters work.

launch_ros/launch_ros/parameter_descriptions.py
```diff
@@ -0,0 +1,156 @@
...
+    ) -> None:
+        """
+        Construct a parameter value description.
+
+        :param value: Value or substitution that can be resolved to a value.
+        :param value_type: Used when `value` is a substitution, to coerce the result.
+            Can be one of:
+                - Scalar types: int, bool, float, str
+                - List types: List[int], List[bool], List[float], List[str]
+                - None: Will use yaml rules, and check that the result is one of the above.
```
hidmic, line 57: @ivanpauno nit^2:
```suggestion
                - None: will use yaml rules, and check that the result matches one of the above.
```

test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py
```diff
@@ -47,13 +48,13 @@ def test_launch_remapping_xml():
...
-                    <!-- TODO(hidmic): revisit after https://github.com/ros2/launch_ros/pull/137 gets merged -->
                     <param name="param8" value="&quot;'2'&quot;, &quot;'5'&quot;, &quot;'8'&quot;" value-sep=", "/>
                     <param name="param9" value="\'2\', \'5\', \'8\'" value-sep=", "/>
-                    <!-- TODO(hidmic): revisit after https://github.com/ros2/launch_ros/pull/137 gets merged -->
                     <param name="param10" value="&quot;'asd'&quot;, &quot;'bsd'&quot;, &quot;'csd'&quot;" value-sep=", "/>
                     <param name="param11" value="'\asd', '\bsd', '\csd'" value-sep=", "/>
-                    <param name="param12" value=""/>
+                    <param name="param12" value="''"/>
+                    <param name="param13" value="$(var my_value)" type="str"/>
+                    <param name="param14" value="'2', '5', '8'" value-sep=", " type="list_of_str"/>
```
hidmic, line 57: @ivanpauno should it be 
```suggestion
                    <param name="param14" value="2, 5, 8" value-sep=", " type="list_of_str"/>
```
instead, to actually make sure YAML rules won't get in the way? Same below.

ivanpauno: the expected output is `["'2'", "'5'", "'8'"]`, this is checking that YAML rules aren't used in the way.

ivanpauno:
> Yes, that's what I thought. Though I wonder if we should keep that test case to make sure XML special characters work.

That's still being tested, I've only deleted the TODOs.

ivanpauno:
I've to re-release launch to make the Rpr checker happy.

test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py
hidmic: Checking that YAML rules are not used to parse a list, but not checking that YAML rules are not used for each element.

ivanpauno: > Checking that YAML rules are not used to parse a list, but not checking that YAML rules are not used for each element.

The first element is `"'2'"`. If yaml rules were being used, the output would be `'2'`. This is checking that the output is `"'2'"`.
If I do the change you ask the test doesn't pass.

hidmic: > If I make the change you ask for, the test won't pass.

How so?

ivanpauno:

ivanpauno:

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11762/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6837/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9595/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11703/)

PR #141, opened 2020-04-15
avoid deprecation warning, use from_parent

dirk-thomas:
Same as ros2/launch#402.


dirk-thomas:
Not needed - changing ros2/launch#402 instead.

PR #153, opened 2020-06-19
Fix no specified namespace

ivanpauno:
If the user doesn't explicitly pass a namespace to `Node` action neither uses `PushRosNamespace`, the node namespace can't be known at launch time neither the fully qualified node name:

https://github.com/ros2/launch_ros/blob/7974bd4baa545dc821b5afbda9c82fbc74647df7/launch_ros/launch_ros/actions/node.py#L355-L358

This PR fixes that incorrect assumption.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -61,6 +61,9 @@ import yaml
 class Node(ExecuteProcess):
     """Action that executes a ROS node."""
 
+    UNSPECIFIED_NODE_NAME = '<node_name_unspecified>'
+    UNSPECIFIED_NODE_NAMESPACE = '<node_namespace_unspecified>'
+
     def __init__(
         self, *,
         executable: Optional[SomeSubstitutionsType] = None,
@@ -215,7 +218,7 @@ class Node(ExecuteProcess):
         self.__remappings = [] if remappings is None else remappings
         self.__arguments = arguments
 
-        self.__expanded_node_name = '<node_name_unspecified>'
+        self.__expanded_node_name = self.UNSPECIFIED_NODE_NAME
         self.__expanded_node_namespace = ''
         self.__final_node_name = None  # type: Optional[Text]
         self.__expanded_parameter_files = None  # type: Optional[List[Text]]
@@ -321,6 +324,10 @@ class Node(ExecuteProcess):
             raise RuntimeError("cannot access 'node_name' before executing action")
         return self.__final_node_name
 
+    def is_node_name_fully_specified(self):
+        keywords = (self.UNSPECIFIED_NODE_NAME, self.UNSPECIFIED_NODE_NAMESPACE)
+        return all(x not in self.node_name for x in keywords)
+
     def _create_params_file_from_dict(self, params):
         with NamedTemporaryFile(mode='w', prefix='launch_params_', delete=False) as h:
             param_file_path = h.name
@@ -370,6 +377,8 @@ class Node(ExecuteProcess):
         self.__final_node_name = ''
         if self.__expanded_node_namespace not in ['', '/']:
             self.__final_node_name += self.__expanded_node_namespace
+        elif self.__expanded_node_namespace == '':
+            self.__final_node_name += self.UNSPECIFIED_NODE_NAMESPACE
         self.__final_node_name += '/' + self.__expanded_node_name
         # expand parameters too
         if self.__parameters is not None:
@@ -423,7 +432,7 @@ class Node(ExecuteProcess):
         context.extend_locals({'ros_specific_arguments': ros_specific_arguments})
         ret = super().execute(context)
 
-        if '<node_name_unspecified>' not in self.node_name:
+        if self.is_node_name_fully_specified():
             add_node_name(context, self.node_name)
             node_name_count = get_node_name_count(context, self.node_name)
             if node_name_count > 1:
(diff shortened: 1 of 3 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -370,6 +377,8 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
         self.__final_node_name = ''
         if self.__expanded_node_namespace not in ['', '/']:
             self.__final_node_name += self.__expanded_node_namespace
+        elif self.__expanded_node_namespace == '':
+            self.__final_node_name += self.UNSPECIFIED_NODE_NAMESPACE
```
hidmic, line 381: @ivanpauno hmm, an empty string for a namespace is a valid argument for node creation (see [`rcl_node_init()`](https://github.com/ros2/rcl/blob/ab20449f19ea90670dc4419e9806e0c1b7b07d2d/rcl/include/rcl/node.h#L78-L79) documentation). 

I think that assuming an unspecified namespace is the root namespace would be the consistent thing to do.

hidmic: I didn't noticed it in #154, sorry about that.

ivanpauno: > @ivanpauno hmm, an empty string for a namespace is a valid argument for node creation (see rcl_node_init() documentation).

Sounds good.
I think that defaulting the namespace to `''` instead of `None` was a bad decision taken when `PushRosNamespace` action was added.
https://github.com/ros2/launch_ros/pull/153/commits/3933e356f12790e155968f0691bae43e066497b0 should fix the problem.

launch_ros/launch_ros/actions/node.py
```diff
@@ -340,19 +347,24 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
-            if self.__expanded_node_namespace != '':
+            expanded_node_namespace = None
+            if self.__node_namespace is not None:
+                expanded_node_namespace = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(self.__node_namespace))
+            base_ns = context.launch_configurations.get('ros_namespace', None)
+            if any(x is not None for x in (base_ns, expanded_node_namespace)):
+                expanded_node_namespace = (
+                    '' if expanded_node_namespace is None else expanded_node_namespace
+                )
```
hidmic, line 358: @ivanpauno nit^2: consider 
```python
if expanded_node_namespace is None:
    expanded_node_namespace = ''
```
instead.

launch_ros/launch_ros/actions/node.py
```diff
@@ -340,19 +347,24 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
+                base_ns = '' if base_ns is None else base_ns
```
hidmic, line 359: @ivanpauno nit^2: same as above.

launch_ros/launch_ros/actions/node.py
```diff
@@ -340,19 +347,24 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
-                    self.__expanded_node_namespace.startswith('/')
-                ):
-                    self.__expanded_node_namespace = '/' + self.__expanded_node_namespace
-            if self.__expanded_node_namespace != '':
+            expanded_node_namespace = None
+            if self.__node_namespace is not None:
+                expanded_node_namespace = perform_substitutions(
+                    context, normalize_to_list_of_substitutions(self.__node_namespace))
+            base_ns = context.launch_configurations.get('ros_namespace', None)
+            if any(x is not None for x in (base_ns, expanded_node_namespace)):
```
hidmic, line 355: @ivanpauno out of curiosity, why not simply:
```suggestion
            if base_ns is not None or expanded_node_namespace is not None:
```
?

launch_ros/launch_ros/actions/node.py
hidmic: Based on above's code (enforcing `self.__expanded_node_namespace` is a valid namespace or the `unspecified` token), wouldn't this code reduce to simply checking that `self.__expanded_node_namespace` isn't the root namespace before prepending?

ivanpauno: yeap, I've to add more tests to cover that case.

launch_ros/launch_ros/actions/node.py
ivanpauno: Done https://github.com/ros2/launch_ros/pull/153/commits/6538d10350599d8aa288e820d88f053bb53b091b

launch_ros/launch_ros/actions/node.py
ivanpauno: See https://github.com/ros2/launch_ros/pull/153/commits/6538d10350599d8aa288e820d88f053bb53b091b

launch_ros/launch_ros/actions/node.py
ivanpauno: IDK, I have a tendency of using `any` and `all` unnecessarily :joy:

launch_ros/launch_ros/actions/node.py
ivanpauno: Added more test cases in https://github.com/ros2/launch_ros/pull/153/commits/2e1f1ef3489547792941715983b9146ff24637dd

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11238/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6504/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9200/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11152/)

ivanpauno:
I'm not sure of how this can be backported.
It should be somehow combined with https://github.com/ros2/launch_ros/pull/157.

I will try to figure out if there's a no API breaking way of backporting the fix.

jacobperron:
@ivanpauno Friendly ping. Let me know if you still plan to look at backporting this change and #157. I think backporting those changes will make other backports easier to do (currently I get conflicts since they touch similar lines of code).

PR #154, opened 2020-06-19
Avoid passing parameters using a wildcard if fully qualified node name is known at launch time

ivanpauno:
This can be considered a fix of the problem described [here](https://github.com/ros2/rclcpp/issues/953#issuecomment-642784377).

This only solves the problem if both `name` and `namespace` were explicitly specified at launch.
If we finally decide to also change `rcl` behavior, this change doesn't do any harm.

Depends on https://github.com/ros2/launch_ros/pull/153.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -331,8 +331,10 @@ class Node(ExecuteProcess):
     def _create_params_file_from_dict(self, params):
         with NamedTemporaryFile(mode='w', prefix='launch_params_', delete=False) as h:
             param_file_path = h.name
-            # TODO(dhood): clean up generated parameter files.
-            param_dict = {'/**': {'ros__parameters': params}}
+            param_dict = {
+                self.node_name if self.is_node_name_fully_specified() else '/**':
+                {'ros__parameters': params}
+            }
             yaml.dump(param_dict, h, default_flow_style=False)
             return param_file_path
 
--- a/test_launch_ros/test/test_launch_ros/actions/test_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_node.py
@@ -165,7 +165,7 @@ class TestNode(unittest.TestCase):
         with open(expanded_parameter_files[0], 'r') as h:
             expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
             assert expanded_parameters_dict == {
-                '/**': {
+                '/my_ns/my_node': {
                     'ros__parameters': {
                         'param1': 'param1_value',
                         'param2': 'param2_value',


ivanpauno:
I will have to rebase on master after #153 gets merged.

ivanpauno:
@clalancette @jacobperron @hidmic do you consider this a fix to https://github.com/ros2/rclcpp/issues/953? should we change behavior in `rcl` too?

I can also add a warning if somebody passes both parameter files and a dictionary, telling the user that the passed parameters can't override values in the parameter file if they don't specify the fully qualified node name at launch time.

launch_ros/launch_ros/actions/node.py
```diff
@@ -321,11 +324,17 @@ def node_name(self):
             raise RuntimeError("cannot access 'node_name' before executing action")
         return self.__final_node_name
 
+    def is_node_name_fully_specified(self):
+        keywords = (self.UNSPECIFIED_NODE_NAME, self.UNSPECIFIED_NODE_NAMESPACE)
+        return all(x not in self.node_name for x in keywords)
```
hidmic, line 329: @ivanpauno meta, feel free to ignore: if we made this a free function (with all its implications), code would read better e.g.:

```python
if is_node_name_fully_specified(self.node_name):
    # ...
```

hidmic:
> do you consider this a fix to `ros2/rclcpp#953`? should we change behavior in `rcl` too?

I think they are orthogonal.

launch_ros/launch_ros/actions/node.py
ivanpauno: I don't think it's more readable. I also don't think that the free function will have uses not related with the Node class.

---

This comment applies to https://github.com/ros2/launch_ros/pull/153, in which this PR depends on. Please continue the review there.

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -164,8 +164,9 @@ def test_launch_node_with_parameter_dict(self):
         assert len(expanded_parameter_files) == 1
         with open(expanded_parameter_files[0], 'r') as h:
             expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
+            print(node_action.node_name)
```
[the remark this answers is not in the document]
ivanpauno: Done https://github.com/ros2/launch_ros/pull/154/commits/5ba49844de08bd2632b531e688b12a128b30cdc4.

launch_ros/launch_ros/actions/node.py
```diff
@@ -423,7 +434,7 @@ def execute(self, context: LaunchContext) -> Optional[List[Action]]:
         context.extend_locals({'ros_specific_arguments': ros_specific_arguments})
         ret = super().execute(context)
 
-        if '<node_name_unspecified>' not in self.node_name:
+        if self.is_node_name_fully_specified():
```
jacobperron, line 437: This if-condition doesn't look equivalent to the condition before this change. Should this be fixed so that it's okay if no namespace was specified?

I think this would change back the original behavior of `test_launch_node_with_name_without_namespace` below.

jacobperron: Nevermind. I see what's happening in https://github.com/ros2/launch_ros/pull/153

jacobperron:
> do you consider this a fix to ros2/rclcpp#953? should we change behavior in rcl too?

I agree that this change is separate from the rclcpp issue. I'm still not sure what the correct behavior should be at the `rcl` level.

ivanpauno:
Rebased after https://github.com/ros2/launch_ros/pull/153 was merged

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11284/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6533/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9234/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11189/)

PR #157, opened 2020-06-23
Make namespace parameter mandatory in LifecycleNode constructor

ivanpauno:
https://github.com/ros2/launch_ros/pull/153 introduced some regressions in nightlies:

https://ci.ros2.org/view/nightly/job/nightly_linux_release/1582/testReport/junit/(root)/projectroot/test_test_lifecycle_py/
https://ci.ros2.org/view/nightly/job/nightly_linux_release/1582/testReport/junit/lifecycle/TestLifecyclePubSub/test_talker_lifecycle/
https://ci.ros2.org/view/nightly/job/nightly_linux_release/1582/testReport/junit/lifecycle/TestLifecyclePubSubAfterShutdown/test_talker_graceful_shutdown/

The problem isn't #153 IMO.
The problem is that we were assuming that the node fqn name is known, even when we were not passing a `__ns` remapping rule (if the executable used a namespace different to `/` and a namespace wasn't being passed explicitly, LifecycleNode action was already broken).

There are two options:
- Make namespace argument required.
- Default to `/` (overriding whatever was specified in the executable).

I prefer the first option.
This is implementing the second option, as the first requires multiple PRs in other repos.

### DIFF
--- a/launch_ros/examples/lifecycle_pub_sub_launch.py
+++ b/launch_ros/examples/lifecycle_pub_sub_launch.py
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    name='listener', namespace=''
+                    name='listener', namespace='',
                     package='lifecycle', executable='lifecycle_listener', output='screen'),
             ],
         )
--- a/launch_ros/examples/lifecycle_pub_sub_launch.py
+++ b/launch_ros/examples/lifecycle_pub_sub_launch.py
@@ -36,7 +36,7 @@ def main(argv=sys.argv[1:]):
 
     # Prepare the talker node.
     talker_node = launch_ros.actions.LifecycleNode(
-        node_name='talker',
+        name='talker', namespace='',
         package='lifecycle', executable='lifecycle_talker', output='screen')
 
     # When the talker reaches the 'inactive' state, make it take the 'activate' transition.
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    node_name='listener',
+                    name='listener', namespace=''
                     package='lifecycle', executable='lifecycle_listener', output='screen'),
             ],
         )
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -23,6 +23,7 @@ from typing import Text
 import warnings
 
 import launch
+from launch import SomeSubstitutionsType
 from launch.action import Action
 import launch.logging
 
@@ -42,8 +43,9 @@ class LifecycleNode(Node):
     def __init__(
         self,
         *,
-        name: Optional[Text] = None,
-        node_name: Optional[Text] = None,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: SomeSubstitutionsType = '',
+        node_name: Optional[SomeSubstitutionsType] = None,
         **kwargs
     ) -> None:
         """
@@ -87,7 +89,7 @@ class LifecycleNode(Node):
         # TODO(jacobperron): Remove default value and this check when deprecated API is removed
         if name is None:
             raise RuntimeError("'name' must not be None.'")
-        super().__init__(name=name, **kwargs)
+        super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
         self.__current_state = \
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -43,7 +43,7 @@ class LifecycleNode(Node):
         self,
         *,
         name: Optional[SomeSubstitutionsType] = None,
-        namespace: SomeSubstitutionsType = '',
+        namespace: SomeSubstitutionsType,
         node_name: Optional[SomeSubstitutionsType] = None,
         **kwargs
     ) -> None:
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -19,7 +19,6 @@ import threading
 from typing import cast
 from typing import List
 from typing import Optional
-from typing import Text
 import warnings
 
 import launch
--- a/test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
@@ -21,14 +21,14 @@ import pytest
 
 
 def test_lifecycle_node_constructor():
-    # Construction without name
+    # Construction without namespace
     with pytest.raises(TypeError):
         LifecycleNode(
             package='asd',
             executable='bsd',
             name='my_node',
         )
-    # Construction without namespace
+    # Construction without name
     # TODO(ivanpauno): This should raise TypeError.
     with pytest.raises(RuntimeError):
         LifecycleNode(
--- a/test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
@@ -1,0 +1,57 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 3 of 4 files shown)


launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -87,7 +89,7 @@ def __init__(
         # TODO(jacobperron): Remove default value and this check when deprecated API is removed
         if name is None:
             raise RuntimeError("'name' must not be None.'")
-        super().__init__(name=name, **kwargs)
+        super().__init__(name=name, namespace=namespace, **kwargs)
```
sloretz, line 92: If `namespace` is already passed in via `**kwargs`, then it seems like this changes the default from `None` to `''`, which makes the default value used by `LifecycleNode` different from the one used by `Node`. I think they should have the same default value if possible to avoid behavior differences between the two actions.

https://github.com/ros2/launch_ros/blob/e8894ce6155b516b6c5f4dda1c5b4cba76c3e725/launch_ros/launch_ros/actions/node.py#L73

ivanpauno: > If namespace is already passed in via **kwargs, then it seems like this changes the default from None to '', which makes the default value used by LifecycleNode different from the one used by Node

Yes, this is changing only the default.

`LifecycleNode` action must know the fully qualified node name at launch time, if not it can't interact with the lifecycle services.

The other option is:

> Make namespace argument required.

As I commented in the post description, that's my preferred solution.
The only disadvantage is that I will have to create multiple PRs.

hidmic: I agree with you @ivanpauno, but I don't quite grok why you need multiple PRs to enforce a  namespace is given. Why isn't updating [line 161](https://github.com/ros2/launch_ros/blob/e8894ce6155b516b6c5f4dda1c5b4cba76c3e725/launch_ros/launch_ros/actions/lifecycle_node.py#L161) enough?

ivanpauno: Every place where the namespace wasn't specified in a `LifecycleNode` action have to change:
https://github.com/ros2/demos/blob/bda392ed2e4ce4c9ad428935382e97502ba7c676/lifecycle/launch/lifecycle_demo.launch.py#L22-L23

PS: I think that's the only place luckily

hidmic: I see. If that's the sole location wouldn't it be better to just fix it?

sloretz: > LifecycleNode action must know the fully qualified node name at launch time, if not it can't interact with the lifecycle services.

Ah I see, `LifecycleNode` needs to know the namespace, so it makes sense for it to have different behavior from Node. +1 to `LifecycleNode` requiring the `namespace` parameter instead of defaulting to `/`

> I see. If that's the sole location wouldn't it be better to just fix it?

+1

sloretz: If I understand correctly, this alone will fix the test https://github.com/ros2/demos/pull/455

I still agree making LifecycleNode require namespace makes sense, but the test fix can go in first while that's being discussed.

ivanpauno: See https://github.com/ros2/demos/pull/456.

> I see. If that's the sole location wouldn't it be better to just fix it?

I wasn't sure what was the intended behavior the original author thought, and thus I took the easiest approach in the first iteration.
Maybe the intention was "use the root namespace implicitly if no namespace is specified".

If we want to backport some of the fixes in https://github.com/ros2/launch_ros/pull/153 and here, we will have to do something like that, as changing an optional argument to required is not acceptable in a backport.

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -42,8 +42,9 @@ class LifecycleNode(Node):
     def __init__(
         self,
         *,
-        name: Optional[Text] = None,
-        node_name: Optional[Text] = None,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: SomeSubstitutionsType,
+        node_name: Optional[SomeSubstitutionsType] = None,
```
sloretz, line 47: Would expect SyntaxError here

```python
>>> def foo(a=2, b, c=3):
...    print(f"{a} {b} {c}")
... 
  File "<stdin>", line 1
SyntaxError: non-default argument follows default argument
```

launch_ros/examples/lifecycle_pub_sub_launch.py
```diff
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    node_name='listener',
+                    name='listener', namespace=''
```
sloretz, line 65: Same comment about `''` vs `'/'` https://github.com/ros2/demos/pull/456#discussion_r444403220

ivanpauno: See https://github.com/ros2/demos/pull/456#discussion_r444414183

launch_ros/launch_ros/actions/lifecycle_node.py
ivanpauno: AFAIK it doesn't apply to only keyword arguments, only to positional arguments

sloretz: Ah, sure enough

```python3
>>> def foo(*, a=2, b, c=3):
...    print(f"{a} {b} {c}")
... 
>>> foo(a=1, b=2, c=3)
1 2 3
```

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -300,3 +299,33 @@ def test_launch_node_with_invalid_parameter_dicts(self):
...
+
+    node_with_fqn = launch_ros.actions.Node(
+        package='asd',
+        executable='bsd',
+        name='my_node',
+        namespace='my_ns',
+    )
+    lc = LaunchContext()
+    node_with_fqn._perform_substitutions(lc)
+    assert node_with_fqn.is_node_name_fully_specified()
```
sloretz, line 331: I think it would make sense to add a test with this last check using `LifecycleNode` since that's the class changed by this PR.

The rest of the test make sense to me, though they don't seem to be related to this PR. I would approve regardless as long as a `LifecycleNode` test using `namespace` is added.

ivanpauno: Yeah, those test are a bit unrelated.
I should have added them in #153.

After https://github.com/ros2/launch_ros/pull/157/commits/a9e4c1045d3cd3c35b84021035a47598d69f5c14, the added tests are covering LifecycleNode too.

ivanpauno: It seems that I forgot to run tests locally before pushing :open_mouth:.
https://github.com/ros2/launch_ros/pull/157/commits/4253a34821937ca68f1d3d3d7a129bd60daa3060 should fix the problems.

ivanpauno:
@hidmic @sloretz @wjwwood friendly ping

launch_ros/launch_ros/actions/lifecycle_node.py
hidmic, line 47: @ivanpauno how's that `name` is optional but `namespace` isn't?

test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
```diff
@@ -0,0 +1,57 @@
...
+
+def test_lifecycle_node_constructor():
+    # Construction without name
+    with pytest.raises(TypeError):
+        LifecycleNode(
+            package='asd',
+            executable='bsd',
+            name='my_node',
+        )
+    # Construction without namespace
```
hidmic, line 31: @ivanpauno nit: comments do not match assertions

launch_ros/launch_ros/actions/lifecycle_node.py
ivanpauno: https://github.com/ros2/launch_ros/blob/e8894ce6155b516b6c5f4dda1c5b4cba76c3e725/launch_ros/launch_ros/actions/lifecycle_node.py#L80-L89

test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
ivanpauno: Ups, fixed in https://github.com/ros2/launch_ros/pull/157/commits/620801e6efd80713b5a19eaa98e17ae630bfbb1c.

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11283/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6532/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9233/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11187/)

launch_ros/launch_ros/actions/lifecycle_node.py
hidmic: Hmm, right, the deprecated argument. It was [deprecated on Foxy](https://github.com/ros2/launch_ros/blob/3775aa3255ddb8c7271d47c71f08dd0f7cf14d2c/launch_ros/launch_ros/actions/node.py#L129-L131) though, so I think we can drop it (on a follow-up a PR).

ivanpauno:
@sloretz @wjwwood let me know if there are any further comments about this change, and I will address them.

wjwwood:
I looked it over just now. I think it was fine as-is.

Issue #159, opened 2020-06-25
PushRosNamespace action should affect ComposableNode descriptions


ivanpauno:
## Bug report

The following launch description:

```python3
def generate_launch_description():
    """Generate launch description with multiple components."""
    push_ns = PushRosNamespace('my_ns')
    container = ComposableNodeContainer(
            name='my_container',
            namespace='',
            package='rclcpp_components',
            executable='component_container',
            composable_node_descriptions=[
                ComposableNode(
                    package='composition',
                    plugin='composition::Talker',
                    name='talker'),
                ComposableNode(
                    package='composition',
                    plugin='composition::Listener',
                    name='listener')
            ],
            output='screen',
    )

    return launch.LaunchDescription([push_ns, container])
```

Should result in nodes named:
* /my_ns/talker
* /my_ns/listener

and they are resulting in:

* /talker
* /listener

Disclaimer: I have to double check this. I will do it after merging https://github.com/ros2/launch_ros/pull/158, which may have conflicting changes.

dirk-thomas:
@ivanpauno Will you be able to look into this? If yes, please assign the ticket to yourself.

PR #162, opened 2020-07-14
Fix ComposableNode ignoring PushRosNamespace actions

ivanpauno:
Fixes https://github.com/ros2/launch_ros/issues/159.

I refactored common namespace handling logic in a utilities module.


launch_ros/launch_ros/utilities/namespace_utils.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def prefix_namespace(
+    base_ns: Optional[Text],
+    ns: Optional[Text],
+    *,
+    return_absolute_ns: bool = True,
+) -> Optional[Text]:
+    """
+    Return `ns` prefixed with `base_ns` if `ns` is relative, returns `ns` if not.
```
hidmic, line 39: @ivanpauno nit: 
```suggestion
    Return `ns` prefixed with `base_ns` if `ns` is relative, return `ns` if not.
```

launch_ros/launch_ros/utilities/namespace_utils.py
```diff
@@ -0,0 +1,96 @@
...
+        `ns` if `base_ns` is `None`, or
+        `ns` if `ns` is absolute, or
+        `ns` prefixed with `base_ns`.
+        In all cases, trailing `/` are stripped from the result.
+
+    ## Examples:
+
+    ```python3
+    combined_ns = prefix_namespace('my_ns', 'original_ns', return_absolute_ns=True)
+    assert comined_ns == '/my_ns/original_ns'
```
hidmic, line 55: @ivanpauno nit: 
```suggestion
    assert combined_ns == '/my_ns/original_ns'
```
same elsewhere.

launch_ros/launch_ros/utilities/namespace_utils.py
```diff
@@ -0,0 +1,96 @@
...
+    ns: Optional[Text],
+    *,
+    return_absolute_ns: bool = True,
+) -> Optional[Text]:
+    """
+    Return `ns` prefixed with `base_ns` if `ns` is relative, returns `ns` if not.
+
+    :param `base_ns`: Prefix to be added to `ns`.
+    :param `ns`: Namespace to be prefixed.
+    :param `return_absolute_ns`: if `True`, the result will be made absolute.
```
hidmic, line 43: @ivanpauno meta: instead of having this argument, consider using `make_absolute_namespace` at `prefix_namespace` call site when necessary e.g.:

```python
make_absolute_namespace(prefix_namespace(base_ns, ns))
```

ivanpauno: I thought about that, the problem is that I actually need:
```python3
combined_ns = prefix_namespace(base_ns, ns)
if combined_ns is not None:
  combined_ns = make_absolute_namespace(combined_ns)
```

and I was too lazy too write all that :joy:.

I can also modify `make_absolute_namespace` to handle `None` inputs, but I preferred not accepting that.
WDYT?

hidmic: Hmm, since `prefix_namespace` does handle `None`, changing `make_absolute_name` to do so as well doesn't sound bad. But I don't mind to leave as-is.

launch_ros/launch_ros/utilities/namespace_utils.py
hidmic: Check the other examples below, they have the same defect.

launch_ros/launch_ros/utilities/namespace_utils.py
hidmic: Maybe remove all `return_absolute_ns` references from the documentation?

ivanpauno: thanks for catching that

ivanpauno: Fixed in https://github.com/ros2/launch_ros/pull/162/commits/5c22c0f8262b8b970c5417063124d2ddccac8d91.

ivanpauno:
Rpr checker is passing, merging!

PR #158, opened 2020-06-25
Add a SetParameter action that sets a parameter to all nodes in the same scope

ivanpauno:
Similarly to how `PushRosNamespace` action works, this allows a way to set a parameter affecting all `Node` actions and `ComposableNode` descriptions in the same scope.

This partially fixes https://github.com/ros2/launch_ros/issues/45, the other half is doing the same for remapping rules.

This introduces some conflicts with https://github.com/ros2/launch_ros/pull/137, but either way I already have to rebase that PR.

PR #163, opened 2020-07-23
Add a way to set remapping rules for all nodes in the same scope

### LINKED ISSUE #45: Add a way of grouping remapping rules and parameters
#### Feature description

We don't have a `SetRemapping` action.
It could be useful, is the same remapping has to be applied in many nodes.

We could store them in the `LaunchContext`. The group action will do its magic for limiting their scope.

Previous discussion: https://github.com/ros2/ros2_documentation/pull/302#discussion_r306931142

----

We could do something similar for parameters.
But it won't set a global parameter, as in ROS 1, but it'll set the same parameter in each node in the scope.

Previous discussion:
https://github.com/ros2/ros2_documentation/pull/302#discussion_r306927844

ivanpauno:
Fixes https://github.com/ros2/launch_ros/issues/45.
The other half of https://github.com/ros2/launch_ros/pull/158.

I'm unsure of how precedence should work.
When importing a launch file, it might become handy to override the remapping rule that was specified in the node.
Should that be the default?

For parameters, I took the opposite approach ... maybe we should revisit that after this PR.
Related discussion: https://github.com/ros2/design/pull/289.

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -200,12 +200,17 @@ def get_composable_node_load_request(
     if combined_ns is not None:
         request.node_namespace = combined_ns
     # request.log_level = perform_substitutions(context, node_description.log_level)
-    if composable_node_description.remappings is not None:
-        for from_, to in composable_node_description.remappings:
-            request.remap_rules.append('{}:={}'.format(
-                perform_substitutions(context, list(from_)),
-                perform_substitutions(context, list(to)),
-            ))
+    remappings = []
+    global_remaps = context.launch_configurations.get('ros_remaps', None)
+    if global_remaps:
+        remappings.extend([f'{src}:={dst}' for src, dst in global_remaps])
+    if composable_node_description.remappings:
+        remappings.extend([
+            f'{perform_substitutions(context, src)}:={perform_substitutions(context, dst)}'
+            for src, dst in composable_node_description.remappings
+        ])
+    if remappings:
+        request.remap_rules = remappings
     global_params = context.launch_configurations.get('ros_params', None)
     parameters = []
     if global_params is not None:
--- a/launch_ros/launch_ros/actions/set_remap.py
+++ b/launch_ros/launch_ros/actions/set_remap.py
@@ -32,7 +32,8 @@ class SetRemap(Action):
     """
     Action that sets a remapping rule in the current context.
 
-    This remapping rule will be passed to all the nodes launched in the same scope.
+    This remapping rule will be passed to all the nodes launched in the same scope, overriding
+    the ones specified in the `Node` action constructor.
     e.g.:
     ```python3
         LaunchDescription([
@@ -40,7 +41,7 @@ class SetRemap(Action):
             GroupAction(
                 actions = [
                     ...,
-                    SetRemap(from='asd', to='bsd'),
+                    SetRemap(src='asd', dst='bsd'),
                     ...,
                     Node(...),  // the remap rule will be passed to this node
                     ...,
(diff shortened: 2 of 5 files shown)


wjwwood (approved):
lgtm

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11619/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6753/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9498/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11580/)

launch_ros/launch_ros/actions/set_remap.py
```diff
@@ -0,0 +1,90 @@
...
+
+    This remapping rule will be passed to all the nodes launched in the same scope.
+    e.g.:
+    ```python3
+        LaunchDescription([
+            ...,
+            GroupAction(
+                actions = [
+                    ...,
+                    SetRemap(from='asd', to='bsd'),
```
hidmic, line 43: @ivanpauno nit: I imagine you moved away from `from` to avoid colliding with the keyword. If so, this example needs an update.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -200,12 +200,17 @@ def get_composable_node_load_request(
...
         request.node_namespace = combined_ns
     # request.log_level = perform_substitutions(context, node_description.log_level)
-    if composable_node_description.remappings is not None:
-        for from_, to in composable_node_description.remappings:
-            request.remap_rules.append('{}:={}'.format(
-                perform_substitutions(context, list(from_)),
-                perform_substitutions(context, list(to)),
-            ))
+    remappings = []
+    global_remaps = context.launch_configurations.get('ros_remaps', None)
```
hidmic, line 204: @ivanpauno consider documenting that global remaps take precedence.

test_launch_ros/test/test_launch_ros/actions/test_set_remap.py
```diff
@@ -0,0 +1,106 @@
...
+def test_set_remap_with_node():
+    lc = MockContext()
+    node = Node(
+        package='asd',
+        executable='bsd',
+        name='my_node',
+        namespace='my_ns',
+        remappings=[('from2', 'to2')]
+    )
+    set_remap = SetRemap('from1', 'to1')
```
hidmic, line 84: @ivanpauno mind to add another test to verify that global remaps take precedence?

launch_ros/launch_ros/actions/set_remap.py
ivanpauno: > I imagine you moved away from from to avoid colliding with the keyword

Yeap, done

launch_ros/launch_ros/actions/load_composable_nodes.py
ivanpauno: done

test_launch_ros/test/test_launch_ros/actions/test_set_remap.py
ivanpauno: > @ivanpauno mind to add another test to verify that global remaps take precedence?

They take precedence by being prepended (because `rcl` precedence works like that for remapping rules).
This test case already tests correctly that the "global" remapping rule is prepended.

The test in L55 already verifies (for the second parameter), that the `SetRemap` action at the upper scope will take precedence.

hidmic: Yeah, misread it.

hidmic (approved):
LGTM !

ivanpauno:
Last commit only improved doc blocks, merging!

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/tf-tree-in-a-multi-robot-setup-in-ros2/41426/8

Issue #114, opened 2020-01-16
LaunchConfiguration is not read from ComposableNode when included inside a group action


jacobperron:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 18.04
- Installation type:
  - source
- Version or commit hash:
  - c5e2aa2bf0eb60d61c66dbe0a6da365230b3387d
- DDS implementation:
  - N/A
- Client library (if applicable):
  - N/A

#### Steps to reproduce issue

Create the following two launch files:

`foo.launch.xml`
```
<launch>
  <group>
    <include file='bar.launch.py'>
    </include>
  </group>
</launch>
```

`bar.launch.py`
```
import launch
import launch_ros


def generate_launch_description():
    return launch.LaunchDescription([
        launch.actions.DeclareLaunchArgument(name='bar_arg', default_value='True'),
        launch_ros.actions.ComposableNodeContainer(
            package='rclcpp_components', node_executable='component_container',
            node_name='my_container', node_namespace='',
            composable_node_descriptions=[
                launch_ros.descriptions.ComposableNode(
                    package='demo_nodes_cpp',
                    node_plugin='demo_nodes_cpp::Talker',
                    parameters=[{'use_sim_time': launch.substitutions.LaunchConfiguration('bar_arg')}]
                ),
            ]
        ),
        # launch_ros.actions.Node(
        #     name='talker', package='demo_nodes_cpp', node_executable='talker', output='screen',
        #     parameters=[{'use_sim_time': launch.substitutions.LaunchConfiguration('bar_arg')}]),
    ])
```

Launch `foo.launch.xml`:

    ros2 launch foo.launch.xml

#### Expected behavior

No launch errors.

#### Actual behavior

Launch error:

```
[INFO] [launch]: All log files can be found below /home/jacob/.ros/log/2020-01-16-13-53-42-380127-warner-9368
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [component_container-1]: process started with pid [9379]
[ERROR] [launch]: Caught exception in launch (see debug for traceback): launch configuration 'bar_arg' does not exist
[INFO] [component_container-1]: sending signal 'SIGINT' to process[component_container-1]
[INFO] [component_container-1]: process has finished cleanly [pid 9379]
```

<details>
<summary>Debug output</summary>
<pre>
[DEBUG] [launch.launch_context]: emitting event synchronously: 'launch.events.IncludeLaunchDescription'
[...]

PR #166, opened 2020-08-04
Fix race with launch context changes when loading composable nodes

jacobperron:
This bug was discovered when trying load composable nodes from a GroupAction.
The ROS namespace (and presumably other remaps)  pushed onto the context stack was popped after the LoadComposableNodes execute() function finished.
But because the loading happens asynchronously, we need to make sure we get the necessary information from the context before execute() finishes.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
@@ -14,7 +14,6 @@
 
 """Tests for the LoadComposableNodes Action."""
 
-import asyncio
 import threading
 
 from composition_interfaces.srv import LoadNode
@@ -27,8 +26,6 @@ from launch_ros.actions import PushRosNamespace
 from launch_ros.actions import SetRemap
 from launch_ros.descriptions import ComposableNode
 from launch_ros.utilities import get_node_name_count
-
-import osrf_pycommon.process_utils
 
 import pytest
 
@@ -82,18 +79,11 @@ class MockComponentContainer(rclpy.node.Node):
         self._thread.join()
 
 
-def _assert_launch_no_errors(actions, *, timeout_sec=1):
+def _assert_launch_no_errors(actions):
     ld = LaunchDescription(actions)
     ls = LaunchService(debug=True)
     ls.include_launch_description(ld)
-
-    loop = osrf_pycommon.process_utils.get_loop()
-    launch_task = loop.create_task(ls.run_async())
-    loop.run_until_complete(asyncio.sleep(timeout_sec))
-    if not launch_task.done():
-        loop.create_task(ls.shutdown())
-        loop.run_until_complete(launch_task)
-    assert 0 == launch_task.result()
+    assert 0 == ls.run()
     return ls.context
 
 
--- a/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
@@ -59,9 +59,14 @@ def _create_node_container(*, parameters=None, remappings=None, namespace=''):
     )
 
 
-def _load_composable_node(*, parameters=None, remappings=None):
+def _load_composable_node(
+    *,
+    parameters=None,
+    remappings=None,
+    target_container=TEST_CONTAINER_NAME
+):
     return LoadComposableNodes(
-        target_container=TEST_CONTAINER_NAME,
+        target_container=target_container,
         composable_node_descriptions=[
             ComposableNode(
                 package='composition',
--- a/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
@@ -1,0 +1,148 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Tests for the LoadComposableNodes Action."""
+
+import asyncio
+
+from launch import LaunchDescription
+from launch import LaunchService
+from launch.actions import GroupAction
+from launch_ros.actions import ComposableNodeContainer
+from launch_ros.actions import LoadComposableNodes
+from launch_ros.actions import PushRosNamespace
+from launch_ros.actions import SetRemap
+from launch_ros.descriptions import ComposableNode
+from launch_ros.utilities import get_node_name_count
+
+import osrf_pycommon.process_utils
+
+TEST_CONTAINER_NAME = 'test_load_composable_nodes_container'
+TEST_NODE_NAME = 'test_load_composable_nodes_node'
+
+
+def _assert_launch_no_errors(actions, *, timeout_sec=1):
+    ld = LaunchDescription(actions)
+    ls = LaunchService(debug=True)
+    ls.include_launch_description(ld)
+
+    loop = osrf_pycommon.process_utils.get_loop()
+    launch_task = loop.create_task(ls.run_async())
+    loop.run_until_complete(asyncio.sleep(timeout_sec))
+    if not launch_task.done():
+        loop.create_task(ls.shutdown())
+        loop.run_until_complete(launch_task)
+    assert 0 == launch_task.result()
+    return ls.context
+
+
+def _create_node_container(*, parameters=None, remappings=None, namespace=''):
+    return ComposableNodeContainer(
+        package='rclcpp_components',
+        executable='component_container',
+        name=TEST_CONTAINER_NAME,
+        namespace=namespace,
+        output='screen',
+        parameters=parameters,
+        remappings=remappings,
+    )
+
+
+def _load_composable_node(*, parameters=None, remappings=None):
+    return LoadComposableNodes(
+        target_container=TEST_CONTAINER_NAME,
+        composable_node_descriptions=[
+            ComposableNode(
+                package='composition',
+                plugin='composition::Listener',
+                name=TEST_NODE_NAME,
+                parameters=parameters,
+                remappings=remappings,
+            )
+        ])
+
+
+def test_load_invalid_node():
+    """Test loading an invalid node."""
+    load_composable_nodes = LoadComposableNodes(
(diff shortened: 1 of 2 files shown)


ivanpauno:
@jacobperron Does this fix https://github.com/ros2/launch_ros/issues/114?

jacobperron:
> Can you add a regression test?

Will do.

> Does this fix #114?

I had completely forgotten about that issue. I'll check to see if it fixes it.

jacobperron:
So, it appears this PR fixes the example mentioned in https://github.com/ros2/launch_ros/issues/114#issuecomment-624904000, which involves setting the namespace, but not the original issue that involves setting parameters on the node. I'll look into if a similar fix to the one proposed in this PR can be used to fix #114.

ivanpauno:
> So, it appears this PR fixes the example mentioned in [#114 (comment)](https://github.com/ros2/launch_ros/issues/114#issuecomment-624904000), which involves setting the namespace, but not the original issue that involves setting parameters on the node. I'll look into if a similar fix to the one proposed in this PR can be used to fix #114.

That's weird, because `get_composable_node_load_request` also resolves the parameters to be loaded.

jacobperron:
> That's weird, because get_composable_node_load_request also resolves the parameters to be loaded.

Is there a difference between "global" params (aka launch config `ros_params`) and parameters passed directly to a node?

jacobperron:
hmm, but it's not the parameters that are being dropped, but specifically a user-defined launch configuration "bar_arg". I can do some more debugging.

jacobperron:
I've added some regression tests in 719ced8.

There's still a bug related to parameters (and some other corner-cases); I'll open up separate PRs for those.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -0,0 +1,148 @@
...
+        namespace=namespace,
+        output='screen',
+        parameters=parameters,
+        remappings=remappings,
+    )
+
+
+def _load_composable_node(*, parameters=None, remappings=None):
+    return LoadComposableNodes(
+        target_container=TEST_CONTAINER_NAME,
```
ivanpauno, line 64: what about the container namespace?
`_create_node_container` is optionally using one.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -0,0 +1,148 @@
...
+        composable_node_descriptions=[
+            ComposableNode(
+                package='nonexistent_package',
+                plugin='this_plugin_should_not_exist',
+                name=TEST_NODE_NAME,
+            )
+        ])
+    context = _assert_launch_no_errors([_create_node_container(), load_composable_nodes])
+
+    assert get_node_name_count(context, f'/{TEST_NODE_NAME}') == 0
```
ivanpauno, line 89: should this be tested here?
sounds unrelated to me.

Maybe, what you want is checking the value of the [response](https://github.com/ros2/launch_ros/blob/9250313932c2c7477a982ef0c0cfe5124d3cfddf/launch_ros/launch_ros/actions/load_composable_nodes.py#L116). That could be introspected if the action would store the responses.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -0,0 +1,148 @@
...
+            ],
+            scoped=True,
+        ),
+    ])
+
+    assert get_node_name_count(context, f'/{TEST_NODE_NAME}') == 1
+    assert get_node_name_count(context, f'/{TEST_CONTAINER_NAME}') == 1
+
+    # Check the remaps in load service request
+    assert len(load_composable_nodes_action._LoadComposableNodes__load_node_requests) == 1
```
ivanpauno, line 121: maybe the sent request could be a property, so we don't have to check a private member variable

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
ivanpauno: It could actually be checked in the request instead of the response

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
jacobperron: I followed a similar pattern that is used in [test_node.py](https://github.com/ros2/launch_ros/blob/9250313932c2c7477a982ef0c0cfe5124d3cfddf/test_launch_ros/test/test_launch_ros/actions/test_node.py#L91).

I think a user should have access to that information; it seems like an implementation detail subject to change.

ivanpauno: > I followed a similar pattern that is used in test_node.py.

Yeah, it's not a great pattern, but acceptable I guess.

> I think a user should have access to that information; it seems like an implementation detail subject to change.

That sounds contradictory, I guess you meant "should not".

---

I prefer not testing implementation details, but I don't feel strongly in this case.

jacobperron: Yeah, sorry, I meant "should not" (updated my comment).

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
jacobperron: I wanted to check that the node is not launched. I think this is a proxy for that; maybe there's a better way. The request  contains the same information we passed in the composable node description, regardless if the loading was successful or not.

We *could* check the response too, but I was avoiding the storing more information than was necessary. I can add it if you think it would be better.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
jacobperron: Good point, I've made the target container a parameter in a12906b

We're not using the namespace in any tests, but I might use of it in follow-up patches.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
jacobperron: I guess the ideal way to write tests in this situation would be to write a mock component container and assert things about it (e.g. if the correct service request was received). It adds a bit more to the test complexity, but could be better long-term. What do you think?

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
ivanpauno: thanks

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
ivanpauno: Ok, I see what this is testing now.
The approach is a bit weird, but it seems acceptable.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
ivanpauno: > What do you think?

Sounds like a good proposal to me, though I don't mind approving this as-is.

jacobperron: I went ahead and refactored the tests to use a mock object. I think it's much nicer. PTAL 5d9dcf3

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -0,0 +1,289 @@
...
+    assert request.plugin_name == 'bar_plugin'
+    assert request.node_name == 'test_node_name'
+    assert request.node_namespace == '/test_node_namespace'
+    assert len(request.remap_rules) == 0
+    assert len(request.parameters) == 2
+    assert request.parameters[0].name == 'test_param1'
+    assert request.parameters[0].value.type == ParameterType.PARAMETER_STRING
+    assert request.parameters[0].value.string_value == 'test_value_param1'
+    assert request.parameters[1].name == 'test.param2'
+    # TODO(jacobperron): I would expect this to be a double value, but we can only pass strings
```
jacobperron, line 220: I'll take a look at resolving this in a follow-up.

jacobperron:
ugh, there's an issue with the tests where they hang sometimes... I can reproduce locally. I think there's an issue shutting down the launch context *before* LoadComposableNodes has had a chance to make a service request.

jacobperron:
I've discovered a race in launch (reported in https://github.com/ros2/launch_ros/issues/169), with a proposed fix in https://github.com/ros2/launch/pull/449. With the proposed fix, the tests in the PR pass, otherwise we can expect them to hang forever.

ivanpauno:
> I've discovered a race in launch (reported in #169), with a proposed fix in [ros2/launch#449](https://github.com/ros2/launch/pull/449). With the proposed fix, the tests in the PR pass, otherwise we can expect them to hang forever.

I understand the race, but what is causing it here?
The thread you're creating isn't emiting any event, so I would like to understand what exactly is causing the race here (the launch service is running in a single threaded event loop)

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11934/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6977/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9724/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11873/)

ivanpauno:
@jacobperron I've cancelled a couple of jobs that were hanging

jacobperron:
Sadly, it still seems like there's a race somewhere causing the tests to hang somteimes :disappointed:

jacobperron:
I wasn't properly shutting down the mock container and so the synchronous service call in `LoadComposableNodes` was blocking indefinitely. I've fixed the test in 9c50db2 and opened up a follow-up ticket to improve `LoadComposableNodes` such that we avoid this situation https://github.com/ros2/launch_ros/issues/171.

jacobperron:
Note, I don't expect the PR CI job to pass until ros2/launch#449 is released.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/11948/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/6991/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/9740/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/11891/)

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -144,6 +146,7 @@ def test_load_node(mock_component_container):
     assert len(request.extra_arguments) == 0
 
 
+
```
ivanpauno: nit: too much spacing(?)

jacobperron: Fixed in cf1437a

jacobperron:

PR #175, opened 2020-08-23
Address security bug in yaml loading

vmayoral:
Though testing code, still worth curating the bug. 

Similar to https://github.com/ros/actionlib/pull/171 and connected 
with https://github.com/aliasrobotics/RVD/issues/2401.

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_node.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_node.py
@@ -219,7 +219,7 @@ class TestNode(unittest.TestCase):
         file_path, is_file = expanded_parameter_arguments[0]
         assert is_file
         with open(file_path, 'r') as h:
-            expanded_parameters_dict = yaml.safe_load(h, Loader=yaml.FullLoader)
+            expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
             assert expanded_parameters_dict == {
                 '/my_ns/my_node': {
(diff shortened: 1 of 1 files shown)


test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -219,7 +219,7 @@ def test_launch_node_with_parameter_dict(self):
         file_path, is_file = expanded_parameter_arguments[0]
         assert is_file
         with open(file_path, 'r') as h:
-            expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
+            expanded_parameters_dict = yaml.safe_load(h, Loader=yaml.FullLoader)
```
dirk-thomas, line 222: This change currently fails CI:

> TypeError: safe_load() got an unexpected keyword argument 'Loader'

dirk-thomas: Since CI (https://ci.ros2.org/view/All/job/test_ci_linux/42/pytest-warnings-with-location/) only marks the first case as a warning I went ahead and reverted the second change in ef54e1c7b73e54acfcf3b2339337707d52bd3316.

dirk-thomas:
@vmayoral I will go ahead and merge the reduced patch for now since it resolves a warning shown in the CI output. If you still think the reverted case needs to be changed please create a follow up pull request. Thanks.

PR #178, opened 2020-09-17
Clean up temporary YAML parameter file

mabelzhang:
Addresses issue #5 

Opt-out option:
As described in the original issue, the test requires an opt-out option so that it can validate the outputted file.
Not sure if the current way (a `__param_file_cleanup` parameter) is the best way to implement that option. Suggestions of a more proper way?

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -338,7 +338,12 @@ class Node(ExecuteProcess):
                 {'ros__parameters': params}
             }
             yaml.dump(param_dict, h, default_flow_style=False)
-            return param_file_path
+
+            param_file_cleanup = True
+            if params['__param_file_cleanup'] == 0 or \
+                    params['__param_file_cleanup'].lower() == 'false':
+                param_file_cleanup = False
+            return param_file_path, param_file_cleanup
 
     def _get_parameter_rule(self, param: 'Parameter', context: LaunchContext):
         name, value = param.evaluate(context)
@@ -347,6 +352,8 @@ class Node(ExecuteProcess):
     def _perform_substitutions(self, context: LaunchContext) -> None:
         # Here to avoid cyclic import
         from ..descriptions import Parameter
+        param_file_path = None
+        param_file_cleanup = True
         try:
             if self.__substitutions_performed:
                 # This function may have already been called by a subclass' `execute`, for example.
@@ -388,7 +395,7 @@ class Node(ExecuteProcess):
         if global_params is not None or self.__parameters is not None:
             self.__expanded_parameter_arguments = []
         if global_params is not None:
-            param_file_path = self._create_params_file_from_dict(global_params)
+            param_file_path, param_file_cleanup = self._create_params_file_from_dict(global_params)
             self.__expanded_parameter_arguments.append((param_file_path, True))
             cmd_extension = ['--params-file', f'{param_file_path}']
             self.cmd.extend([normalize_to_list_of_substitutions(x) for x in cmd_extension])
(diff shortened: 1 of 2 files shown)


clalancette:
> Not sure if the current way (a `__param_file_cleanup` parameter) is the best way to implement that option. Suggestions of a more proper way?

Yeah, I don't like having a "magic" parameter that controls the behavior of an implementation detail.  Some thoughts:
1.  Do we really need to keep the file around in any cases?  I understand the current test needs it, but can we refactor that test to do something similar but not require the file lying around?
1.  If we really want to have the ability to control cleanup behavior (maybe for debugging?), I think an environment variable is one better way to do it.
1.  @hidmic or @ivanpauno may have another idea on how to pass a parameter like this to the launch_ros machinery.  I'm just not familiar enough.

ivanpauno:
> Do we really need to keep the file around in any cases? I understand the current test needs it, but can we refactor that test to do something similar but not require the file lying around?

We need to keep the file while the node executable is running, because we don't know when it will be parsed by `rcl`.

> If we really want to have the ability to control cleanup behavior (maybe for debugging?), I think an environment variable is one better way to do it.

I don't see why we need a way to opt-out this.

> @hidmic or @ivanpauno may have another idea on how to pass a parameter like this to the launch_ros machinery. I'm just not familiar enough.

We could stop generating a temporal file and start using command line arguments to set them, but I don't think we need to do that to fix the issue.

launch_ros/launch_ros/actions/node.py
```diff
@@ -461,6 +471,11 @@ def execute(self, context: LaunchContext) -> Optional[List[Action]]:
                     'launch context'.format(node_name_count, self.node_name)
                 )
 
+        # Delete temporary parameter yaml file
+        if param_file_cleanup and param_file_path is not None:
```
ivanpauno, line 475: The file cannot be removed here, as the executable may not have parsed it yet.
We could delete the files in an [OnExit](https://github.com/ros2/launch/blob/4370b2b829946cd8c5ec5a76cbb3bb6a70c95e97/launch/launch/actions/execute_process.py#L106) event (that is executed when the process is finished).

ivanpauno (changes requested):

launch_ros/launch_ros/actions/node.py
mabelzhang: a30d634

mabelzhang:
Should we just remove the opt-out option then? If we don't have the file around, what's a way to test that the node is getting the parameters as expected while it's running?

launch_ros/launch_ros/actions/node.py
ivanpauno: You need to register an event handler, e.g.: [here](https://github.com/ros2/launch/blob/4370b2b829946cd8c5ec5a76cbb3bb6a70c95e97/launch/launch/actions/execute_process.py#L831).

The launch event loop will trigger the registered `OnExit` event handler when the process finishes.

ivanpauno:
> Should we just remove the opt-out option then?

IMO, yes.
The file must be created before executing the process and kept alive until the process has finished.
I don't see the value of opting that out.

> If we don't have the file around, what's a way to test that the node is getting the parameters as expected while it's running?

mmm I'm not sure what you're asking for.
If you want to write a test case, it's possible to create another node and use the "GetParams" service to check everything went well.
But there shouldn't be logic like that in the `Node` action, we should simply keep the file alive while the process is running.

---

It would be good to also delete the files in the `__del__` method, just in case the `OnExit` event handler isn't run for some reason.

hidmic:
@mabelzhang are you still working on this?

mabelzhang:
I haven't had time to continue this lately. Not sure when I'll be able to pick it up again.

clalancette:
I'm going to close this out, since it is not clear that this is the right direction or that we'll have time to work on this.  Feel free to reopen if you disagree.

PR #179, opened 2020-09-18
[foxy backport] Fix no specified namespace (#153)

jacobperron:
Backporting #153 and #157 to Foxy.

In order to maintain backwards compatibility, I've kept the `namespace` argument to `LifecycleNode` as optional (f7c5990).

@ivanpauno PTAL, I *think* this is safe to backport, but you should have more context. I want to backport this, since it is a prerequisite for several other fixes I have made related to launching composable nodes.

### DIFF
--- a/launch_ros/examples/lifecycle_pub_sub_launch.py
+++ b/launch_ros/examples/lifecycle_pub_sub_launch.py
@@ -36,7 +36,7 @@ def main(argv=sys.argv[1:]):
 
     # Prepare the talker node.
     talker_node = launch_ros.actions.LifecycleNode(
-        node_name='talker',
+        name='talker', namespace='',
         package='lifecycle', executable='lifecycle_talker', output='screen')
 
     # When the talker reaches the 'inactive' state, make it take the 'activate' transition.
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    node_name='listener',
+                    name='listener', namespace='',
                     package='lifecycle', executable='lifecycle_listener', output='screen'),
             ],
         )
--- a/launch_ros/examples/lifecycle_pub_sub_launch.py
+++ b/launch_ros/examples/lifecycle_pub_sub_launch.py
@@ -36,7 +36,7 @@ def main(argv=sys.argv[1:]):
 
     # Prepare the talker node.
     talker_node = launch_ros.actions.LifecycleNode(
-        name='talker', namespace='',
+        name='talker',
         package='lifecycle', executable='lifecycle_talker', output='screen')
 
     # When the talker reaches the 'inactive' state, make it take the 'activate' transition.
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    name='listener', namespace='',
+                    name='listener',
                     package='lifecycle', executable='lifecycle_listener', output='screen'),
             ],
         )
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -19,10 +19,10 @@ import threading
 from typing import cast
 from typing import List
 from typing import Optional
-from typing import Text
 import warnings
 
 import launch
+from launch import SomeSubstitutionsType
 from launch.action import Action
 import launch.logging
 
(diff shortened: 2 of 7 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -61,15 +61,18 @@
...
         self, *,
         executable: Optional[SomeSubstitutionsType] = None,
         node_executable: Optional[SomeSubstitutionsType] = None,
         package: Optional[SomeSubstitutionsType] = None,
         name: Optional[SomeSubstitutionsType] = None,
-        namespace: Optional[SomeSubstitutionsType] = '',
+        namespace: Optional[SomeSubstitutionsType] = None,
         node_name: Optional[SomeSubstitutionsType] = None,
-        node_namespace: SomeSubstitutionsType = '',
+        node_namespace: SomeSubstitutionsType = None,
```
ivanpauno, line 75: is this change an API break?

ivanpauno: To me more clear, changing the default isn't the problem.
The problem is that `namespace=''` meant that the namespace in code was going to be used (no remapping rule passed), and now it means that the namespace is remapped to `/` (`__ns:=/` rule passed).

I think that is a (subtle) API break.
I think that the behavior now is the correct one, but I'm not sure if changing the behavior in a previous distro is ok.

jacobperron: Thanks for the review. I've amended the logic so that we don't change the behavior in Foxy (e7e08a7). Now `namespace=None` and `namespace=''` act the same (ie. no remapping rule is passed).

launch_ros/examples/lifecycle_pub_sub_launch.py
```diff
@@ -36,7 +36,7 @@ def main(argv=sys.argv[1:]):
 
     # Prepare the talker node.
     talker_node = launch_ros.actions.LifecycleNode(
-        node_name='talker',
+        name='talker', namespace='',
```
ivanpauno, line 39: The current change is ok, but this other version should still be working:
```suggestion
        name='talker'
```

launch_ros/examples/lifecycle_pub_sub_launch.py
```diff
@@ -62,7 +62,7 @@ def main(argv=sys.argv[1:]):
                 launch.actions.LogInfo(
                     msg="node 'talker' reached the 'active' state, launching 'listener'."),
                 launch_ros.actions.LifecycleNode(
-                    node_name='listener',
+                    name='listener', namespace='',
```
ivanpauno, line 65:
```suggestion
                    name='listener'
```

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -42,8 +42,9 @@ class LifecycleNode(Node):
     def __init__(
         self,
         *,
-        name: Optional[Text] = None,
-        node_name: Optional[Text] = None,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: SomeSubstitutionsType = '',
```
ivanpauno, line 46: this should be the same:
```suggestion
        namespace: SomeSubstitutionsType = None,
```

ivanpauno (approved):
This seems to fix the bug without breaking API, thanks @jacobperron !

launch_ros/launch_ros/actions/node.py
```diff
@@ -348,7 +348,7 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
                 validate_node_name(self.__expanded_node_name)
             self.__expanded_node_name.lstrip('/')
             expanded_node_namespace = None
-            if self.__node_namespace is not None:
+            if self.__node_namespace is not None and self.__node_namespace != '':
```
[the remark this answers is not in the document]
jacobperron: Done: 7ddcc1a

launch_ros/examples/lifecycle_pub_sub_launch.py
jacobperron: 7ddcc1a

launch_ros/launch_ros/actions/lifecycle_node.py
jacobperron: 7ddcc1a

launch_ros/examples/lifecycle_pub_sub_launch.py
jacobperron: 7ddcc1a

jacobperron:
@ivanpauno I've addressed your comments, and also fixed a bug launching LifecycleNode without a namespace (7ddcc1a). The example was not working, but now it is.

ivanpauno (approved):

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12617/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7579/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10325/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12543/)

Issue #185, opened 2020-10-19
[Foxy]: Can't launch a ComposableNodeContainer with the latest Foxy packages (0.10.3)


clalancette:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - Foxy debian packages
- Version or commit hash:
  - ros-foxy-launch-ros_0.10.3
- DDS implementation:
  - Fast-RTPS
- Client library (if applicable):
  - rclpy

#### Steps to reproduce issue

I have a fairly straightforward launch file:

```
from launch import LaunchDescription
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode

def generate_launch_description():
    params = {
        'target_frame': 'base_footprint',
        'min_height': 0.10,
        'max_height': 2.10,
        'angle_min': -2.356194490192345,
        'angle_max': 2.356194490192345,
        'angle_increment': 0.005759586531581288,
        'range_min': 0.4,
        'range_max': 200.0,
        'use_inf': False,
    }

    return LaunchDescription([
        ComposableNodeContainer(
            name='velodyne_container',
            namespace='',
            package='rclcpp_components',
            executable='component_container',
            composable_node_descriptions=[
                ComposableNode(package='pointcloud_to_laserscan',
                               plugin='pointcloud_to_laserscan::PointCloudToLaserScanNode',
                               name='pointcloud_to_laserscan',
                               remappings=[('cloud_in', 'velodyne_points')],
                               parameters=[params, {'use_sim_time': True}])
            ],
            output='both',
        ),
    ])
```

#### Expected behavior

Container successfully launches

#### Actual behavior

Throws an exception:

```
$ ros2 launch bug-launch.py 
[INFO] [launch]: All log files can be found below /home/ubuntu/.ros/log/2020-10-19-20-46-35-369504-ros2-focal-1778566
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [component_container-1]: process started with pid [1778568]
Task exception was never retrieved
future: <Task finished name='Task-5' coro=<LaunchService._process_one_event() done, defined at /opt/ros/foxy/lib/python3.8/site-packages/launch/launch_service.py:271> exception=InvalidServiceNameException("Invalid service name: topic name must not contain characters other than alphanumerics, '_', '~', '{', or '}':\n  '<node_namespace_unspecified>/velodyne_container/_container/load_node'\n   ^")>
Traceback (most recent call last):
[...]

Issue #123, opened 2020-02-12
Remove deprecated Node API after Foxy release


jacobperron:
#### Feature description

API was deprecated in https://github.com/ros2/launch_ros/pull/122.
After Foxy is released we can remove the deprecated API in preparation for the next major release.

ivanpauno:
Fixed in https://github.com/ros2/launch_ros/pull/190.

PR #189, opened 2020-10-20
Make name and namespace mandatory in ComposableNodeContainer, remove deprecated alternatives

ivanpauno:
Based on discussion in https://github.com/ros2/launch_ros/issues/185, the name and namespace are mandatory.

I think that the intention was to check that, but it was overlooked when adding support for the deprecated name of the argument.
This delete the deprecated name (they were deprecated before Foxy release), and now it's checked.

### DIFF
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -16,7 +16,6 @@
 
 from typing import List
 from typing import Optional
-import warnings
 
 from launch.action import Action
 from launch.launch_context import LaunchContext
@@ -33,10 +32,8 @@ class ComposableNodeContainer(Node):
     def __init__(
         self,
         *,
-        name: Optional[SomeSubstitutionsType] = None,
-        namespace: Optional[SomeSubstitutionsType] = None,
-        node_name: Optional[SomeSubstitutionsType] = None,
-        node_namespace: Optional[SomeSubstitutionsType] = None,
+        name: SomeSubstitutionsType,
+        namespace: SomeSubstitutionsType,
         composable_node_descriptions: Optional[List[ComposableNode]] = None,
         **kwargs
     ) -> None:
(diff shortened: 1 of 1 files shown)


jacobperron:
This partially addresses https://github.com/ros2/launch_ros/issues/123.

jacobperron (approved):

ivanpauno:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12821/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7778/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10535/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12766/)

hidmic (approved):
I'm not uber convinced we should be making name and namespace mandatory anywhere BUT we control this container executable so I guess it's fine.

LGTM pending green CI !

clalancette:
One other thing to note; if we go forward with this, please add a note to https://index.ros.org/doc/ros2/Releases/Release-Galactic-Geochelone/ , since this is a change in behavior.

ivanpauno:
> One other thing to note; if we go forward with this, please add a note to https://index.ros.org/doc/ros2/Releases/Release-Galactic-Geochelone/ , since this is a change in behavior.

In Foxy, it was already mandatory both the name and the namespace.
The only difference is that the error wasn't handled correctly and `rcl` was showing an error (instead of launch detecting the error), but it was impossible to use the action correctly without passing both the name and namespace explicitly (and not `None`).

See [eloquent code](https://github.com/ros2/launch_ros/blob/0b889397c0a9aad00aa1921e861fde0c9a95e107/launch_ros/launch_ros/actions/composable_node_container.py#L37-L38).

jacobperron:
If anything, we should add a patch to Foxy to produce a better error message from the ComposableNodeContainer `__init__`.

jacobperron:
See https://github.com/ros2/launch_ros/pull/191

PR #190, opened 2020-10-20
Remove constructors arguments deprecated since Foxy

ivanpauno:
Cleaning up things deprecated before Foxy release.

Follow up of https://github.com/ros2/launch_ros/pull/189.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -25,8 +25,6 @@ from typing import Text  # noqa: F401
 from typing import Tuple  # noqa: F401
 from typing import Union
 
-import warnings
-
 from launch.action import Action
 from launch.actions import ExecuteProcess
 from launch.frontend import Entity
@@ -72,12 +70,9 @@ class Node(ExecuteProcess):
     def __init__(
         self, *,
         executable: Optional[SomeSubstitutionsType] = None,
-        node_executable: Optional[SomeSubstitutionsType] = None,
         package: Optional[SomeSubstitutionsType] = None,
         name: Optional[SomeSubstitutionsType] = None,
         namespace: Optional[SomeSubstitutionsType] = None,
-        node_name: Optional[SomeSubstitutionsType] = None,
-        node_namespace: SomeSubstitutionsType = None,
         exec_name: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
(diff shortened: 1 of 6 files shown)


jacobperron (commented):
Based on CI results, it looks like some tests need to be updated too.

hidmic (approved):
LGTM pending green CI

ivanpauno:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12887/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7842/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10604/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12830/)

jacobperron (approved):

cottsay:
This change regressed the performance tests with a cryptic exception: http://build.ros2.org/view/Rci/job/Rci__nightly-performance_ubuntu_focal_amd64/112/

```
usage: launch_test.py [-h] [--package-name PACKAGE_NAME] [-v] [-s]
                      [--junit-xml XMLPATH]
                      launch_test_file
                      [launch_arguments [launch_arguments ...]]
launch_test.py: error: 'NoneType' object is not iterable
```

jacobperron:
@cottsay 
hm, there should have been deprecation warnings visible if using `node_executable` prior to this change.

cottsay:
> there should have been deprecation warnings

There were, but it didn't make the build go yellow and they were lost in the wall of text coming out of `colcon test`...

clalancette:
> There were, but it didn't make the build go yellow and they were lost in the wall of text coming out of `colcon test`...

I think this is an instance where https://github.com/ros2/ci/pull/509 would have helped.  It's on the long-term TODO list, but we need to spend some time on it.

Separately, I will say that that error message is hard for the end user to tell what happened.  @ivanpauno @jacobperron Could we throw a nicer exception for unknown keyword arguments?

ivanpauno:
> Separately, I will say that that error message is hard for the end user to tell what happened. @ivanpauno @jacobperron Could we throw a nicer exception for unknown keyword arguments?

The error message wasn't obvious because of a mistake in this patch.
https://github.com/ros2/launch_ros/pull/197 is fixing that.

launch_ros/launch_ros/actions/node.py
```diff
@@ -72,12 +70,9 @@ class Node(ExecuteProcess):
     def __init__(
         self, *,
         executable: Optional[SomeSubstitutionsType] = None,
```
jacobperron, line 72: executable should not be optional.

jacobperron (commented):
I found another bug.

launch_ros/launch_ros/actions/node.py
jacobperron: nevermind, addressed in https://github.com/ros2/launch_ros/pull/197/
Sorry for the noise.

PR #186, opened 2020-10-19
[foxy] Assume root namespace if not provided in ComposableNodeContainer

jacobperron:
Fixes a regression introduced in #179

Fixes #185.

We apply the same logic in LifecycleNode to maintain backwards compatibility in Foxy.

PR #191, opened 2020-10-21
[foxy] Improve error message if ComposableNodeContainer 'name' or 'namespace' arg is None

jacobperron:
This restores the original behavior in Eloquent and Foxy, but with an improved error message.

The error was accidentally removed in #186.

### DIFF
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -77,6 +77,10 @@ class ComposableNodeContainer(Node):
                 )
             namespace = node_namespace
 
+
+        if name is None:
+            raise RuntimeError("'name' is a required argument")
+
         if namespace is None:
             raise RuntimeError("'namespace' is a required argument")
 
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -77,7 +77,10 @@ class ComposableNodeContainer(Node):
                 )
             namespace = node_namespace
 
-        if not namespace:
+        if namespace is None:
+            raise RuntimeError("'namespace' is a required argument")
+
+        if namespace == '':
             namespace = '/'
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__composable_node_descriptions = composable_node_descriptions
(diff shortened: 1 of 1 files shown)


launch_ros/launch_ros/actions/composable_node_container.py
```diff
@@ -77,7 +77,10 @@ def __init__(
                 )
             namespace = node_namespace
 
-        if not namespace:
+        if namespace is None:
+            raise RuntimeError("'namespace' is a required argument")
```
ivanpauno, line 81: same check for `name`?

jacobperron: See 7c11b83

ivanpauno (approved):

launch_ros/launch_ros/actions/composable_node_container.py
```diff
@@ -77,7 +77,13 @@ def __init__(
...
             namespace = node_namespace
 
-        if not namespace:
+        if name is None:
+            raise RuntimeError("'name' is a required argument")
+
+        if namespace is None:
+            raise RuntimeError("'namespace' is a required argument")
+
+        if namespace == '':
```
clalancette, line 86: Just out of curiosity, what happens if `name == ''`?

ivanpauno: I think that that fails, because `Node` action is checking the validity of the name before launching the process.

The semantics of `namespace` is a bit difference:
- A relative `namespace` in launch means that it will be affected by previous `PushRosNamespace` actions.
- An absolute one no.

`Node` will evaluate an actual namespace based on the previous `PushRosNamespace` and the user provided argument, and then validate.
In Foxy `namespace=''` was behaving like no passing a remapping rule, thus that extra check is needed.

clalancette: Makes sense, thanks.

clalancette (approved):

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/12826/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/7783/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/10540/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/12771/)

PR #205, opened 2020-11-19
[dashing] Use yaml.safe_load

jacobperron:
This is a kind of backport of #175. Though #175 doesn't apply due to significant changes to the test code since Dashing, I think the idea is the same.


mjeronimo (approved):

nuclearsandwich:
The test failure doesn't appear in the Dashing devel job [Build Status](http://build.ros2.org/job/Ddev__launch_ros__ubuntu_bionic_amd64/37/)

jacobperron:
Yeah, I guess we can't actually use safe_load in this case. I think there is nothing to backport.

PR #207, opened 2020-11-20
Fix docblock in LoadComposableNodes

jacobperron:


mjeronimo (approved):

jacobperron:
For posterity, the docblock should have been updated as part of #166

PR #215, opened 2021-01-25
Execute process refactor

roger-strain:
This is an initial commit toward the process described in ros2/design#272.

Part of implementation of refactor for ros2/launch#114, and depends on ros2/launch#454; unit tests will fail without those changes.


roger-strain:
This has taken longer to get in than I would like, but such is the way of things sometimes. A few questions/notes as I was getting this together:

- In the refactoring, I had carried `arguments` through to the `Node` level, but I wonder now if they might better belong to the `Executable`; they seem to be more related to that concept, whereas `parameters` tie neatly to `Node`.
- We discussed adding support for multiple `Node`s in a single `Executable`, but fully implementing that still seems a bit tricky, at least as part of this PR. I'm starting to think we might be better served by getting this one done, with only having one `Node` truly supported, then come back and add in multiple-node support later. (As an example, we had discussed how to pass down parameters to individual nodes when multiple exist, but then I stumbled when trying to figure out how to support renaming multiple nodes in a single command line.)
- Some of the composable node unit tests didn't seem to be working, but I'm not sure it was due to these changes. Are they all in good shape? I'll keep digging into them if so.
- There probably are plenty of other discussion points, but I'll wait for an initial review before diving too far into other things.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,324 @@
...
+        """Getter for expanded_node_name."""
+        return self.__expanded_node_name
+
+    @property
+    def expanded_node_namespace(self):
+        """Getter for expanded_node_namespace."""
+        return self.__expanded_node_namespace
+
+    @property
+    def expanded_parameter_arguments(self):
```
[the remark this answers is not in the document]
roger-strain: In all honesty, I'm not sure what direction they should be sorted into. For most of this, I've been trying to keep things generally like they were previously. I maybe added a couple of accessors that weren't needed? If you have any guidance about which ones to remove/streamline, I'll gladly do so.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,324 @@
...
+                {'ros__parameters': params}
+            }
+            yaml.dump(param_dict, h, default_flow_style=False)
+            return param_file_path
+
+    def _get_parameter_rule(self, param: 'Parameter', context: LaunchContext):
+        name, value = param.evaluate(context)
+        return f'{name}:={yaml.dump(value)}'
+
+    def prepare(self, context: LaunchContext, executable: Executable) -> None:
```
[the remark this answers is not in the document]
roger-strain: Possibly so; those are going to come into play more strongly with the next couple of PRs/commits related to this as we get into composable/lifecycle concepts. At the moment, they were just placeholders with no functionality. I'll see about adding in the appropriate calls. As you noted elsewhere, at the moment they call for an `Action`, whereas here I have `Executable` instead. I'll need to sanity check what's really necessary, and may have to add more plumbing depending on that result.

hidmic:
> In the refactoring, I had carried arguments through to the Node level, but I wonder now if they might better belong to the Executable; they seem to be more related to that concept, whereas parameters tie neatly to Node.

I'd say `arguments` belong to the `Executable` (and `RosExecutable`).

> but then I stumbled when trying to figure out how to support renaming multiple nodes in a single command line

See https://index.ros.org/doc/ros2/Tutorials/Node-arguments/#passing-remapping-arguments-to-specific-nodes.

---

BTW check your DCO!

launch_ros/launch_ros/descriptions/node.py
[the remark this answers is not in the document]
hidmic: Yes! A review of the call chain would be healthy.

launch_ros/launch_ros/descriptions/node.py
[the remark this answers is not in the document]
hidmic: > For most of this, I've been trying to keep things generally like they were previously.

Oof, and you did. It is a mess already. Nevermind. I guess it's the biggest red flag whenever I review your patches and it calls my attention, but it's not on you.

hidmic:
Overall, I think this is going in the right direction. I'd revisit the architecture, make sure that entity setup via descriptions and/or traits  is clean (no crust like redundant arguments, no issues like unnecessary coupling). I'll try to do a deeper review later this week to both feature branches.

roger-strain:
> BTW check your DCO!

I'm going to feel like a fool for asking, but "DCO"?

roger-strain:
@hidmic And here's the corresponding changes on this side. I've tried to simplify and clean up some of the argument lists. I was trying to remove some of the extra layers of calls going on, but had to reintroduce one to make the unit tests happy; they wanted certain parts of the substitutions to happen, but I couldn't let the full preparation go through or it would have needed a lot more mock structure in the tests than I thought was really warranted.

If this looks reasonable, I can go ahead and look at the Composable/Lifecycle stuff next. If there are other things you want to see here first, though, please let me know. The more specific the better. :)

roger-strain:
@hidmic Just wanted to check in and see if there's anything else I can do to help on this at the moment.

launch_ros/launch_ros/actions/node.py
```diff
@@ -138,43 +114,23 @@ def __init__(
...
+        self.__node_desc = NodeDescription(node_name=name, node_namespace=namespace,
+                                           parameters=parameters, remappings=remappings)
+        ros_exec_kwargs = {'additional_env': additional_env, 'arguments': arguments}
+        self.__ros_exec = RosExecutable(package=package, executable=executable,
+                                        nodes=[self.__node_desc], **ros_exec_kwargs)
+        super().__init__(process_description=self.__ros_exec, **kwargs)
 
-        self.__substitutions_performed = False
+    def prepare(self, context: LaunchContext):
+        self.__node_desc.prepare(context, self.__ros_exec, self)
```
hidmic, line 129: @roger-strain why preparing the node description here? Won't `ExecuteLocal.prepare()` invoke `RosExecutable.prepare()`, which in turn will invoke `NodeDescription.prepare()`? That'd make sense to me.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,327 @@
...
+        self, *,
+        node_name: Optional[SomeSubstitutionsType] = None,
+        node_namespace: Optional[SomeSubstitutionsType] = None,
+        parameters: Optional[SomeParameters] = None,
+        remappings: Optional[SomeRemapRules] = None,
+        traits: Optional[Iterable[NodeTrait]] = None,
+        **kwargs
+    ) -> None:
+        """
+        Construct an Node description.
```
hidmic, line 83: @roger-strain nit:
```suggestion
        Construct a Node description.
```

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,327 @@
...
+
+        Many arguments are passed eventually to
+        :class:`launch.actions.ExecuteProcess`, so see the documentation of
```
hidmic, line 86: @roger-strain did you mean to `launch.actions.ExecuteLocal`? At any rate, it shouldn't specify an executor. Perhaps documentation needs a revisit?

test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
```diff
@@ -51,5 +51,5 @@ def test_node_name():
         namespace='my_ns',
     )
     lc = LaunchContext()
-    node_object._perform_substitutions(lc)
+    node_object._Node__node_desc._perform_substitutions(lc, [])
```
hidmic, line 54: @roger-strain is it necessary to access a "private" attribute like this? Isn't the functionality under test observable via public API? Same everywhere else below.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,327 @@
...
+            for src, dst in self.__expanded_remappings:
+                cmd_ext.extend(['-r', f'{src}:={dst}'])
+            cmd.extend([normalize_to_list_of_substitutions(x) for x in cmd_ext])
+        # Prepare the ros_specific_arguments list and add it to the context so that the
+        # LocalSubstitution placeholders added to the the cmd can be expanded using the contents.
+        ros_specific_arguments: Dict[str, Union[str, List[str]]] = {}
+        if self.__node_name is not None:
+            ros_specific_arguments['name'] = '__node:={}'.format(self.__expanded_node_name)
+        if self.__expanded_node_namespace != '':
+            ros_specific_arguments['ns'] = '__ns:={}'.format(self.__expanded_node_namespace)
```
hidmic, line 316: @roger-strain meta: this will probably need a revisit. A blanket `__node:=` remapping will apply to all nodes in the same executable. Probably not what a user would intend in this case. Same for `__ns:=`.

launch_ros/launch_ros/descriptions/node_trait.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Note that this class provides no functionality itself, and is used
+        as a base class for traits which provide actual functionality. As
+        such, the base class itself should not be directly used by application
+        code.
+        """
+        pass
+
+    def prepare(self, node: 'Node', context: LaunchContext, action: Action):
+        """Perform any actions necessary to prepare the node for execution."""
```
hidmic, line 49: @roger-strain missing `pass` after docstring?

hidmic (commented):
@roger-strain my apologies for not circling back sooner. Just did another pass. Check your failing PR job!

launch_ros/launch_ros/actions/node.py
roger-strain: On reflection, I believe you're correct. I think this was probably added earlier in the development cycle, before the `RosExecutable` portion came in.

launch_ros/launch_ros/descriptions/node.py
roger-strain: Definitely was a copy/paste leftover from something. I've removed the clause entirely. I think the rest of it is okay, but if there are other deficiencies please let me know.

test_launch_ros/test/test_launch_ros/actions/test_lifecycle_node.py
roger-strain: It actually wasn't (easily) available without making some assumptions about data types that I didn't want to make. I've added a new getter on the `actions.Node` for `ros_exec`, which will allow us to easily get access to the ROS-style executable, rather than the more generic `Executable` description, which didn't provide the things we needed.

launch_ros/launch_ros/descriptions/node.py
roger-strain: This is probably a carry-over from the earlier design which assumed one `Node` per executable process. I'm looking at the docs here (https://docs.ros.org/en/foxy/Guides/Node-arguments.html) and it looks like I need to prefix things with the "old" name of the node. That's something that I don't believe we've captured in the `descriptions.Node` structure to this point. I'm adding a new parameter to capture `original_node_name`; if it's not set, there's nothing we can do but let the global apply. (This will actually work out okay, though, because I think it maintains current functionality for backward-compatibility.)

I'm now having another thought, though, which is related to this. For "normal" (read: Not node name/namespace) parameters, the docs show the ability to prefix individual parameters with the (new) node name. I'm guessing this is something we'll need to bake in at this layer as well? How does that come into play if we're passed or creating parameter dictionary files? Not sure how far down this rabbit hole is the responsibility of this layer, vs telling the developer they need to specify things with the appropriate prefix. Would love your thoughts on which parts to tackle internally.

roger-strain:
> Check your failing PR job!

About to push up a new version that includes several of the changes discussed above. Something about my local unit test setup isn't capturing things the same as the build server does, so I may end up having to do a few passes to get that sorted, apologies for that.

(If you have any tips on setting up a good environment to make sure the tests are working locally, I'd love to hear it.)

launch_ros/launch_ros/descriptions/node.py
hidmic: I'll start by saying that some of these rough edges when dealing with remapping predate your work, and that I'm not pushing to get them all addressed here.

I don't think `launch` should be mutating parameter files to propagate a remapping and such. That's on the underlying executable's plate at best. IMHO `launch` should simply provide a way to express the intent to remap. 

For executables that bear many nodes, the current command line interface allows both node-specific and blanket configurations. To support the latter, we can simply let the `ROSExecutable` action take extra command line arguments. To support the former, `Node` descriptions have to bear the original name, that's right. We can allow a single `Node` description to not specify it for convenience, in which case I think the `ROSExecutable` action should enforce only one such description is given.

roger-strain:
@hidmic 
> Check your failing PR job!

Oh, just realized some of the failures here may be because this PR depends on the PR over in `launch`, which isn't merged yet. Is there any way to configure the build/test here to use that branch instead of master? Or do we just need to wait until that one goes through before we can get a better view of how this one is doing build/test-wise?

launch_ros/launch_ros/descriptions/node.py
roger-strain: I think I ended up trying to make this more complicated than it needed to be. If we're setting aside the dictionary files, then in all cases we know what the applied node name should be by the time we're setting up command line remaps, so I can just add that as the normal way of doing it.

Here's a question on parameters, though. Is there such a thing as a node-specific (non-remapping) parameter? If so, I don't see it in the sample documentation I've been referring to.

roger-strain:
@hidmic Hey, just wanted to check back in. I know we're still pending the `launch` PR before we can really confirm build/test behavior on this one. I've pushed some other updates here in the meantime to address other comments. Is there anything else you'd like to see done here in the meantime, or should we just hold this until the other one gets sorted?

I've been considering moving forward with the Composable and Lifecycle portions of the refactor, but didn't want to get too much pending at once.

mlanting:
@hidmic I've merged our changes back on top of master but Jenkins is running into errors because it doesn't seem to have the changes that were merged into the launch repository. Is there something I need to do to tell it to update?
  
Also, the ros2launch_security repo will require a couple changes to work with the refactored code. I can submit a PR, but I thought this might warrant a bit of discussion first and want to make sure the discussion happens here instead of being split across two PR's.           

The ros2launch_security plugin was based on an extension to launch_ros called NodeActionExtension. However, this refactor deprecates the Node action in favor of simply passing a RosExecutable description to a standard ExecuteLocal action. The RosExecutable description contains one or more Node descriptions. I've moved NodeActionExtension into the new Node description class and changed the function signature of `prepare_for_execute` to take a RosExecutable and a Node description instead of a Node action.

I noticed that the NodeActionExtension interface looks very similar to the NodeTrait description we included in anticipation of adding support for things like composable nodes and lifecycle nodes that might need additional setup steps before being fully launched. Would it be worth wrapping the two concepts (node traits and node action extensions) into one? Security as a node trait makes some sense, but I'm not so sure about composable/lifecycle nodes as plugins/extensions. Still, the interfaces are similar enough (almost identical) that I figure it's worth mentioning.

clalancette:
> @hidmic I've merged our changes back on top of master but Jenkins is running into errors because it doesn't seem to have the changes that were merged into the launch repository. Is there something I need to do to tell it to update?

No, we need to do a new release of `launch` in order to get those changes out.  I'll go ahead and do that.

hidmic:
> Would it be worth wrapping the two concepts (node traits and node action extensions) into one? 

Ideally, consolidating would be best. Those extensions are plugin-based and implicit though. Not sure what the best path forward is here. Perhaps some traits can also be plugin-based and implicit? I'm open to ideas. FYI @gbiggs.

hidmic:
@ros-pull-request-builder retest this please.

hidmic:
@mlanting there's still a bunch test failures in that Rpr job.

mlanting:
@hidmic Yeah, I still had a few I was working through but wanted to get some fixes pushed up before I left for vacation. I've fixed a few more of them. The remaining failures that I'm seeing are all failing when the testing infrastructure tries to call tests in the `talker_listener_launch_test.py` file in `launch_testing_ros`.

It's giving type errors along the lines of `TypeError: test_fuzzy_data() missing 3 required positional arguments: 'launch_service', 'listener', and 'proc_output'` I wonder if you or anyone else might have some insights as to what the issue could be here? I'll keep looking into it.

mlanting:
I created a pull request (https://github.com/ros2/launch/pull/587) with a fix for a test function in the launch repo to address the remaining fix.

For an intermittent timer failure I was seeing, I simply increased the tolerance a bit. The tolerance was originally 0.1 seconds. The largest difference I saw in the few times I occurred was 0.37 seconds, so I just increased the tolerance to 0.5.

While running the tests repeatedly trying to get the timer to fail, I also saw the error described here on a couple occasions: https://github.com/ros2/launch_ros/issues/280. I did not make any attempts to address it, but I do want to note that the issue can apparently happen on Ubuntu and is not just limited to Windows.

mlanting:
@hidmic friendly ping

mlanting:
@hidmic friendly ping

launch_ros/launch_ros/descriptions/ros_executable.py
```diff
@@ -0,0 +1,97 @@
...
+    """Describes an executable with ROS features which may be run by the launch system."""
+
+    def __init__(
+            self, *,
+            executable: Optional[SomeSubstitutionsType] = None,
+            package: Optional[SomeSubstitutionsType] = None,
+            nodes: Iterable[Node] = None,
+            ros_arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
+            arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
+            **kwargs
```
hidmic, line 48: @mlanting meta: I wonder if it would make sense to allow for "global" parameters and remapping rules here (i.e. without a node name prefix to scope them).

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,477 @@
...
+        node.
+        Multiple dictionaries/files can be passed: each file path will be
+        passed in in order to the node (where the last definition of a
+        parameter takes effect).
+
+        :param: node_name the name of the node
+        :param: node_namespace the ROS namespace for this Node
+        :param: original_node_name the name of the node before remapping; if not specified,
+            remappings/parameters (including node name/namespace changes) may be applied
+            to all nodes which share a command line executable with this one
```
hidmic, line 176: @mlanting hmm, this is a bit clunky. Thinking out loud, perhaps `node_name` and `node_namespace` shouldn't be prescriptive but descriptive. That is, to be mandatory and expected to match what the actual node already defines, leaving name and namespace remapping to be done separately e.g.:

```python
Node(
   name = 'actual_name',
   namespace = 'actual_namespace'
   remappings = Node.naming(
        name = 'my_new_name', 
        namespace = 'my_new_namespace'
   ) + [('/chit', '/chat')],
)
```

CC @wjwwood @jacobperron for feedback.

launch_ros/launch_ros/descriptions/ros_executable.py
hidmic: @mlanting even more so, there was (and probably still is) some confusion around parameter file formats and how do each apply to `launch_ros` actions, mainly owing to the fact that the file format expected by `rcl` accounts for multiple nodes in the same process whereas the file format expected by `rclcpp_components` is bound to a single node.

We could potentially sort that out. `RosExecutable` would accept the former while `Node` descriptions would accept the latter.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,477 @@
...
+
+import yaml
+
+from .node_trait import NodeTrait
+from ..parameter_descriptions import Parameter
+from ..parameters_type import SomeParameters
+from ..remap_rule_type import SomeRemapRules
+
+
+class NodeActionExtension:
```
hidmic, line 73: > Would it be worth wrapping the two concepts (node traits and node action extensions) into one? Security as a node trait makes some sense, but I'm not so sure about composable/lifecycle nodes as plugins/extensions.

@mlanting circling back to this. These extensions were added specifically to support [`osrf/ros2launch_security`](https://github.com/osrf/ros2launch_security). Looking at those extensions, it'd seem like hoisting this to the `RosExecutable` action level would make the most sense (since `--enclave` is a process-wide ROS argument).

test_launch_ros/test/test_launch_ros/actions/test_node.py
```diff
@@ -184,7 +186,8 @@ def test_launch_node_with_parameter_descriptions(self):
         )
         self._assert_launch_no_errors([node_action])
 
-        expanded_parameter_arguments = node_action._Node__expanded_parameter_arguments
+        for node in node_action.ros_exec.nodes:
+            expanded_parameter_arguments = node.expanded_parameter_arguments
```
hidmic, line 190: @mlanting hmm, should these be extending the `expanded_parameter_arguments` list instead of overwriting? Same below.

hidmic (commented):
@mlantting did another pass. I'm sorry for the delay. I've been swamped.

launch_ros/launch_ros/descriptions/node.py
```diff
@@ -0,0 +1,477 @@
...
+    """Describes a ROS node."""
+
+    UNSPECIFIED_NODE_NAME = '<node_name_unspecified>'
+    UNSPECIFIED_NODE_NAMESPACE = '<node_namespace_unspecified>'
+
+    def __init__(
+        self, *,
+        node_name: Optional[SomeSubstitutionsType] = None,
+        node_namespace: Optional[SomeSubstitutionsType] = None,
+        original_node_name: Optional[SomeSubstitutionsType] = None,
```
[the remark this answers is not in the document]
jacobperron: FWIW, we deprecated (and then removed) the arguments with the "node_" prefix in the existing Node action.

launch_ros/launch_ros/descriptions/node.py
jacobperron: Either way it seems a bit brittle to me if we expect the name in launch to match the name in the code. You have to remember to update both places; and worse if the launch file is not maintained by the node author (e.g. in some end-users package).

However, I don't see another way around this problem besides introducing some far-fetched mechanism to query the coded node name from launch. I like @hidmic's suggestion since I think it will result in less hidden surprises for users (i.e. it avoids accidentally renaming a bunch of nodes in a process to the same name). But I'm also worried that being strict about providing the correct node name will be frustrating for users, who currently don't have to do this.

I'm torn.

launch_ros/launch_ros/descriptions/ros_executable.py
```diff
@@ -0,0 +1,97 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# DISTRIBUTION A. Approved for public release; distribution unlimited.
+# OPSEC #4584.
+#
+# Delivered to the U.S. Government with Unlimited Rights, as defined in DFARS
+# Part 252.227-7013 or 7014 (Feb 2014).
```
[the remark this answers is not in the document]
mlanting: You're correct, I left this in by mistake. I'll remove it.

launch_ros/launch_ros/actions/node.py
```diff
@@ -197,47 +123,20 @@ def __init__(
             passed to the node as ROS remapping rules
         :param: ros_arguments list of ROS arguments for the node
         :param: arguments list of extra arguments for the node
+        :param additional_env: Dictionary of environment variables to be added. If env was
+            None, they are added to the current environment. If not, env is updated with
+            additional_env.
```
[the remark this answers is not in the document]
mlanting: I like the idea of pulling out the keywords args and passing them via a pared-down kwargs variable. I used the `inspect` module to get the available keywords from the `Executable` class to filter out the appropriate pairs.

launch_ros/launch_ros/actions/node.py
```diff
@@ -197,47 +123,20 @@ def __init__(
...
-        self.__final_node_name = None  # type: Optional[Text]
-        self.__expanded_remappings = None  # type: Optional[List[Tuple[Text, Text]]]
-
-        self.__substitutions_performed = False
-
-        self.__logger = launch.logging.get_logger(__name__)
+        self.__node_desc = NodeDescription(node_name=name, node_namespace=namespace,
+                                           parameters=parameters, remappings=remappings)
+        ros_exec_kwargs = {'additional_env': additional_env}
+        self.__ros_exec = RosExecutable(package=package, executable=executable,
```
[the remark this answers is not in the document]
mlanting: I'll make the change.

launch_ros/launch_ros/actions/node.py
```diff
@@ -197,47 +123,20 @@ def __init__(
...
-        self.__expanded_node_name = self.UNSPECIFIED_NODE_NAME
-        self.__expanded_node_namespace = self.UNSPECIFIED_NODE_NAMESPACE
-        self.__expanded_parameter_arguments = None  # type: Optional[List[Tuple[Text, bool]]]
-        self.__final_node_name = None  # type: Optional[Text]
-        self.__expanded_remappings = None  # type: Optional[List[Tuple[Text, Text]]]
-
-        self.__substitutions_performed = False
-
-        self.__logger = launch.logging.get_logger(__name__)
+        self.__node_desc = NodeDescription(node_name=name, node_namespace=namespace,
```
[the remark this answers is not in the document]
mlanting: This class only exists for backward compatibility with the old node action which really only supported a single node and this was just an artifact of the way we were thinking about the class when it was written. I've left it as is, but I don't have strong feelings either way and can change it if you'd prefer the other way.

launch_ros/launch_ros/descriptions/node.py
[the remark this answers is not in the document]
mlanting: I've removed the node_ prefixes

mlanting:
@hidmic friendly ping

mlanting:
@hidmic friendly ping. Can we get this merged in soon?

mlanting:
Does anyone understand the error the checks are throwing and how to fix it? Is it just confused by the change of base branch?

clalancette:
I've seen this before, but I still don't understand what is happening with it.  I suspect that if you rebase this onto the latest rolling (rather than merge rolling in), it will be happier, but I'm not sure.

mlanting:
I just tried doing the rebase instead of a merge, but it doesn't seem to have helped. Good thought though. Is this a blocker for getting this merged in, or can we move on that soon?

clalancette:
> I just tried doing the rebase instead of a merge, but it doesn't seem to have helped.

Oh, actually, looking at the error again, I wonder if this is something different.  Can you try pushing the `rolling` branch to the https://github.com/roger-strain/launch_ros fork?

mlanting:
Oh good catch. I thought that would have been done when I did the fetch upstream from the github UI to bring the PR up to date but I guess it didn't make the new branch. I've done that now.

clalancette:

clalancette:
This still didn't work.  I think that this is caching problem on the buildfarm side, due to some recent changes we made.  This should be cleared up tonight, I believe.

That said, while we like to have the Rpr jobs green, it isn't totally blocking for this.  What is blocking is getting this approved in review and also running the https://ci.ros2.org jobs for all platforms.

roger-strain:
@clalancette @hidmic I think this PR has fallen through the cracks, and I wanted to check in with you guys on whether it's something we should continue to put energy into. This was intended to be a stepping stone towards some other design goals, but I'm honestly concerned at this point that other developments may have impacted what we were trying to do here. I'd appreciate your candid thoughts on how to best move forward at this point.

vknisley-swri:
This PR is a couple years old, but I've rebased and resolved some warnings. Is this functionality that we would still like to include in launch_ros?

vknisley-swri:
@clalancette @hidmic I just wanted to check whether this PR is still something we would like to include in launch_ros. It was originally opened a few years ago, but I've rebased it to include the latest rolling changes (from approximately 3 weeks ago). Thanks!

PR #216, opened 2021-01-26
Add options extensions to ros2launch and extensibility to the node action

gbiggs:
This PR is a response to #180. In that PR, a request was made that extensibility be added to the launch system and the `--secure` option be added as an extension. This PR provides the extensibility support. The security extension is provided in [a PR on a new repository](https://github.com/osrf/ros2launch_security/pull/1).

It adds:
- The ability to load extensions that provide additional options to the `ros2 launch` command. For example, an extension can add an option `--secure`. Aside from getting the chance to add as many options as desired to the command line flags, extensions are given the chance to execute code prior to the `LaunchService` starting, prior to the launch being executed, and after the launch process completes (i.e. on shutdown?).
- The ability to load extensions that add additional command line flags to the command lines that launch nodes, and execute code just before the node is launched. This is a short-term solution to the need for extensibility. In the long term we plan to refactor all of `launch` to be more amenable to extensibility.

For an example, see [this PR](https://github.com/osrf/ros2launch_security/pull/1) which adds a security extension.

Once both PRs are accepted, the [ros2launch_security repository](https://github.com/osrf/ros2launch_security) will need to be transferred to the security WG.


gbiggs:
Ready for review.

kyrofa:
@gbiggs thank you for doing this. I will leave a code review to more qualified people, but I'll say we do believe this covers the functionality we would need for what we're doing.

gbiggs:
@kyrofa Let us know who you want to set as maintainer, and if the security WG wants to take the plugins one.

kyrofa:
@gbiggs I can take it and add the necessary people if that works.

gbiggs:
Great, thanks! I've added you as maintainer to the new package.

launch_ros/launch_ros/actions/node.py
```diff
@@ -14,8 +14,10 @@
 
 """Module for the Node action."""
 
+import importlib_metadata
```
Arnatious, line 17: Third party package, should be in the same block as `yaml`

launch_ros/launch_ros/actions/node.py
```diff
@@ -14,8 +14,10 @@
...
 import os
 import pathlib
+from packaging.version import Version
```
Arnatious, line 20: Third party package, should be in the same block as `yaml`

ros2launch/ros2launch/api/api.py
```diff
@@ -154,8 +163,19 @@ def launch_a_launch_file(*, launch_file_path, launch_file_arguments, debug=False
...
     ])
+    tails = {}
+    for name in sorted(option_extensions.keys()):
+        result = option_extensions[name].prelaunch(
+            launch_description,
+            args
+        )
+        launch_description = result[0]
+        if len(result) > 1:
+            tails[name] = result[1:]
```
Arnatious, line 174: Is this variable used solely for debugging? It isn't ever read from.

Arnatious (commented):
Looking great, only some minor nitpicks.

launch_ros/launch_ros/actions/node.py
gbiggs: Fixed in https://github.com/ros2/launch_ros/pull/216/commits/ded9e1d9f33cef40762a43310af29e9cb374cebe

launch_ros/launch_ros/actions/node.py
gbiggs: Fixed in https://github.com/ros2/launch_ros/pull/216/commits/ded9e1d9f33cef40762a43310af29e9cb374cebe

ros2launch/ros2launch/api/api.py
gbiggs: Yes, it was. Fixed in https://github.com/ros2/launch_ros/pull/216/commits/22d14af0b1bcd5f969531e210713df93db619046

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/ros-2-tsc-meeting-minutes-2021-02-18/19102/1

ros2launch/ros2launch/option/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+class OptionExtension:
+    """
+    The extension point for ros2 launch 'option' extensions.
+
+    The following properties must be defined:
+    * `NAME` (will be set to the entry point name)
+
+    The following methods may be defined:
+    * `add_arguments` to add arguments to the argparse parser
```
hidmic, line 28: @gbiggs nit: consider moving documentation to each method, and only listing which ones are required here.

ros2launch/ros2launch/option/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    def add_arguments(self, parser, cli_name, *, argv=None):
+        return
+
+    def prestart(self, options, args):
+        return
+
+    def prelaunch(self, launch_description, options, args):
+        return (launch_description,)
+
+    def postlaunch(self, launch_return_code, options, args):
```
hidmic, line 58: @gbiggs hmm, what is this `options` argument for? BTW it doesn't seem like `prelaunch` and `postlaunch` are provided with it in any call site.

launch_ros/launch_ros/actions/node.py
```diff
@@ -54,12 +54,51 @@
...
 
+class NodeActionExtension:
+    """
+    The extension point for launch_ros node action extensions.
+
+    The following properties must be defined:
+    * `NAME` (will be set to the entry point name)
+
+    The following methods may be defined:
+    * `command_extension` to extend the command used to launch the node's
```
hidmic, line 73: @gbiggs nit: consider moving documentation to each method, and only listing which ones are required here.

launch_ros/launch_ros/actions/node.py
```diff
@@ -408,6 +453,10 @@ def execute(self, context: LaunchContext) -> Optional[List[Action]]:
             ros_specific_arguments['name'] = '__node:={}'.format(self.__expanded_node_name)
         if self.__expanded_node_namespace != '':
             ros_specific_arguments['ns'] = '__ns:={}'.format(self.__expanded_node_namespace)
+        node_info = NodeActionExtension.NodeInfo(self.__package, self.__node_executable, self.node_name)
+        for extension in self.__extensions.values():
+            ros_specific_arguments = extension.execute(context, ros_specific_arguments, node_info)
```
hidmic, line 458: @gbiggs why not passing the node action itself instead of this ad-hoc data structure?

launch_ros/launch_ros/actions/node.py
```diff
@@ -342,6 +383,10 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
             raise
         self.__final_node_name = prefix_namespace(
             self.__expanded_node_namespace, self.__expanded_node_name)
+
+        for extension in self.__extensions.values():
+            self.cmd.extend(extension.command_extension(context))
```
hidmic, line 388: @gbiggs why a method to specifically extend the command? Isn't it possible to do so (even resolving substitutions) upon extension execution?

launch_ros/launch_ros/actions/node.py
```diff
@@ -432,3 +481,86 @@ def expanded_node_namespace(self):
     def expanded_remapping_rules(self):
         """Getter for expanded_remappings."""
         return self.__expanded_remappings
+
+    def _get_extensions(self):
```
hidmic, line 485: @gbiggs nit: consider making this a free function ~or `staticmethod`~ (better not, too much support logic inside the `Node` class) that returns the list of extensions.

launch_ros/launch_ros/actions/node.py
```diff
@@ -432,3 +481,86 @@ def expanded_node_namespace(self):
...
+            'Extension point is too old (%s), the extension requires '
+            "'%s'" % (extension_point_version, extension_version))
+
+    if extension_point_version >= next_extension_version:
+        raise PluginException(
+            'Extension point is newer (%s), than what the extension '
+            "supports '%s'" % (extension_point_version, extension_version))
+
+
+def get_upper_bound_caret_version(version):
```
hidmic, line 556: @gbiggs consider moving these functions to another file.

hidmic (commented):
First pass!

hidmic:
Also, this patch needs conflicts resolved.

gbiggs:
> Also, this patch needs conflicts resolved.

Rebased to fix this. I'll start addressing your other comments now.

launch_ros/launch_ros/actions/node.py
gbiggs: This function accesses the node's logger to output an error message that is specific to the function's internal implementation. I could refactor that out into returning an optional error message, but I'm wary of going to far for something that's meant to be temporary. How strongly do you feel about this?

launch_ros/launch_ros/actions/node.py
gbiggs: I can't remember specifically, but I suspect I was following the existing pattern of the `Node` class, which splits up building the command and executing it. Possibly it does things that way for debug output?

ros2launch/ros2launch/option/__init__.py
gbiggs: Done in 1e185752cbd8d7d69df2fd12ad46bdf199e54c28

ros2launch/ros2launch/command/launch.py
```diff
@@ -145,6 +150,7 @@ def main(self, *, parser, args):
         else:
             raise RuntimeError('unexpected mode')
         launch_arguments.extend(args.launch_arguments)
+
```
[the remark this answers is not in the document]
gbiggs: This vanished when I rebased.

ros2launch/ros2launch/option/__init__.py
gbiggs: It looks like it was vestigial of an earlier iteration of the changes. I've removed the arguments in 40ee929158c6299bca1136ab402e0e8df3a7d0d3.

launch_ros/launch_ros/actions/node.py
gbiggs: Done in 6754e9a366589dec1b1535eeef2e08ef765b1301.

launch_ros/launch_ros/actions/node.py
gbiggs: Done in 6754e9a366589dec1b1535eeef2e08ef765b1301.

launch_ros/launch_ros/actions/node.py
gbiggs: It seemed cleaner to pass in a minimal data structure rather than have extensions need to know about the entire `Node` action class and its internal variables.

gbiggs:
@hidmic Thanks for the comments. This is ready for another pass.

launch_ros/launch_ros/actions/node.py
hidmic: I see. I will point out that this data in particular (e.g. the package name) could be exposed via public API. You wouldn't have to access internal implementations details from the extension in that case.

launch_ros/launch_ros/actions/node.py
hidmic: Hmm, neither `NodeExtension.command_extension` nor `NodeExtension.execute` deal with execution. Both configure the command line, in direct and indirect ways respectively.

hidmic (commented):
Overall LGTM but for a few comments.

Once we settle those we need CI.

launch_ros/launch_ros/actions/node.py
hidmic: [`SecurityNodeActionExtension.execute`](https://github.com/osrf/ros2launch_security/pull/1/files#diff-d16580b357825167e86092c99e3945c9d131a8d1b8a4b4cf37de482298cef574R40-R44) could fetch an enclave and extend the command line w/o the extra step.

launch_ros/launch_ros/actions/node.py
gbiggs: I've added public API to expose the package and executable, and removed the `node_info` structure in 2d9f3d1.

launch_ros/launch_ros/actions/node.py
gbiggs: It's not just fetching an enclave, it's doing stuff that has side effects. In the case of the security extension, it's actually creating and using files on disc. There's also a need to add an option to the `ros2 launch` command line for the user to use, which is why there is the separate `command_extension` method. The function documentation for that was wrong; I've fixed it in c2b2b7f.

launch_ros/launch_ros/actions/node.py
gbiggs: I've made the two functions free in 0d9cb67.

gbiggs: > Until we get around to replace it, it is permanent

Fair point! :)

launch_ros/launch_ros/actions/node.py
hidmic: Hmm, I'm a bit confused. From what I can see in this patch, `NodeActionExtension.command_extension` is not used by `ros2 launch` directly, but within `Node._perform_substitutions` to extend `Node.cmd`. In which case, my point is that both `NodeActionExtension.command` and `NodeActionExtension.pre_execute` happen on `Action.execute`. There's no configuration stage in between. An indirection through context locals would make sense if there was a context global to override (or not).

Anyhow, I won't block on this. But we do need that documentation sorted out.

hidmic (approved):

launch_ros/launch_ros/actions/node.py
gbiggs: OK, I was the one confused. Sorry about that. I was getting it mixed up with the `ros2 launch` command line flag adding, which is handled by the options extension.

I looked over the code and I agree that it is possible to combine the two functions. I can make an argument against combining them that the current way follows the structure of the `Node` action in that it partitions preparing the command line from preparing ROS-specific arguments and executing the command, but it's a weak argument. So I went ahead and changed it to match your suggestion in da09c67.

hidmic (approved):
CI up to `launch_ros`, `test_launch_ros`, and `ros2launch`:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14135/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8941/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11822/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14234/)

hidmic:
@gbiggs `flake8` isn't happy, mind to check?

gbiggs:
@hidmic I fixed the two flake8 errors I could see in the output.

clalancette:
I fixed a few more flake8 fixes that I saw locally (and in CI).

However, I am still seeing this warning:

```
launch_ros/actions/test_extension.py:18
  /home/ubuntu/ros2_ws/src/ros2/launch_ros/launch_ros/launch_ros/actions/test_extension.py:18: PytestCollectionWarning: cannot collect test class 'TestNodeActionExtension' because it has a __init__ constructor (from: launch_ros/actions/test_extension.py)
    class TestNodeActionExtension(NodeActionExtension):
```

I think what is happening here is that pytest is getting confused because the class starts with the name `Test`.  Looking around, I don't see this class or file being used anywhere.  Is it?

Arnatious:
> I think what is happening here is that pytest is getting confused because the class starts with the name `Test`. Looking around, I don't see this class or file being used anywhere. Is it?

For reference setting the class atrtibute `__test__ = False` will squelch that error

gbiggs:
Sorry about that. That's a class I was using to do some quick tests. I thought I had deleted it, turns out I had committed it instead by mistake. I've pushed the removal of them.

clalancette:
All right, tests are green locally.  Let's run them in CI again and see what happens:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14233/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9031/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11924/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14337/)

clalancette:
All right, this is approved by @hidmic, and CI is clean (other than the Windows failure, but that is unrelated to this PR).  I'm going to go ahead and merge this, thanks for all of the work @gbiggs , and thanks for the reviews @hidmic.

hidmic:
@clalancette shall we run a combined CI with https://github.com/osrf/ros2launch_security/pull/1?

clalancette:
Good idea, I'll go ahead and do that over on that pull request.

PR #182, opened 2020-09-30
Handle any substitution types for SetParameter name argument

jacobperron:
This fixes an error when trying to use the SetParameter action in a front-end launch file (04a8a78).
The parsed 'name' attribute is returned a list of substitutions, which cannot be used as a key in a dictionary.

This PR also fixes some type annotations (abff193).

PR #218, opened 2021-02-08
backport PR: [Foxy] Handle any substitution types for SetParameter name argument (#182)

mjeronimo:
* Handle any substitution types for SetParameter name argument

This fixes an error when trying to use the SetParameter action in a front-end launch file.
The parsed 'name' attribute is returned a list of substitutions, which cannot be used as a key in a dictionary.

Signed-off-by: Jacob Perron <jacob@openrobotics.org>

* Fix SetParameter type annotations

Signed-off-by: Jacob Perron <jacob@openrobotics.org>

### DIFF
--- a/launch_ros/launch_ros/actions/set_parameter.py
+++ b/launch_ros/launch_ros/actions/set_parameter.py
@@ -14,16 +14,16 @@
 
 """Module for the `SetParameter` action."""
 
-from typing import List
-
 from launch import Action
-from launch import Substitution
 from launch.frontend import Entity
 from launch.frontend import expose_action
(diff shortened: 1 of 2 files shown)


jacobperron (approved):

jacobperron:
I may have approved prematurely. Looks like a test is failing. It might be due to API differences in Foxy; I didn't look closely.

mjeronimo:
Sure, I'll investigate.

mjeronimo:
@jacobperron I removed a test case that is invalid for Foxy.

jacobperron (approved):
Thanks! LGTM

PR #222, opened 2021-03-26
Fix namespace with LifecycleNode

### LINKED ISSUE #220: push-ros-namespace is not working for LifecycleNode if namespace is empty
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 20.04
- Installation type:
  - binaries
- Version or commit hash:
  - 0.11.1
- DDS implementation:
  - default

#### Steps to reproduce issue

```xml
// test.launch.xml
<launch>
  <group>
    <push-ros-namespace namespace="test"/>
    <include file="$(find-pkg-share lifecycle)/launch/lifecycle_demo.launch.py" >
    </include>
  </group>
</launch>
```

```sh
$ ros2 launch ./test.launch.xml
$ ros2 node list
/launch_ros_2173028
/lc_talker
/test/lc_client
/test/lc_listener
```

#### Expected behavior

Lifecycle nodes will be also namespaced.
In this case, `/lc_talker` should be `/test/lc_talker`.

#### Actual behavior

No namespace is given to the lifecycle node.

#### Additional information

This is because [if namespace is `''` or `None`, it's set to `/`](https://github.com/ros2/launch_ros/blob/4e74f5a60e023b8d810f17df463692a584cd8575/launch_ros/launch_ros/actions/lifecycle_node.py#L93-L94).
[lifecycle_demo.launch.py](https://github.com/ros2/demos/blob/b0c1389ae61172788cccf4eb4e34b3b84c408a9d/lifecycle/launch/lifecycle_demo.launch.py#L22-L23) uses `None` for namespace.

This behavior seems to be introduced in https://github.com/ros2/launch_ros/pull/179.

I found a workaround, using `TextSubstitution` can avoid this problem.

```py
// lifecycle_demo_workaround.launch.py
from launch.substitutions import TextSubstitution
from launch_ros.actions import LifecycleNode
from launch_ros.actions import Node

    return LaunchDescription([
        LifecycleNode(package='lifecycle', executable='lifecycle_talker',
                      name='lc_talker', namespace=TextSubstitution(text=''), output='screen'),
        Node(package='lifecycle', executable='lifecycle_listener', output='screen'),
        Node(package='lifecycle', executable='lifecycle_service_client', output='screen')
    ])
```

```xml
// workaround.launch.xml
<launch>
  <group>
    <push-ros-namespace namespace="test"/>
    <include file="./lifecycle_demo_workaround.launch.py" >
    </include>
  </group>
</launch>
```

```sh
$ ros2 launch ./workaround.launch.xml
$ ros2 node list
/launch_ros_2173789
/test/lc_client
/test/lc_listener
/test/lc_talker
```

jacobperron:
Fixes #220.
Regression since #179.

The expectation prior to #179 was that providing no namespace argument
to LifecycleNode would result in an empty, relative namespace.
This change fixes the code to meet that expectation.

### DIFF
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -91,7 +91,7 @@ class LifecycleNode(Node):
         if name is None:
             raise RuntimeError("'name' must not be None.'")
         if not namespace:
-            namespace = '/'
+            namespace = ''
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -337,7 +337,7 @@ class Node(ExecuteProcess):
                 validate_node_name(self.__expanded_node_name)
             self.__expanded_node_name.lstrip('/')
             expanded_node_namespace: Optional[Text] = None
-            if self.__node_namespace:
+            if self.__node_namespace is not None:
                 expanded_node_namespace = perform_substitutions(
                     context, normalize_to_list_of_substitutions(self.__node_namespace))
             base_ns = context.launch_configurations.get('ros_namespace', None)


jacobperron:
I'm hoping this doesn't break something else, but AFAICT it fixes the reported issue and all of the existing tests pass.

launch_ros/launch_ros/actions/node.py
```diff
@@ -337,7 +337,7 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
                 validate_node_name(self.__expanded_node_name)
             self.__expanded_node_name.lstrip('/')
             expanded_node_namespace: Optional[Text] = None
-            if self.__node_namespace:
+            if self.__node_namespace is not None:
```
ivanpauno, line 340: doesn't [this comment](https://github.com/ros2/launch_ros/pull/179#discussion_r492017050) still apply?

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -91,7 +91,7 @@ def __init__(
         if name is None:
             raise RuntimeError("'name' must not be None.'")
         if not namespace:
-            namespace = '/'
+            namespace = ''
```
ivanpauno, line 94: maybe an empty `TextSubstitution()` will do the trick (?)

launch_ros/launch_ros/actions/node.py
jacobperron: IIUC, that comment is concerned about the case when the node sets a namespace in code. I tried doing this and it behaves the same before and after the change in this PR: if a namespace is given in the launch file then the node uses that namespace, otherwise it will use the namespace set in the node.

So, I guess not?

jacobperron: err, I guess the comment does still apply. I see differences if I pass an explicitly empty namespace in launch. The behavior was broken in #179. Prior to #179, `namespace=""` uses the namespace set in code and now it uses `/`...

jacobperron: However, this is the same behavior on Rolling. Assuming this is the desired behavior, and we already released this behavior into Foxy, then I'm inclined to leave it as-is. What do you think?

ivanpauno: I agree that having the same behavior that `Rolling` is nice, but it might break some previous launch files so IDK what's best.

ivanpauno (approved):

launch_ros/launch_ros/actions/node.py
jacobperron: If it's going to break launch files in Foxy, then it already has since #179 was released a while ago. I think this PR isn't breaking an existing launch file unless they were depending on the bug described in #220.

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14138/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/8944/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/11825/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14237/)

PR #227, opened 2021-04-07
Use underscores in setup.cfg instead of dashes.

mikepurvis:
Silences this build time warning:

```
lib/python3.8/site-packages/setuptools/dist.py:642: UserWarning: Usage of dash-separated 'script-dir' will not be supported in future versions. Please use the underscore name 'script_dir' instead
```

### DIFF
--- a/launch_testing_ros/setup.cfg
+++ b/launch_testing_ros/setup.cfg
@@ -3,6 +3,6 @@
 source = .
 omit = setup.py
 [develop]
-script-dir=$base/lib/launch_testing_ros
+script_dir=$base/lib/launch_testing_ros
 [install]
-install-scripts=$base/lib/launch_testing_ros
+install_scripts=$base/lib/launch_testing_ros


clalancette:
Unfortunately, we can't do this at this time.

When building binaries for Ubuntu 20.04, we use the system version of setuptools, which is 45.2.0.  That version does *not* support the underscores at all.  So we can't make this change and still build binaries.

What I'm going to do for now is to make this a draft PR, and add it to the H-Turtle board.  When we switch to Ubuntu 22.04 (late this year), we'll be able to merge this.

mikepurvis:
Ah fun, thanks for explaining the reasoning.

clalancette:
It turns out I was wrong about this not working in Ubuntu Focal.  Thus, we can merge this.  I'm going to add this back to the Galactic board and turn it back into a real PR.

clalancette (approved):

ivanpauno (approved):
LGTM!

clalancette:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14418/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9191/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12092/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14517/)

PR #238, opened 2021-05-18
[Dashing backport]: Fix race with launch context changes when loading composable nodes (#…

clalancette:
…166)

* Fix race with launch context changes when loading composable nodes

This bug was discovered when trying load composable nodes from a GroupAction.
The ROS namespace (and presumably other remaps) pushed onto the context stack
was popped after the LoadComposableNodes execute() function finished.
But because the loading happens asynchronously, we need to make sure we get the
necessary information from the context before execute() finishes.

Signed-off-by: Jacob Perron <jacob@openrobotics.org>

* Add regression tests for LoadComposableNode

Signed-off-by: Jacob Perron <jacob@openrobotics.org>

* Properly shutdown mock conatiner node

Also added some debug logs to the load node action for posterity.

Backporting note: As far as I can tell, the *Composable* launch_ros classes in
Dashing are not setup to be able to be tested.  We would need to add a bunch
of additional dependencies to write tests here.  I didn't think
that was worth it for Dashing, so I just removed the tests.

Signed-off-by: Jacob Perron <jacob@openrobotics.org>
Signed-off-by: Chris Lalancette <clalancette@openrobotics.org>

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -68,13 +68,13 @@ class LoadComposableNodes(Action):
 
     def _load_node(
         self,
-        composable_node_description: ComposableNode,
+        request: composition_interfaces.srv.LoadNode.Request,
         context: LaunchContext
     ) -> None:
         """
         Load node synchronously.
 
-        :param composable_node_description: description of composable node to be loaded
+        :param request: service request to load a node
         :param context: current launch context
         """
         while not self.__rclpy_load_node_client.wait_for_service(timeout_sec=1.0):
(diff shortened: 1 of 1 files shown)


jacobperron (approved):

hidmic (approved):
LGTM pending green Dashing CI

clalancette:
CI (`--packages-above launch_ros`):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14666/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9418/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12329/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14809/)

clalancette:
Looking at the CI output, the flake8/pep257 errors are pre-existing, and not because of this PR.  But I'd like confirmation from @nuclearsandwich before merging.

nuclearsandwich:
> Looking at the CI output, the flake8/pep257 errors are pre-existing, and not because of this PR. But I'd like confirmation from @nuclearsandwich before merging.

Yeah, those linters weren't labeled correctly in Dashing so they show even with linters disabled.

nuclearsandwich (approved):

PR #244, opened 2021-06-10
ROS Timer Action

### LINKED ISSUE #239: ROS Timer Action
#### Feature description

Similar to the [TimerAction](https://github.com/ros2/launch/blob/3a5c9df2e384271637ac029f48f020b7916498d6/launch/launch/actions/timer_action.py), it would be great to have a timer action that was aware of ROS time (ie. rather than wall clock, it could use ROS time as defined [here](https://design.ros2.org/articles/clock_and_time.html)).

I imagine usage would look almost identical to `TimerAction`, e.g.

```python
RosTimerAction(
    period=5.0,
    actions=[
        # some list of actions to execute every 5 seconds ...
    ],
    cancel_on_shutdown=True,
)
```

If we ensure the ROS node in the launch context is using ROS time, then this may be as easy as registering a ROS timer.

rebecca-butler:
Closes #239


test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+from builtin_interfaces.msg import Time
+
+
+def test_multiple_launch_with_timers():
+    def generate_launch_description():
+        return launch.LaunchDescription([
+
+            launch.actions.ExecuteProcess(
+                cmd=[sys.executable, '-c', 'while True: pass'],
+            ),
```
jacobperron, line 40: minor:
It's not immediately clear why we need this; I suppose this was copied over from `test_timer_action.py`. I'm pretty sure it's here to keep the launch process alive. Since we create the same action throughout the tests, I suggest refactoring into a semantically meaningful class or function in this file (e.g. "KeepAlive"). Note, we have a similar utility in `launch_testing` (though I wouldn't depend on it here): https://github.com/ros2/launch/blob/3a5c9df2e384271637ac029f48f020b7916498d6/launch_testing/launch_testing/util/__init__.py#L24

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+    ])
+
+    ls = launch.LaunchService()
+    ls.include_launch_description(ld)
+    assert 0 == ls.run()
+    assert len(shutdown_reasons) == 1
+    assert shutdown_reasons[0].reason == 'fast shutdown'
+
+
+def test_time_is_passing():
```
jacobperron, line 128: nit: I think this test could be combined with earlier "sanity check". I'd keep the timer at 1 second to keep the duration of running tests shorter.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+        ),
+
+        _shutdown_listener_factory(shutdown_reasons),
+    ])
+
+    start_time = time.time()
+    ls = launch.LaunchService()
+    ls.include_launch_description(ld)
+    assert 0 == ls.run()
+    assert abs(time.time() - start_time) > 5
```
jacobperron, line 154: Because of system time jitter, we might want to relax this assertion to avoid a flaky test. E.g.

```suggestion
    assert abs(time.time() - start_time) > (5 - tolerance)
```
where `tolerance` is 0.1 or something.

We could also check that timer didn't take too long either:

```python
    assert abs(time.time() - start_time) < (5 + tolerance)
```

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+    assert shutdown_reasons[0].reason == 'Five second timeout'
+
+
+def test_timer_uses_sim_time():
+    """Test that timer uses time from /clock topic."""
+    # Create clock publisher node
+    rclpy.init()
+    node = rclpy.create_node('clock_publisher_node')
```
jacobperron, line 162: nit: I would prefix the test node so it's clear where this came from, in case there's an unclean shutdown and the process lingers
```suggestion
    node = rclpy.create_node('test_ros_timer_action_node')
```

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+import launch
+import launch.event_handlers
+
+from launch_ros.actions import RosTimerAction
+
+import rclpy
+from rclpy.clock import Clock, ClockType
+
+from rosgraph_msgs.msg import Clock as ClockMsg
+from builtin_interfaces.msg import Time
```
jacobperron, line 31: We should declare test dependencies to these packages in the [package.xml](https://github.com/ros2/launch_ros/blob/master/test_launch_ros/package.xml).

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+
+    start_time = time.time()
+
+    ls = launch.LaunchService()
+    ls.include_launch_description(ld)
+    assert 0 == ls.run()
+
+    # Simulated time is 100x faster, so 200 sec timer should finish in 2 sec
+    # (Some extra time provided)
+    assert (time.time() - start_time < 6)
```
jacobperron, line 201: 6 second seems like a very large tolerance, is that necessary?

jacobperron (commented):
Good stuff! Thanks for adding tests :fire:

launch_ros/launch_ros/actions/ros_timer_action.py
```diff
@@ -0,0 +1,220 @@
...
+from launch.some_substitutions_type import SomeSubstitutionsType_types_tuple
+from launch.utilities import create_future
+from launch.utilities import ensure_argument_type
+from launch.utilities import is_a_subclass
+from launch.utilities import type_utils
+
+from rclpy.parameter import Parameter
+
+
+@expose_action('ros_timer_action')
```
ivanpauno, line 54:
```suggestion
@expose_action('ros_timer')
```

We typically don't add an `_action` in those names

launch_ros/launch_ros/actions/ros_timer_action.py
```diff
@@ -0,0 +1,220 @@
...
+class RosTimerAction(Action):
```
ivanpauno, line 55: is it possible to share more code with `TimerAction`?
I haven't looked too closely, but maybe you can inherit from it and override `__wait_to_fire_event()/parse()`.
`execute()` and most of the other methods seem to not have changed.

Note: the fact that `__wait_to_fire_event()` starts with a double underscores makes it harder to be overriden, but it's still possible (see https://stackoverflow.com/questions/6444576/python-method-name-with-double-underscore-is-overridden). We could also rename the method to `_wait_to_fire_event()` in `launch` to make things easier.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
rebecca-butler: I've removed this as it doesn't seem to be necessary for the tests.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
rebecca-butler: I've combined them.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
rebecca-butler: I added these as well as a dependency on `rclpy`.

launch_ros/launch_ros/actions/ros_timer_action.py
```diff
@@ -0,0 +1,220 @@
...
+    async def __wait_to_fire_event(self, context):
+        node = get_ros_node(context)
+
+        if type_utils.perform_typed_substitution(context, self.__use_sim_time, bool):
+            param = Parameter(
+                'use_sim_time',
+                Parameter.Type.BOOL,
+                True
+            )
+            node.set_parameters([param])
```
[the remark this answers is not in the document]
rebecca-butler: Your reasoning makes sense. I've added a SetUseSimTime action.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
rebecca-butler: I've added a 0.1 second tolerance to both tests.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
rebecca-butler: I changed these tests so that the start and end times are now recorded using the `OpaqueFunction` action while the launch service is running. Measuring the timeout is now much more accurate, so the tolerance can be smaller.

launch_ros/launch_ros/actions/ros_timer_action.py
```diff
@@ -0,0 +1,220 @@
...
+        **kwargs
+    ) -> None:
+        """Create a RosTimerAction."""
+        super().__init__(**kwargs)
+        period_types = list(SomeSubstitutionsType_types_tuple) + [float]
+        ensure_argument_type(period, period_types, 'period', 'RosTimerAction')
+        ensure_argument_type(actions, collections.abc.Iterable, 'actions', 'RosTimerAction')
+        self.__period = type_utils.normalize_typed_substitution(period, float)
+        self.__actions = actions
+        self.__context_locals = {}  # type: Dict[Text, Any]
```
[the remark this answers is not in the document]
rebecca-butler: Fixed.

launch_ros/launch_ros/actions/ros_timer_action.py
rebecca-butler: I've changed the name from RosTimerAction to RosTimer.

test_launch_ros/test/test_launch_ros/actions/test_ros_timer_action.py
```diff
@@ -0,0 +1,201 @@
...
+        ),
+    ])
+
+    start_time = time.time()
+
+    ls = launch.LaunchService()
+    ls.include_launch_description(ld)
+    assert 0 == ls.run()
+
+    # Simulated time is 100x faster, so 200 sec timer should finish in 2 sec
```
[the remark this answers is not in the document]
rebecca-butler: It doesn't actually take 200 seconds to run. This test contains a clock publisher node that increases the sim time by 100 seconds for every second of system time. So the timer is running for 200 simulated seconds, which is only 2 seconds in wall clock time.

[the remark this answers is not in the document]
ivanpauno: ah, I read the exact opposite.
That makes sense to me.

launch_ros/launch_ros/actions/ros_timer_action.py
```diff
@@ -0,0 +1,220 @@
...
+            self.__logger.debug(
+                'timer {} not waiting because it was canceled before being executed'.format(self),
+            )
+            self.__completed_future.set_result(None)
+            return None
+
+        # Once per context, install the general purpose OnTimerEvent event handler.
+        if not hasattr(context, '_RosTimerAction__event_handler_has_been_installed'):
+            context.register_event_handler(EventHandler(
+                matcher=lambda event: is_a_subclass(event, TimerEvent),
```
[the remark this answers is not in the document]
jacobperron: We discussed this offline. I misunderstood the logic originally. This code block is only executed once per timer action, registering a "global" (in a sense) event handler. It's a bit confusing, but my concern is nothing to worry about.

jacobperron (commented):
LGTM. I've left some minor comments.

I agree with Ivan that it would be nice if we could inherit from `launch.actions.TimerAction` to cut down on redundant code. I think it's worth trying, though I won't block this PR if we want to tackle it in a follow-up.

rebecca-butler:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14743/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9493/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12407/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14890/)

test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+def test_multiple_launch_with_timers():
+    def generate_launch_description():
+        return launch.LaunchDescription([
+
+            RosTimer(
+                period=1.,
+                actions=[
+                    launch.actions.Shutdown(reason='Timer expired')
```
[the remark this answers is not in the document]
rebecca-butler: I've removed this shutdown as it wasn't needed.

launch_ros/launch_ros/actions/set_use_sim_time.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def execute(self, context: LaunchContext):
+        """Execute the action."""
+        node = get_ros_node(context)
+        param = Parameter(
+            'use_sim_time',
+            Parameter.Type.BOOL,
+            self.value
+        )
+        node.set_parameters([param])
```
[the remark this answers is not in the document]
rebecca-butler: I've added a check that raises a RuntimeError.

rebecca-butler:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14746/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9496/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12410/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14893/)

launch_ros/launch_ros/actions/set_use_sim_time.py
```diff
@@ -0,0 +1,62 @@
...
+from launch.frontend import expose_action
+from launch.frontend import Parser
+from launch.launch_context import LaunchContext
+
+from launch_ros.ros_adapters import get_ros_node
+
+from rclpy.parameter import Parameter
+
+
+@expose_action('set_use_sim_tim')
```
wep21, line 27: typo?(tim->time)

rebecca-butler: Thanks for catching that!

jacobperron (approved):

launch_ros/launch_ros/actions/ros_timer_action.py
rebecca-butler: I'll address this in a follow-up PR.

jacobperron:
For posterity, there was a bug where we weren't handling substitutions properly. Fixed in https://github.com/ros2/launch_ros/pull/264

PR #249, opened 2021-06-24
Add ros_arguments option to Node action

### LINKED ISSUE #60: Add `ros_arguments` to `Node` action constructor
#### Feature description
See https://github.com/ros2/rmw_fastrtps/issues/265#issuecomment-526181148.

christophebedard:
Closes #60

This adds a `ros_arguments` option to the `Node` action. It's simply: `ros_arguments == ['--ros-args'] + arguments`.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -125,6 +125,7 @@ class Node(ExecuteProcess):
         exec_name: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
+        ros_arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         **kwargs
     ) -> None:
@@ -180,6 +181,9 @@ class Node(ExecuteProcess):
         wildcard namespace (`/**`) and other specific parameter declarations
         may overwrite it.
 
+        Using `ros_arguments` is equivalent to using `arguments` with a
+        prepended '--ros-args' item.
+
         :param: executable the name of the executable to find if a package
             is provided or otherwise a path to the executable to run.
         :param: package the package in which the node executable can be found
@@ -191,6 +195,7 @@ class Node(ExecuteProcess):
             or dictionaries of parameters.
         :param: remappings ordered list of 'to' and 'from' string pairs to be
             passed to the node as ROS remapping rules
+        :param: ros_arguments list of ROS arguments for the node
         :param: arguments list of extra arguments for the node
         """
         if package is not None:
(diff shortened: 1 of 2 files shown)


launch_ros/launch_ros/actions/node.py
```diff
@@ -180,6 +181,9 @@ def __init__(
         wildcard namespace (`/**`) and other specific parameter declarations
         may overwrite it.
 
+        Using `ros_arguments` is equivalent to using `arguments` with a
+        prepended '--ros-args' item.
```
hidmic, line 185: @christophebedard I see the convenience BUT is it necessary? Most ROS arguments can be given in an idiomatic way  (e.g. via `remappings`) already. If we need `--log-level`, I'd argue we should add it. CC @wjwwood @ivanpauno.

christophebedard: I found #60 while answering an answers.ros.org question: https://answers.ros.org/question/380883/specify-log-level-on-ros2-node-in-launch-file/#380965. In this case, it seemed to confuse the person. They weren't sure whether they needed to use `--` beause `launch_ros` appends `--ros-args ...` too (even if in practice `rcl` handles it just fine). I think having `ros_arguments` would make it much simpler, but yeah it's up to you the maintainers :stuck_out_tongue_closed_eyes:

> If we need `--log-level`

you mean like a `LogLevel` substitution?

hidmic: > I found #60 while answering an answers.ros.org question: https://answers.ros.org/question/380883/specify-log-level-on-ros2-node-in-launch-file/#380965. In this case, it seemed to confuse the person.

I see. I was wondering about the use case. If not having to write out `--ros-args` helps people, I'm onboard.

> you mean like a LogLevel substitution?

Like a `log_level` argument that can be a substitution. But that's a separate PR.

wjwwood: Seems like we can handle things better if the user explicitly tells us when something is a ros argument, so +1 from me.

christophebedard:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14817/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9567/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12483/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14972/)

hidmic (approved):
LGTM

christophebedard:
@hidmic I can't merge PRs so I'll leave that to you!

hidmic:
Going in ! Thanks for the patience @christophebedard.

christophebedard:
> Argh, @christophebedard one thing we missed is adding support for this attribute in the `Node.parse` function. Would you mind?

oh yeah I really didn't think about that! I'll create another PR as a follow-up.

PR #250, opened 2021-06-24
Add `push_ros_namespace` alias to `push-ros-namespace`

### LINKED ISSUE #142: Add `push_ros_namespace` alias to `push-ros-namespace`
## Bug report

`push-ros-namespace` is the current decorator used for pushing a ROS namespace within a Launch GroupAction.  However, we think it is more aesthetically nice to use `push_ros_namespace`, so we should add that decorator as well.  See https://github.com/ros2/ros2_documentation/pull/539/files#r388486312 for more details.

christophebedard:
Closes #142

Once this is approved, I will update the documentation to use the alias with the underscores (https://github.com/ros2/ros2_documentation/pull/539/files#r409568794).

### DIFF
--- a/launch_ros/launch_ros/actions/push_ros_namespace.py
+++ b/launch_ros/launch_ros/actions/push_ros_namespace.py
@@ -33,6 +33,7 @@ from launch_ros.utilities import prefix_namespace
 from rclpy.validate_namespace import validate_namespace
 
 
+@expose_action('push_ros_namespace')
 @expose_action('push-ros-namespace')
 class PushRosNamespace(Action):
     """
--- a/test_launch_ros/test/test_launch_ros/frontend/test_push_ros_namespace_frontend.py
+++ b/test_launch_ros/test/test_launch_ros/frontend/test_push_ros_namespace_frontend.py
@@ -31,6 +31,16 @@ def test_launch_namespace_yaml():
     with io.StringIO(yaml_file) as f:
         check_launch_namespace(f)
 
+    yaml_file_alias = textwrap.dedent(
+        r"""
+        launch:
+           - push_ros_namespace:
+               namespace: 'asd'
+        """
+    )
+    with io.StringIO(yaml_file_alias) as f:
+        check_launch_namespace(f)
+
 
 def test_launch_namespace_xml():
     xml_file = textwrap.dedent(
@@ -41,6 +51,16 @@ def test_launch_namespace_xml():
         """
     )
     with io.StringIO(xml_file) as f:
+        check_launch_namespace(f)
+
+    xml_file_alias = textwrap.dedent(
+        r"""
+        <launch>
+            <push_ros_namespace namespace="asd"/>
+        </launch>
+        """
+    )
+    with io.StringIO(xml_file_alias) as f:
         check_launch_namespace(f)
 
 


christophebedard:
> Once this is approved, I will update the documentation to use the alias with the underscores ([ros2/ros2_documentation#539 (files)](https://github.com/ros2/ros2_documentation/pull/539/files#r409568794)).

It was pretty straightforward, so I just did it: https://github.com/ros2/ros2_documentation/pull/1703

launch_ros/launch_ros/actions/push_ros_namespace.py
```diff
@@ -33,6 +33,7 @@
 from rclpy.validate_namespace import validate_namespace
 
 
+@expose_action('push_ros_namespace')
 @expose_action('push-ros-namespace')
```
hidmic, line 37: @christophebedard meta: I wonder if it'd make sense to change `expose_action` to automatically register "underscore" and "dash" aliases. CC @ivanpauno @wjwwood.

wjwwood: Even if we did this (just this pr or the automatic mapping) I really think we should pick one way and stick with it. 

Ideally there wouldn't be another way to do it. 

I'm slightly 👎 to changing the documentation to use underscores after merging this, if we use dashes everywhere else.

christophebedard: > I'm slightly  to changing the documentation to use underscores after merging this, if we use dashes everywhere else.

underscores are used everywhere else. This one action is the only one that uses dashes (https://github.com/ros2/ros2_documentation/pull/539/files#r388486312)

wjwwood: In that case I'd say 👍 for this change and doc change but less so for the automatic mapping. Again preferring one good way to do things. 

I'd even be in favor of deprecating the use of dashes. 

But I could be convinced otherwise on each of those points I think.

christophebedard: picking one and sticking with it does make sense, as does deprecating the current name with the dashes (although I'm not sure where in `launch_ros` I'd put a warning message for it).

hidmic: > Even if we did this (just this pr or the automatic mapping) I really think we should pick one way and stick with it.

That's fair. IIRC aliases had this purely cosmetic purpose back when we first thought about this problem. Specifically, underscores are somewhat less aesthetically pleasing in XML than dashes. Accepting both does not hurt format translation. But we can be strict too. 

We do not enforce one tag per entity though. Maybe we should.

> picking one and sticking with it does make sense, as does deprecating the current name with the dashes

There's no support for it. If we want to be strict, we can add a list of deprecated aliases to [`expose_*`](https://github.com/ros2/launch/blob/master/launch/launch/frontend/expose.py) decorators. We can register the same parsing function for those aliases but decorated to emit a deprecation warning.

wjwwood: Sounds like deprecating isn't worth it. We can just change the documentation, which is probably enough. 

I actually don't mind if there is the ability to use dash or underscore automatically that much, but I do think we should be consistent in the documentation and avoid dashes some place and underscores in others. 

Either way sounds like this pr is good and the doc changes are a good idea too.

hidmic: > Sounds like deprecating isn't worth it. We can just change the documentation, which is probably enough.

Yeah, I'm tempted to just wipe it. Specially if it's just one. But I will say that, eventually, we'll need tick-tock deprecation cycles here too.

wjwwood: Oh maybe I wasn't clear when I said deprecation wasn't worth it. I was thinking we'd just leave it but document that you should use underscores. So never remove it.

hidmic: Ooh. Yeah. I misunderstood you. I'm slightly inclined towards enforcing rather than documenting. None of this documentation is available online just yet. Until then, most people won't read it.

It doesn't have to happen in this PR though.

hidmic (approved):
LGTM pending green CI.

We should revisit frontend deprecation and one-tag-per-entity policy at some point in time.

christophebedard:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14823/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9573/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12489/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14978/)

hidmic:
Alright, this one's ready to go.

PR #251, opened 2021-06-24
Add `.launch` as an extension for launch files auto-completion

### LINKED ISSUE #149: No tab autocomplete or suggesting for xml launch files
## Bug report

**Required Info:**

- Operating System:
  - 
- Installation type:
  -  binaries 
- Version or commit hash:
  - 0.9.6-1bionic.20200121.220323 for launch and 0.9.6-1bionic.20200121.220435 for launch xml
- DDS implementation:
  - Fast-RTPS

#### Steps to reproduce issue
Install binaries from apt and create an xml launch file in a custom package. Build and source package then perform ros2 launch package_name tab-tab. 

#### Expected behavior
The autocomplete list should show any xml launch files available in that package.

#### Actual behavior
Only .py launch files are shown. If the name of the xml launch file is input into the ros2 launch command and run it launches fine but the xml does not show up when trying to perform tab-completion.

Not sure if this is a bug in existing code or is just not implemented yet but it sure is annoying when you have a lot of xml launch files in a package. I have not tested this issue with yaml launch files.

christophebedard:
Closes #149

This makes `ros2 launch pkg <tab>` auto-completion include files ending with `.launch` as well.

I looked into adding a test for this, but there are currently none for auto-completion (that I could find). Let me know if I should add one.

### DIFF
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -85,6 +85,7 @@ is_launch_file.extensions = [
     'launch.' + extension for extension in Parser.get_available_extensions()
 ]
 is_launch_file.extensions.append('launch.py')
+is_launch_file.extensions.append('.launch')
 is_launch_file.extensions = tuple(is_launch_file.extensions)
 
 


christophebedard:
Would you prefer this instead?

> Perhaps each parser could provide a list of well-known file extensions.

https://github.com/ros2/launch_ros/issues/149#issuecomment-713670587

christophebedard:
This is a semi-relevant thought, but I think I remember discussions about launch file extensions and whether or not we should actually allow/support/encourage the `.launch` extensions and other extensions that don't end with the actual file type, like `.yaml` and `.py`. If each parser were to provide a list of well-known file extensions, and if each of them included `.launch` in their lists, then we would explicitly be supporting/encouraging this.

Of course, if parsers were to provide a list of file extensions, we could still just append `.launch` to the list like I did here instead of having the parsers include it in their lists. They are not mutually exclusive.

christophebedard:
> Of course, if parsers were to provide a list of file extensions, we could still just append `.launch` to the list like I did here instead of having the parsers include it in their lists. They are not mutually exclusive.

I opened https://github.com/ros2/launch/pull/516 and https://github.com/ros2/launch_ros/pull/252. This PR could go alongside those, or we could move `.launch` to the set of one or both of the parsers, depending on what we want.

hidmic:
IIUC #252 supersedes #251. Is that right?

christophebedard:
> IIUC #252 supersedes #251. Is that right?

No because #252 doesn't add `.launch` to the set of extensions. It only updates `launch_ros` to use https://github.com/ros2/launch/pull/516

christophebedard:
> No because #252 doesn't add `.launch` to the set of extensions. It only updates `launch_ros` to use [ros2/launch#516](https://github.com/ros2/launch/pull/516)

But I do need to modify this one-line PR for https://github.com/ros2/launch/pull/516 since `.append('.launch')` would become `.add('.launch')`

christophebedard:
Now that I think about it, though, you're right. This wouldn't really work with #252 and https://github.com/ros2/launch/pull/516 if we make more complex changes (like https://github.com/ros2/launch/pull/516#issuecomment-869779401). `.launch` would need to be in the file extensions sets of both `launch_xml` and `launch_yaml` so that it is associated to specific parser implementations.

PR #253, opened 2021-06-29
Properly support ros_args attribute through launch frontends

christophebedard:
Follow-up to #249

This actually adds support for a `ros_args` attribute instead of only supporting `ros_arguments` through the `Node` action constructor.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -310,6 +310,9 @@ class Node(ExecuteProcess):
         args = entity.get_attr('args', optional=True)
         if args is not None:
             kwargs['arguments'] = super()._parse_cmdline(args, parser)
+        ros_args = entity.get_attr('ros_args', optional=True)
+        if ros_args is not None:
+            kwargs['ros_arguments'] = super()._parse_cmdline(ros_args, parser)
         node_name = entity.get_attr('node-name', optional=True)
         if node_name is not None:
             kwargs['node_name'] = parser.parse_substitution(node_name)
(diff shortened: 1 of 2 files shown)


christophebedard:
I assume I should probably document this somewhere, but the launch-related tutorials/guides on docs.ros.org don't really document all the attributes. For example, the `args` attribute isn't used/mentioned anywhere.

christophebedard:
> I assume I should probably document this somewhere

I added it to the launch XML schema (https://github.com/ros2/design/pull/318) and to the list of new features for Humble (https://github.com/ros2/ros2_documentation/pull/1710).

christophebedard:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14832/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9582/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12498/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/14989/)

hidmic (approved):
Looks reasonable to me. @ivanpauno ?

ivanpauno (approved):

andrewjong:
Is there anywhere `ros_args` can be added to the documentation? I had to dig to find this issue to understand that `ros_args` could be used.

christophebedard:
That's a good question. I don't know where this should be documented, especially since in this case the frontends use `ros_args` while the `Node` Python class constructor argument is called `ros_arguments`.

christophebedard:
I created an issue for this: https://github.com/ros2/launch_ros/issues/404

PR #254, opened 2021-07-14
Resolves #37 - Added --launch-prefix argument for 'ros2 launch' command

### LINKED ISSUE #37: [ros2launch] Making `launch-prefix` easily accessible
#### Feature description

The `launch-prefix` option comes handy when debugging/profiling, however it is not 'easily' accessible as of now, I am referring to [this example](https://github.com/ros2/launch/blob/20d5f06395192e4ce47ab162c0de0f9ab74a60d4/launch/examples/launch_counters.py#L71). This scheme offers a fine granularity as which node should benefit what prefix, however sometimes one may simply want to launch an entire launch file under a given prefix.
Given that the `run` verb provides a `--prefix` option, the `launch` verb could similarly provide a `--launch-prefix` option.

One possible implementation boils down to including the desired launch file in a `LaunchDescription` which sets the launch configuration appropriately, something similar to [this script](https://github.com/artivis/ros2_best_practices/blob/master/launch.md#a-launch-prefix-wrapper).

camm73:
Resolves: ros2#37

Makes `launch-prefix` more accessible by adding a CLI argument `--launch-prefix 'LAUNCH PREFIX'` where `'LAUNCH PREFIX'` is the desired command prefix (wrapped in quotes if it contains spaces). This achieves the same behavior as setting the launch file argument `'launch-prefix:=LAUNCH PREFIX'` which applies the prefix to all executable Actions. For simplicity, the `--launch-prefix` argument mirrors the `--prefix` argument provided by `ros2 run`.

I did not find documentation for the `ros2 launch` command specifically, but I can submit a documentation update PR to https://navigation.ros.org/tutorials/docs/get_backtrace.html if that is the best location for it.

### DIFF
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -152,10 +152,17 @@ def launch_a_launch_file(
     """Launch a given launch file (by path) and pass it the given launch file arguments."""
     for name in sorted(option_extensions.keys()):
         option_extensions[name].prestart(args)
+
+    # If 'launch-prefix' launch file argument is also provided in the user input,
+    # the 'launch-prefix' option is applied since the last duplicate argument is used
+    if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
+        launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
+
     launch_service = launch.LaunchService(
         argv=launch_file_arguments,
         noninteractive=noninteractive,
         debug=debug)
+
     parsed_launch_arguments = parse_launch_arguments(launch_file_arguments)
     # Include the user provided launch file using IncludeLaunchDescription so that the
     # location of the current launch file is set.
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -25,6 +25,7 @@ from ament_index_python.packages import PackageNotFoundError
 import launch
 from launch.frontend import Parser
 from launch.launch_description_sources import get_launch_description_from_any_launch_file
+from launch.substitutions import TextSubstitution
 
 
 class MultipleLaunchFilesError(Exception):
@@ -153,15 +154,16 @@ def launch_a_launch_file(
     for name in sorted(option_extensions.keys()):
         option_extensions[name].prestart(args)
 
-    # If 'launch-prefix' launch file argument is also provided in the user input,
-    # the 'launch-prefix' option is applied since the last duplicate argument is used
-    if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
-        launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
+    prefix_sub = None
+    if args.launch_prefix is not None:
+        prefix_sub = TextSubstitution(text=args.launch_prefix)
 
     launch_service = launch.LaunchService(
         argv=launch_file_arguments,
         noninteractive=noninteractive,
-        debug=debug)
+        debug=debug,
+        launch_prefix=prefix_sub
+    )
 
     parsed_launch_arguments = parse_launch_arguments(launch_file_arguments)
     # Include the user provided launch file using IncludeLaunchDescription so that the
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -25,7 +25,6 @@ from ament_index_python.packages import PackageNotFoundError
 import launch
 from launch.frontend import Parser
 from launch.launch_description_sources import get_launch_description_from_any_launch_file
-from launch.substitutions import TextSubstitution
 
 
 class MultipleLaunchFilesError(Exception):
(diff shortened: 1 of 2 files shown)


hidmic (commented):
@camm73 how is this different from `ros2 launch package_name launch_file.py launch-prefix:='...'` ? In other words, how does this manage to inject the prefix into every executable, as described in #37?

camm73:
Maybe I’m misunderstanding the issue, but it sounds like the goal is to make `--launch-prefix 'LAUNCH PREFIX'` syntactic sugar for setting the `'launch-prefix:=LAUNCH PREFIX'` launch configuration so that it is more accessible and mirrors `ros2 run`'s `--prefix` argument.

My understanding—and correct me if I’m wrong—is that by setting the `launch-prefix` launch configuration, it will be prefixed to all executable commands via the `LaunchConfiguration` substitution that takes place [here in ExecuteProcess](https://github.com/ros2/launch/blob/2a97e806ead3b0b6428bd7c1f648c3a01bb9e231/launch/launch/actions/execute_process.py#L684). I understand that this substitution wouldn't take place if `prefix` was explicitly provided for an executable in the launch file; however, my assumption was that the goal is to allow a generic prefix to be applied to all executables without overriding executable-specific prefixes that have been deliberately set in the launch file.

If I'm mistaken in that, however, I can instead directly inject the prefix provided via `--launch-prefix` into each executable command via ExecuteProcess, overriding any existing prefixes specified in the launch file.

hidmic:
> My understanding—and correct me if I’m wrong—is that by setting the launch-prefix launch configuration, it will be prefixed to all executable commands via the LaunchConfiguration substitution that takes place here in ExecuteProcess. 

Ah, that's correct. I missed that one.

> Maybe I’m misunderstanding the issue, but it sounds like the goal is to make --launch-prefix 'LAUNCH PREFIX' syntactic sugar for setting the 'launch-prefix:=LAUNCH PREFIX' launch configuration so that it is more accessible and mirrors ros2 run's --prefix argument.

Hmm, OK. I don't see a benefit right now TBH, but I do know that maybe, in a distant future, the launch argument trick to get to all `ExecuteProcess` actions may not work anymore (if `launch` starts scoping). Then, the flag will come in handy.

camm73:
> I do know that maybe, in a distant future, the launch argument trick to get to all `ExecuteProcess` actions may not work anymore (if `launch` starts scoping). Then, the flag will come in handy.

In that case I'll refactor this PR to pass the `--launch-prefix` flag input to the `LaunchService` instead. Then I can open another PR on the [launch repo](https://github.com/ros2/launch) to handle that argument in `LaunchService` which can then pass it to `LaunchContext`—the same way the `--noninteractive` flag is already handled. Then the prefix will be available in `ExecuteProcess` via the `LaunchContext` and can be directly injected into the command. That should make this more future-proof.

camm73:
I opened up a PR on the launch repo (https://github.com/ros2/launch/pull/521) that adds support to pass the the `--launch-prefix` flag value via `LaunchService` and then via `LaunchContext`. It then uses that value from the Context to directly apply the prefix within `ExecuteProcess`.

hidmic:
Oh. First of all, thanks for being this proactive! Having said that, I'd suggest we stick to using launch configurations. Pushing them through `service.context.launch_configurations` should be easy enough. 

My apologies if I made you think I did not agree with the original approach. Right now `launch` has multiple ways to move data around and I'd rather not add yet another one (i.e. substitution passing through top-level context attributes). IMO all this is in dire need for some standardization.

CC'ing @ivanpauno and @wjwwood, as they might not necessarily agree with me.

camm73:
No problem. I'll just leave it "as-is" until @ivanpauno and @wjwwood get a chance to look at it, and then we can go from there.

camm73:
Just wanted to follow up on this—do you have an estimate of when you'll have time to look over these changes, @ivanpauno @wjwwood?

Also wanted to mention that I added a feature proposal #257 which will be based on these changes. Using them together should help make targeted debugging faster without needing to edit a launch file.

wjwwood:
> the launch argument trick to get to all ExecuteProcess actions may not work anymore (if launch starts scoping).

I don't think that's possible. It would break absolutely everything. But it is true the users may clear the launch configurations when including a launch file in which case it would not affect `ExecuteProcess` actions inside the scope where the configurations were cleared. But that's gonna be an issue anyway, as launch configurations are the only way to affect multiple actions simultaneously. So we'll have to address that if people have issues with it in the future.

For now, I think having the syntactic sugar of `--launch-prefix <foo>` becoming `launch-prefix:=<foo>` is fine. @ivanpauno seemed to express the same in https://github.com/ros2/launch_ros/issues/37#issuecomment-669311399, but maybe he meant something else or misunderstood the approach. So I'll let him comment.

ivanpauno:
Sonds good to me

camm73:
Reverted to the previous implementation, so this PR should be ready for review now.

hidmic (approved):
Alright, let's run CI up to `launch`, `ros2launch`, `launch_ros`, and `test_launch_ros`

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15000/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9739/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12687/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15162/)

Thanks for the patience @camm73.

hidmic:
Alright, all CI is green. Moving forward.

Issue #255, opened 2021-07-14
Declare dependencies on launch_{xml,yaml} parsers to make sure they're available


christophebedard:
#### Feature description

In some cases, if a package uses/runs launch files through `ros2launch` or `launch_testing` and nothing declares a dependency on `launch_xml`/`launch_yaml`, then the XML/YAML parsers won't be available. This causes weird failures and the cause isn't obvious.

Some example packages do declare an `<exec_depend>` or `<test_depend>` on one of the parser packages (e.g. https://github.com/ros2/demos/blob/3af19c93c7abd9a83caca47de0780e941faeb22b/demo_nodes_cpp/package.xml#L26), but in general this is a bit unexpected.

See discussion under https://github.com/ros2/launch/pull/516#issuecomment-879445998

#### Implementation considerations

Simply declare `<exec_depend>`s on `launch_xml` and `launch_yaml` in `ros2launch` and `launch_testing` (note that the latter is not in this repo). This should cover most cases.

We can also declare and use a group dependency (like `launch_frontend_packages`) alongside the hard/direct dependencies. This would be similar to have is done with `rcl_logging_packages` (`rcl` depends on both the default logging implementation *and* the `rcl_logging_packages`group).

christophebedard:
cc @hidmic want me to do it?

hidmic:
That'd be great @christophebedard. I'll make it official.

PR #256, opened 2021-07-15
Use frontend group dependency & explicit dependencies in ros2launch

christophebedard:
Closes https://github.com/ros2/launch_ros/issues/255

Requires https://github.com/ros2/launch/pull/520

As suggested/discussed under https://github.com/ros2/launch/pull/516#issuecomment-880342455, this uses a group dependency for the launch frontends and declares explicit dependencies on `launch_xml` and `launch_yaml` for `ros2launch`.

Using both a group dependency and hard/explicit dependencies allows this to work fine with binaries (with bloom) and also allows us to include any future launch frontend (or any user's custom frontend) in source builds. This is similar to how `rcl` depends on both the `rcl_logging_packages` group and the default logging implementation explicitly.

### DIFF
--- a/ros2launch/package.xml
+++ b/ros2launch/package.xml
@@ -21,6 +21,10 @@
   <depend>ros2cli</depend>
   <depend>ros2pkg</depend>
 
+  <!-- explicitly depend on the main launch frontends -->
+  <exec_depend>launch_xml</exec_depend>
+  <exec_depend>launch_yaml</exec_depend>
+
   <test_depend>ament_copyright</test_depend>
   <test_depend>ament_flake8</test_depend>
   <test_depend>ament_pep257</test_depend>
--- a/ros2launch/package.xml
+++ b/ros2launch/package.xml
@@ -1,8 +1,8 @@
 <?xml version="1.0"?>
 <?xml-model
-  href="http://download.ros.org/schema/package_format2.xsd"
+  href="http://download.ros.org/schema/package_format3.xsd"
   schematypens="http://www.w3.org/2001/XMLSchema"?>
-<package format="2">
+<package format="3">
   <name>ros2launch</name>
   <version>0.15.0</version>
   <description>
--- a/ros2launch/package.xml
+++ b/ros2launch/package.xml
@@ -30,6 +30,8 @@
   <test_depend>ament_pep257</test_depend>
   <test_depend>python3-pytest</test_depend>
 
+  <group_depend>launch_frontend_packages</group_depend>
+
   <export>
     <build_type>ament_python</build_type>
   </export>


christophebedard:
@hidmic @wjwwood here's (along with https://github.com/ros2/launch/pull/520) the solution I'm proposing based on your comments.

wjwwood (approved):
lgtm

wjwwood:
CI (`--packages-above-and-dependencies launch_testing launch_xml launch_yaml ros2launch`, cyclone only):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14895/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9638/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12560/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15049/)

wjwwood:
Anyone have any idea why we have failures on Windows now? I didn't investigate, but it could be to new missing dependencies?

hidmic:
> Anyone have any idea why we have failures on Windows now? I didn't investigate, but it could be to new missing dependencies?

Some `rclpy` libraries is not getting picked up :thinking:

hidmic:
Update: I cannot reproduce it on a plain Windows VM. Something's up with the CI Windows container.

christophebedard:
could this just be a flake?

wjwwood:
I don't see how. If it is a flake, then it is a bug underneath.

wjwwood:
I'll run CI again, just to check:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/14899/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9642/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12568/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15054/)

hidmic:
Same issue. The fun part is that other packages with tests that depend on `rclpy`, like `test_communication`, don't seem to have any issues.

christophebedard:
> The fun part is that other packages with tests that depend on `rclpy`, like `test_communication`, don't seem to have any issues.

yeah `launch_xml` by itself doesn't depend (indirectly or directly) on `rclpy`:

```
$ colcon list -n --packages-up-to launch_xml
ament_copyright
ament_flake8
ament_index_python
ament_lint
ament_pep257
launch
launch_xml
osrf_pycommon
```

this is a bit confusing.

hidmic:
> yeah and `launch_xml` by itself doesn't depend (indirectly or directly) on `rclpy`:

It does indirectly through `launch_ros`, when `launch` loads all the extensions it knows about:

```
test\launch_xml\test_executable.py:62: in test_executable_wrong_subtag
    parser.parse_description(root_entity)
C:\ci\ws\install\Lib\site-packages\launch\frontend\parser.py:117: in parse_description
    actions = [self.parse_action(child) for child in entity.children]
C:\ci\ws\install\Lib\site-packages\launch\frontend\parser.py:117: in <listcomp>
    actions = [self.parse_action(child) for child in entity.children]
C:\ci\ws\install\Lib\site-packages\launch\frontend\parser.py:93: in parse_action
    self.load_launch_extensions()
C:\ci\ws\install\Lib\site-packages\launch\frontend\parser.py:77: in load_launch_extensions
    entry_point.load()
c:\python38\lib\importlib\metadata.py:77: in load
    module = import_module(match.group('module'))
c:\python38\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:1014: in _gcd_import
    ...
C:\ci\ws\install\Lib\site-packages\launch_ros\__init__.py:17: in <module>
    from . import actions
C:\ci\ws\install\Lib\site-packages\launch_ros\actions\__init__.py:17: in <module>
    from .composable_node_container import ComposableNodeContainer
C:\ci\ws\install\Lib\site-packages\launch_ros\actions\composable_node_container.py:27: in <module>
    from .node import Node
C:\ci\ws\install\Lib\site-packages\launch_ros\actions\node.py:61: in <module>
    from rclpy.validate_namespace import validate_namespace
C:\ci\ws\install\Lib\site-packages\rclpy\validate_namespace.py:15: in <module>
    from rclpy.exceptions import InvalidNamespaceException
C:\ci\ws\install\Lib\site-packages\rclpy\exceptions.py:15: in <module>
    from rclpy.impl.implementation_singleton import rclpy_implementation as _rclpy
C:\ci\ws\install\Lib\site-packages\rclpy\impl\implementation_singleton.py:32: in <module>
    rclpy_implementation = import_c_library('._rclpy_pybind11', package)
C:\ci\ws\install\Lib\site-packages\rpyutils\import_c_library.py:39: in import_c_library
    return importlib.import_module(name, package=package)
c:\python38\lib\importlib\__init__.py:127: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
E   ImportError: DLL load failed while importing _rclpy_pybind11: The specified module could not be found.
E   The C extension 'C:\ci\ws\install\Lib\site-packages\rclpy\_rclpy_pybind11.cp38-win_amd64.pyd' failed to be imported while being present on the system. Please refer to 'https://index.ros.org/doc/ros2/Troubleshooting/Installation-Troubleshooting/#import-failing-even-with-library-present-on-the-system' for possible solutions
```

christophebedard:
> It does indirectly through `launch_ros`, when `launch` loads all the extensions it knows about:

yeah, in practice, so isn't that kind of problematic/isn't that the reason why the test is failing?

`launch_xml` doesn't depend on `rclpy`, either directly or indirectly, *according to the `package.xml` files*. This means that, when the test gets run, `PYTHONPATH` doesn't contain the path to `rclpy`'s build directory or the path to `rclpy`'s `lib/python3.8/site-packages` directory. It only contains the paths to its declared direct/indirect dependencies' build/lib directories.

I don't understand why it only fails on Windows and why it seems to be working fine without this PR, and maybe I'm misunderstanding something, but this "undeclared" dependency seems problematic to me.

Note that `launch_xml` doesn't depend on `launch_ros` either. If we didn't build `launch_ros` in the CI run, I'm guessing it wouldn't be found by `launch`'s extension search when running `launch_xml` tests, and so it wouldn't try to find `rclpy` libs, and thus it wouldn't fail, right?

christophebedard:
> Note that `launch_xml` doesn't depend on `launch_ros` either. If we didn't build `launch_ros` in the CI run, I'm guessing it wouldn't be found by `launch`'s extension search when running `launch_xml` tests, and so it wouldn't try to find `rclpy` libs, and thus it wouldn't fail, right?

I re-ran the original Windows job with this PR and the launch PR, but with `--packages-up-to/--packages-select launch_xml launch_yaml`. `launch_ros` isn't built since I removed `ros2launch` from the build args. It runs fine:

* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15064/)

christophebedard:
This might be related to `PATH` instead of `PYTHONPATH`, but I assume it's the same problem https://github.com/ros2/rpyutils/blob/3e41fe4d71bc16f5a727a010f835fb49376b52b3/rpyutils/import_c_library.py#L35-L39

hidmic:
> launch_xml doesn't depend on rclpy, either directly or indirectly, according to the package.xml files. This means that, when the test gets run, PYTHONPATH doesn't contain the path to rclpy's build directory or the path to rclpy's lib/python3.8/site-packages directory. It only contains the paths to its declared direct/indirect dependencies' build/lib directories.

You're absolutely right. 

> Note that launch_xml doesn't depend on launch_ros either. If we didn't build launch_ros in the CI run, I'm guessing it wouldn't be found by launch's extension search when running launch_xml tests, and so it wouldn't try to find rclpy libs, and thus it wouldn't fail, right?

That's also right.

> This might be related to PATH instead of PYTHONPATH,

It could be. We need a Windows-specific explanation as to why `launch_ros` ends up in the `PYTHONPATH` for `launch_xml` while `rclpy` does not.

christophebedard:
> It could be. We need a Windows-specific explanation as to why `launch_ros` ends up in the `PYTHONPATH` for `launch_xml` while `rclpy` does not (neither a dependency of `launch_xml` as far as `package.xml` metadata implies).

I'll try to make time to look into it this week.

wjwwood:
Has anyone had a chance to look into this?

christophebedard:
I haven't had time yet unfortunately. I'll try to look into it tomorrow.

wjwwood:
Yeah, no pressure, just curious.

christophebedard:
I tried to reproduce and I couldn't either. I haven't had time to dig deeper.

jacobperron:
> This might be related to PATH instead of PYTHONPATH, but I assume it's the same problem https://github.com/ros2/rpyutils/blob/3e41fe4d71bc16f5a727a010f835fb49376b52b3/rpyutils/import_c_library.py#L35-L39

For some context, since Python 3.8, `PATH` (and the current working directory) are no longer used for DLL resolution on Windows: https://docs.python.org/3/whatsnew/3.8.html#bpo-36085-whatsnew

To work around this, we introduced that utility in `rpyutils`, which explicitly add directories from `PATH` to the search path when importing a module. (TBH, the whole thing seems to subvert the security measure taken by Python, but I'm not sure how we might handle it differently for ROS 2).

---

I guess what's happening in this case is that the `rclpy` DLL directory is not in `PATH` at test time for `launch_xml`. I'm not sure why, but maybe because it doesn't have `launch_ros` in it's dependency tree (as pointed out earlier).

christophebedard:
Thanks for the info @jacobperron.

I once again tried to reproduce and couldn't. Without being able to reproduce on my local machine, I'm not really sure how to proceed, especially since this is quite time-consuming for such a minor change.

wjwwood:
Well, let's just see what CI says again, since you rebased (I'm assuming that's all):

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10272/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13256/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15808/)

christophebedard:
Yeah I simply rebased.

Those CI jobs don't build/test the same packages as the jobs that failed above, like this one: https://ci.ros2.org/job/ci_windows/15054/

christophebedard:
Here is a run with the same packages:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15592/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10275/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13259/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15811/)

wjwwood:
Oh sorry, I forgot to add https://github.com/ros2/launch/pull/520.

christophebedard:
CI is green, so... I guess it's fixed‽

The issue, as I understood it, was slightly concerning, so I hope this doesn't come back to bite us.

jacobperron:
I think the issue was the same as https://github.com/ros2/rclpy/issues/831
Which we recently fixed in https://github.com/ros2/launch_ros/pull/278

jacobperron (approved):

PR #259, opened 2021-08-13
Support both parameter file configurations for composable nodes

### LINKED ISSUE #156: ComposableNode.parameters with a file path does not take effect.
## Bug report

1. I used  `ros2 param dump /exampleConfig` to generate `exampleConfig.yaml`.
2. I set `parameters=['/module/exampleConfig.yaml']` in my ComposableNode within a launch.py file
3. The parameters were not set, but I do get a warning from libyaml, so I think it's trying but not succeeding...

**Required Info:**

- Operating System: Ubuntu 20.04
- Installation type: Binary
- Version or commit hash: Foxy
- DDS implementation: Fast-RTPS
- Client library (if applicable): rclpy

#### Steps to reproduce issue

launch.py
```python
"""Launch in a component container."""

import launch
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode


    """Generate launch description with multiple components."""
    ld = launch.LaunchDescription()

    ld.add_action(ComposableNodeContainer(
        name='example_container',
        namespace=[],
        package='rclcpp_components',

            ComposableNode(
                package='example',
                plugin='ExampleConfig',
                name='exampleConfig',
                parameters=['/module/exampleConfig.yaml']),
        ],
        output='screen',
    ))
    return ld
```
example_config.cc
```cpp
// Copyright 2019 American Haval Motor Technology.
// All Rights Reserved.
// CONFIDENTIAL - NOT FOR RELEASE
#include <cstdint>
#include <iostream>

#include "rclcpp/rclcpp.hpp"

/**
 * This Example Class is used to show how to read parameters stored in the
 * params.yaml.
 */
class ExampleConfig : public rclcpp::Node {
   public:
    explicit ExampleConfig(const rclcpp::NodeOptions &options)
        // The node name `config` is used in the params.yaml file.
        : Node("config", options) {
        RCLCPP_INFO(this->get_logger(), "Starting %s.", this->get_name());

        // Parameters

        // example int (meters)
        // This is a simple example of an integer parameter
        int example_int = declare_parameter("example_int", 42);

        // example float (hz)
        // This is a simple example of an float parameter
        double example_float = declare_parameter("example_float", 23.0);

        // example string
        // This is a simple example of an string parameter
        std::string example_string =
            declare_parameter("example_string", std::string("Default Value"));

        // Log the parameter values as an example
        RCLCPP_INFO(this->get_logger(), "example_int: %d", example_int);
        RCLCPP_INFO(this->get_logger(), "example_float: %f", example_float);
        RCLCPP_INFO(this->get_logger(), "example_string: %s", example_string.c_str());

        // flushes values from the buffer to the active log.
        std::flush(std::cout);
    }

   private:
    // Prevents this Node from being copied
    RCLCPP_DISABLE_COPY(ExampleConfig)
};

// Used to register this node as a ros 2 component
#include "rclcpp_components/register_node_macro.hpp"
RCLCPP_COMPONENTS_REGISTER_NODE(ExampleConfig)
```

exampleConfig.yaml
```
exampleConfig:
  ros__parameters:
    example_int: 98765
    example_float: 9.8765
    example_string: "example"
    use_sim_time: false
```

#### Expected behavior
parameters should be set in the node

#### Actual behavior
parameters are not set

#### Additional information

Here's the console output of the above:
```
[INFO] [launch]: Default logging verbosity is set to INFO
[INFO] [component_container-1]: process started with pid [12476]
/opt/ros/foxy/lib/python3.8/site-packages/launch_ros/utilities/to_parameters_list.py:49: YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.
  context, normalize_parameter_dict(yaml.load(f))
[component_container-1] [INFO] [1592763714.853802170] [example_container]: Load Library: /module/install/example/lib/libexample_config.so
[component_container-1] [INFO] [1592763714.854225507] [example_container]: Found class: rclcpp_components::NodeFactoryTemplate<ExampleConfig>
[component_container-1] [INFO] [1592763714.854239504] [example_container]: Instantiate class: rclcpp_components::NodeFactoryTemplate<ExampleConfig>
[component_container-1] [INFO] [1592763714.856834111] [exampleConfig]: Starting exampleConfig.
[component_container-1] [INFO] [1592763714.856919332] [exampleConfig]: example_int: 42
[component_container-1] [INFO] [1592763714.856931555] [exampleConfig]: example_float: 23.000000
[component_container-1] [INFO] [1592763714.856943773] [exampleConfig]: example_string: Default Value
[INFO] [launch_ros.actions.load_composable_nodes]: Loaded node '/exampleConfig' in container '/example_container'
```

rebecca-butler:
Closes #156. With these changes, loading a composable node will work with both parameter file configs:

```
param: value
```
and
```
namespace/my_node:
  ros__parameters:
    param: value
```

Signed-off-by: Rebecca Butler <rebecca@openrobotics.org>


launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -45,8 +49,19 @@ def to_parameters_list(
     for params_set_or_path in evaluated_parameters:
         if isinstance(params_set_or_path, pathlib.Path):
             with open(str(params_set_or_path), 'r') as f:
+                if namespace is not None:
+                    node_name = f"{namespace}/{node_name}"
+                node_name_with_slash = f"/{node_name}"
```
ivanpauno, line 54: this might be duplicating the slash, right?
maybe:

```suggestion
                if not node_name.startswith('/'):
                  node_name_with_slash = f"/{node_name}"
```

ivanpauno: and you would also need to compute `node_name_without_slash` if that's the case, to always test for both names in the parameteter dict

ivanpauno (commented):
Seems like a reasonable approach to me:

- Could you add a test for it?
- Could you add a release note related to this?

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -45,8 +49,19 @@ def to_parameters_list(
...
+                param_dict = yaml.safe_load(f)
+                try:
+                    param_dict = param_dict[node_name]["ros__parameters"]
+                except KeyError:
+                    try:
+                        param_dict = param_dict[node_name_with_slash]["ros__parameters"]
+                    except KeyError:
+                        pass
```
hidmic, line 62: @rebecca-butler hmm, a few questions about behavior:

- What if no node name was specified for the node on launch?
- What if there's no entry in the parameter file for that node name? 
- What if a wildcard was used in the file?

rebecca-butler:
I've written a few test cases. While doing this I realized there's another format we should probably support:

```
namespace:
  node_name:
    ros__parameters:
      param: value
```

I changed the implementation so now it combines all the keys that come before "ros__parameters" to get the full name.

launch_ros/launch_ros/utilities/to_parameters_list.py
rebecca-butler: Good point. I think it might be easier to strip the slash here and when getting the name from the dict so we don't have to worry about comparing with & without the slash. Done in 97ede21

rebecca-butler:
@hidmic to address your questions:
> What if no node name was specified for the node on launch?

What is the expected behaviour for this case? Currently if no node name is provided, then parameter files using the `node_name: ros__parameters` format would do nothing since the node name won't match. Parameter files using just the dictionary format will always work regardless of what the node name is.

> What if there's no entry in the parameter file for that node name?

We only remove the `node_name: ros__parameters` part of the dictionary if an entry is found. evaluate_parameter_dict() doesn't work if `node_name: ros__parameters` is still in the dict, so in cases where no entry is found, the parameters aren't evaluated correctly and nothing happens. This means parameters won't be applied to a node unless it matches the name in the yaml file. (The exception is if the yaml file is using the old format that doesn't contain `node_name: ros__parameters`, in which case it will always work.)

> What if a wildcard was used in the file?

Good point, I didn't consider that. I've added support for the `/**` wildcard in 97ede21 since it's easy to do, but I'm not sure how this would work for other wildcards. Do we have a public API somewhere for wildcard matching?

hidmic:
> What is the expected behaviour for this case? Currently if no node name is provided, then parameter files using the `node_name: ros__parameters` format would do nothing since the node name won't match. Parameter files using just the dictionary format will always work regardless of what the node name is.

Assuming `launch` proceeds (silently or with warning) in that case, we should still document this properly i.e. that if `launch` is not explicitly told what the node name is, it won't be able to pick up parameters from a parameter file including `ros__parameters` entries. 

Note this is not the case when launching plain `Node`s (because the process bearing the nodes knows all node names, hard-coded or not, and it can look up the parameter file).

> so in cases where no entry is found, the parameters aren't evaluated correctly and nothing happens.

So long as `launch` handles this case and proceeds, :+1:

 > Do we have a public API somewhere for wildcard matching?

For parameters I think not. There's a [design document](https://design.ros2.org/articles/static_remapping.html) for wildcard matching in remapping rules, from which I believe parameter wildcards are inspired on. Anyhow, IIRC `/**` is the only pattern currently supported.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -45,8 +61,26 @@ def to_parameters_list(
     for params_set_or_path in evaluated_parameters:
         if isinstance(params_set_or_path, pathlib.Path):
             with open(str(params_set_or_path), 'r') as f:
+                node_name.lstrip('/')
+                if namespace:
+                    namespace.lstrip('/')
+                    node_name = f"{namespace}/{node_name}"
```
hidmic, line 67: @rebecca-butler could we move this outside the `for` loop?

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,15 +28,31 @@
...
+def search_for_key(keys, dictionary):
+    """Search for the nested "ros__parameters" key and return the keys above it."""
+    for key, value in dictionary.items():
+        if key == "ros__parameters":
+            return (keys, value)
+        elif type(value) is dict:
+            keys.append(key)
+            return search_for_key(keys, value)
+        else:
+            return (None, None)
```
hidmic, line 40: @rebecca-butler hmm, what if there's several parameter groups in the same file? E.g.:

```
foo:
   node:
       ros__parameters:
             # ...
/**:
   ros__parameters:
       # ...
```

There's a precedence that has to be enforced here too. You may check [precedence in `rclcpp`](https://github.com/ros2/rclcpp/blob/d0cd6bb0a46f5b2841e9f282c9eecfa8cffe1ffa/rclcpp/src/rclcpp/detail/resolve_parameter_overrides.cpp#L56).

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -45,8 +61,26 @@ def to_parameters_list(
...
             with open(str(params_set_or_path), 'r') as f:
+                node_name.lstrip('/')
+                if namespace:
+                    namespace.lstrip('/')
+                    node_name = f"{namespace}/{node_name}"
+                param_dict = yaml.safe_load(f)
+
+                # Get all keys that come before "ros__parameters"
+                keys = []
+                inner_dict = {}
```
hidmic, line 72: @rebecca-butler nit: no need to predefine variables in Python.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -45,8 +61,26 @@ def to_parameters_list(
...
+                (keys, inner_dict) = search_for_key(keys, param_dict)
```
hidmic, line 73: @rebecca-butler nit: parenthesis are redundant when unpacking:
```suggestion
                keys, inner_dict = search_for_key(keys, param_dict)
```

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +28,66 @@
...
+    :param namespace: node namespace
     :param parameters: parameters as either paths to parameter files or name/value pairs.
         See `evaluate_parameters()` documentation for further reference.
     :returns: a list of parameters
     """
     parameters = []  # type: List[rclpy.parameter.Parameter]
+    node_name = node_name.lstrip('/')
+    if not node_name:
+        print('Warning: no node name was provided. Parameter file entries that contain a '
+              'node name will not be applied')
```
hidmic, line 70: @rebecca-butler hmm, we should probably emit a warning if and only if we run into that situation. Also, consider using [`warnings`](https://docs.python.org/3.8/library/warnings.html) for warnings :)

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +28,66 @@
...
+                param_dict = yaml.safe_load(f)
+                formatted_dict = __format_dict(param_dict, keys=[], formatted_dict={})
+
+                if formatted_dict:
+                    param_dict.clear()
+                    if '**' in formatted_dict:
+                        param_dict = formatted_dict['**']
+                    if node_name in formatted_dict:
+                        for param, value in formatted_dict[node_name].items():
+                            param_dict[param] = value
```
hidmic, line 87: @rebecca-butler nit: 
```suggestion
                        param_dict.update(formatted_dict[node_name])
```
has the same effect.

hidmic (commented):
Kudos for all those tests !

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: See 1f1ed6e

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: done: e6de211

jacobperron:
I've address the most recent comments. Ready for another round of review.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +29,69 @@
...
+                    if not warned_once and not node_name:
+                        warnings.warn(
+                            'node name not provided to launch; parameter files will not apply',
+                            UserWarning
+                        )
+                        warned_once = True
+
                 params_set = evaluate_parameter_dict(
-                    context, normalize_parameter_dict(yaml.safe_load(f))
+                    context, normalize_parameter_dict(param_dict)
```
hidmic, line 94: @jacobperron we should skip this if `param_dict` is empty (when a parameter file is given a neither wildcard nor node name yield any parameters).

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +29,69 @@
 from ..parameters_type import EvaluatedParameters
 
 
+def __format_dict(dictionary, keys, formatted_dict):
```
hidmic, line 32: @jacobperron should we try to come up with a better name?

hidmic (commented):
Overall LGTM. `__format_dict` implementation is a bit unusual but OK.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +29,69 @@
...
+    Combine namespaces and their node names into one key and remove the "ros__parameters" key.
+
+    Return an empty dict if the "ros__parameters" key is not found.
+    """
+    for key, value in dictionary.items():
+        if key == 'ros__parameters':
+            node_name = '/'.join(keys)
+            formatted_dict[node_name] = value
+            return formatted_dict
+        elif type(value) is dict:
```
ivanpauno, line 43: do we need an `else` clause?
what should happen if `type(value)` isn't dict.

ivanpauno:
```suggestion
        elif isinstance(value, dict):
```

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,25 +29,69 @@
...
+    :param node_name: node name
+    :param namespace: node namespace
     :param parameters: parameters as either paths to parameter files or name/value pairs.
         See `evaluate_parameters()` documentation for further reference.
     :returns: a list of parameters
     """
     parameters = []  # type: List[rclpy.parameter.Parameter]
+    node_name = node_name.lstrip('/')
+    if namespace:
+        namespace = namespace.lstrip('/')
```
ivanpauno, line 70: should be also `rstrip('/') from the namespace? i.e. just strip('/')

ivanpauno (approved):
I have some minor comments, LGTM!

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: See 9de0767

I don't think an `else` clause is necessary. If we get a value that isn't a dictionary, I think it is safe to ignore it, e.g.

```
foo: 42
bar:
  ros__parameters:
    ...
```

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: I renamed this to `__normalized_parameters_dict` (9de0767)
Please suggest a different name if you think of a better one.

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: Donee: 9de0767

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -28,26 +29,80 @@
...
+            # Return
+            if key == 'ros__parameters':
+                node_name = '/'.join(keys)
+                result_dict[node_name] = value
+                return result_dict
+
+            if isinstance(value, dict):
+                keys.append(key.lstrip('/'))
+                result_dict = normalize_parameters_dict(value, keys, result_dict)
+                keys = []
```
ivanpauno, line 51:
```suggestion
```

ivanpauno (approved):

launch_ros/launch_ros/utilities/to_parameters_list.py
jacobperron: removed: 5adffa8

jacobperron:
On a whim, I've pushed a commit that makes the failing test use a copy of `example_parameters.yaml` (bd1bc63).

ivanpauno (approved):

guru-florida:
I just ran into this on Galactic as well. Seems like an odd and unexpected thing to change parameter config between regular nodes and components. Thanks to everyone here who worked on this!  (awaiting back-patch but glad to have a work-around until then.)

PR #260, opened 2021-08-18
launch_ros action to set parameters for all nodes in scope using a global parameter yaml file

adityapande-1995:
This PR aims to add a new ```launch_ros``` action, ```SetParametersFromFile``` which can accept a yaml file with ROS parameters which can be passed to all nodes in scope. 

### The parameter file:
The parameter yaml file must be written in the following format:
```
demo_node_1:
  ros__parameters:
    param_1: 1
    param_2: 5

demo_node_2:
  ros__parameters:
    param_2: 10
    param_4: hello_world
```

### Launch file in xml:
```demo_node_3``` is not in the same scope, so the ```SetParametesFromFile``` action won't affect it.
```
<launch>
  <group>
    <set_parameters_from_file filename='/path/to/param/file.yaml'>
    <node pkg="demo_nodes_cpp" exec="parameter_blackboard" name="demo_node_1">
    <node pkg="demo_nodes_cpp" exec="parameter_blackboard" name="demo_node_2">
  </group>
  <node pkg="demo_nodes_cpp" exec="parameter_blackboard" name="demo_node_3">
</launch>
```

### Equivalent usage in python:
```
from launch_ros.actions import Node, SetParametersFromFile

  return LaunchDescription([
     GroupAction(
       actions = [
          SetParametersFromFile('path/to/yaml/file.yaml'),

          Node(
            package = 'demo_nodes_cpp',
            executable = 'parameter_blackboard',
            name = 'demo_node_1'
          ),

          Node(
            package = 'demo_nodes_cpp',
            executable = 'parameter_blackboard',
            name = 'demo_node_2'
          )
        ]
      ),

       Node(
          package = 'demo_nodes_cpp',
          executable = 'parameter_blackboard',
          name = 'demo_node_3'
        )
  ])

```

# TODO 
* Add test cases
* Are ```@property``` methods required for ```SetParametersFromFile``` class ? For e.g. https://github.com/ros2/launch_ros/blob/ef1502ea57f16de388635a86861cb5e84e34c919/launch_ros/launch_ros/actions/set_parameter.py#L76
* Cleaner / better way to extract raw text from ```SomeSubstitutionType``` objects when parsing xml launch files

### CODE REVIEW

**Comment by ivanpauno** [comment]
This is only adding the parameterers to Node actions but not to loaded components, which might be a bit unexpected by users.
Maybe we can fix that after https://github.com/ros2/launch_ros/pull/259 is merged.

File: launch_ros/launch_ros/actions/node.py [hunk 1]
```diff
@@ -419,6 +419,30 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
         self.__final_node_name = prefix_namespace(
             self.__expanded_node_namespace, self.__expanded_node_name)
 
+        # Check to see if a global parameter file was provided
+        # First extract the raw name of the node
+        node_name_as_text = None
+        if type(self.__node_name) == list:
+            node_name_as_text = list(self.__node_name[0].__dict__.values())[0]
+        elif type(self.__node_name) == str:
+            node_name_as_text = self.__node_name
```
ivanpauno, line 428: this is not the correct way of using substitutions, I think `self.__final_node_name` in L419 is already what you want

launch_ros/launch_ros/actions/node.py
```diff
@@ -419,6 +419,30 @@ def _perform_substitutions(self, context: LaunchContext) -> None:
...
+
+        params_from_file = context.launch_configurations.get('global_params_from_file', None)
+        node_params_from_file = None
+        # Check if ROS parameters were provided in the file for this node
+        if params_from_file is not None and node_name_as_text is not None:
+            if node_name_as_text in params_from_file:
+                if 'ros__parameters' in params_from_file[node_name_as_text]:
+                    node_params_from_file = params_from_file[node_name_as_text]['ros__parameters']
```
ivanpauno, line 436: you can already use wildcards in node names, e.g. /**.
Maybe it's easier to pass the parameter file to all nodes, `rcl` will already only read the parameters that actually applied to the node

launch_ros/launch_ros/actions/set_parameters_from_file.py
```diff
@@ -0,0 +1,91 @@
...
+    """
+
+    def __init__(
+        self,
+        filename: SomeSubstitutionsType,
+        **kwargs
+    ) -> None:
+        """Create a SetParameterFromFile action."""
+        super().__init__(**kwargs)
+        input_file = self.extract_raw_text(filename)
```
ivanpauno, line 70: please check how [normalize_to_list_of_substitutions()](https://github.com/ros2/launch/blob/b59bb6b88b40eaa82870542d93bdfb1a033d05c3/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py#L26) and [perform_substitutions](https://github.com/ros2/launch/blob/b59bb6b88b40eaa82870542d93bdfb1a033d05c3/launch/launch/utilities/perform_substitutions_impl.py#L24) are used in other places.

You need to use the first one in the constructor and the second one in the execute() method

launch_ros/launch_ros/actions/node.py
adityapande-1995: I added this in case we wanted to pass the same parameter with different values for different nodes. For e.g., ```demo_node_1``` has ```param_2: 5``` and ```demo_node_2``` has ```param_2: 10``` in the example above.

launch_ros/launch_ros/actions/node.py
adityapande-1995: Also, support for namespaces to be added

launch_ros/launch_ros/actions/set_parameters_from_file.py
adityapande-1995: fixed

launch_ros/launch_ros/actions/__init__.py
```diff
@@ -34,5 +35,6 @@
     'RosTimer',
     'SetParameter',
     'SetRemap',
-    'SetUseSimTime'
+    'SetUseSimTime',
+    'SetParametersFromFile'
```
rebecca-butler, line 39: nit: prefer alphabetical order (this should be above 'SetRemap')

launch_ros/launch_ros/actions/set_parameters_from_file.py
```diff
@@ -0,0 +1,80 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
```
rebecca-butler, line 1: nit: 2021

test_launch_ros/test/test_launch_ros/actions/example_parameters_1.yaml
```diff
@@ -0,0 +1,12 @@
+/**:
```
ivanpauno, line 1: nit^N: Now we have a file named, `example_parameters.yaml` and another one named `example_parameters_1.yaml` in the same folder.

I would rather use the same style in both, i.e. `example_parameters_0.yaml` and `example_parameters_1.yaml` (_1 and _2 is also fine).

test_launch_ros/test/test_launch_ros/actions/test_set_parameters_from_file.py
```diff
@@ -0,0 +1,136 @@
...
+
+    with open(actual_command_2[8], 'r') as h:
+        expanded_parameters_dict = yaml.load(h, Loader=yaml.FullLoader)
+        assert expanded_parameters_dict == {
+            '/my_ns/my_other_node': {
+                'ros__parameters': {'asd': 'bsd2'}
+            }
+        }
+
+# TODO: (Aditya) Add test case for composable node
```
ivanpauno, line 136: it would be great to add that before merging

ivanpauno (approved):
LGTM!

test_launch_ros/test/test_launch_ros/actions/test_set_parameters_from_file.py
adityapande-1995: ```SetParametersFromFile``` currently only works for nodes and not composable nodes.
As composable nodes and nodes currently use different styles for accepting parameters from yaml files, https://github.com/ros2/launch_ros/pull/259 would need to be closed before I can add that functionality to composable nodes. We could:

1. Add a note to ```SetParametersFromFile``` saying that this won't affect composable nodes, and close this PR. Implement it in a new PR once #259 is merged.
2. Wait for #259 to be merged.

Personally, I think we can proceed with option 1. How should we proceed ? @ivanpauno

ivanpauno: Yes, merging this and iterating after https://github.com/ros2/launch_ros/pull/259 is finished sounds good to me.

ivanpauno (approved):
LGTM with green CI

test_launch_ros/test/test_launch_ros/actions/test_set_parameter.py
```diff
@@ -42,32 +45,27 @@ def get_set_parameter_test_parameters():
...
             [('my_param', '2')],  # to set
-            {'my_param': '2'},  # expected
+            [('my_param', '2')],  # expected
             id='One param'
         ),
-        pytest.param(
-            [('my_param', '2'), ('another_param', '2'), ('my_param', '3')],
-            {'my_param': '3', 'another_param': '2'},
-            id='Two params, overriding one'
-        ),
```
jacobperron, line 52: Why remove this test case?

test_launch_ros/test/test_launch_ros/actions/test_set_parameters_from_file.py
```diff
@@ -0,0 +1,136 @@
+# Copyright 2020 Open Source Robotics Foundation, Inc.
```
jacobperron, line 1:
```suggestion
# Copyright 2021 Open Source Robotics Foundation, Inc.
```

jacobperron (approved):
LGTM! Just a couple minor comments.

test_launch_ros/test/test_launch_ros/actions/test_set_parameter.py
adityapande-1995: when ``my_param`` is passed again with a different value, it can be reassigned due to ``ros__params`` being a dictionary, with ``my_param`` as a key, and ``3`` as its value. Since we shifted to a list of tuples to preserve order, this would be passed as ``-p my_param:=2 -p another_param:=2 -p my_param:=3`` which would override ``my_param`` anyway.  Thus this test seemed redundant, and won't work as it was originally intended.

adityapande-1995: I did try adding a logic to replace the original declaration when it is overridden, but it seemed redundant as ``rcl`` itself is doing it anyway down the line

jacobperron: I understand, thanks for explaining; makes sense to me, we can leave it as-is.

adityapande-1995:
CI : 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15148/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9878/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12837/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15328/)

PR #261, opened 2021-08-18
Add regex filter for selective launch-prefix application

### LINKED ISSUE #257: Selective application of launch-prefix from CLI
#### Feature description
ros2/launch_ros#254 adds the `--launch-prefix` flag to provide a means of prepending the same prefix (e.g. `xterm -e valgrind`) to each executable command in the launch file. For large launch files this launches many instances of the desired prefix program even if the user may only require it for certain executables. Currently selectively applying prefixes can be achieved by editing the launch file and directly specifying the applied prefix for a given executable; however, this is a slow process if the prefix is frequently changed or the developer only wishes to use a prefix one time.

To provide a means of filtering which executables the `--launch-prefix` value is applied to, I propose that we add a `--launch-prefix-filter` flag (or something to this effect) which allows the user to provide a list of executable names for which the prefix should be applied. When used alone `--launch-prefix` would apply the prefix to all executables, and when used in conjunction with the new `--launch-prefix-filter` flag, the prefix would only be applied to matching executables.

The goal of these two flags is to provide a faster means of debugging multiple, specific executables without requiring the launch file to be edited.

For the `launch_ros` repo:

- Add `--launch-prefix-filter` argument which accepts at least 1 executable name
- Pass list of executable names as a `/` separated list via `service.context.launch_configurations` (this is because Unix & Windows filenames cannot contain `/`)

For the `launch` repo (Assuming ros2/launch_ros#254 provides the prefix via `service.context.launch_configurations`):

- Modify `ExecuteProcess.__expand_substitutions()` such that when a launch prefix filter is available in the launch configuration, the prefix is only prepended to the command if its executable name matches an executable name that was provided via the launch configuration

The `name` property within `ExecuteProcess` is purposely not used so that the behavior is the same between `Node` and `ExecuteProcess`. `Node` sets the `ExecuteProcess` name to the executable name by default, but when the user creates their own `ExecuteProcess` in a launch file, they are able to provide their own `name`, overriding it from being set to the executable name. This could cause an issue where user-named `ExecuteProcess` elements wouldn't be properly selected by the filter, so it's executable name should be extracted directly from the `cmd` property instead.

camm73:
Resolves #257 

This PR shouldn't be merged until #254 and https://github.com/ros2/launch/pull/522 have been accepted and merged.

This PR adds the `--launch-prefix-filter` argument discussed in #257 which allows the user to provide a regex pattern that can be used for selectively applying the command prefix (provided via the `--launch-prefix` argument) to executables which match the pattern. This will allow an easier debugging experience without having to edit the launch file.

### DIFF
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -156,7 +156,7 @@ def launch_a_launch_file(
     if args and args.launch_prefix:
         launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
 
-    if args.launch_prefix_filter is not None and len(args.launch_prefix_filter):
+    if args and args.launch_prefix_filter:
         launch_file_arguments.append(f'launch-prefix-filter:={args.launch_prefix_filter}')
 
     launch_service = launch.LaunchService(
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -156,6 +156,15 @@ def launch_a_launch_file(
     if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
         launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
 
+    # Pass launch prefix filter as a forward-slash separated list
+    if args is not None and args.launch_prefix_filter is not None:
+        filter_str = ''
+        for exec_str in args.launch_prefix_filter:
+            filter_str += f'{exec_str}/'
+        filter_str = filter_str.strip('/')
+        if len(filter_str):
+            launch_file_arguments.append(f'launch-prefix-filter:={filter_str}')
+
     launch_service = launch.LaunchService(
         argv=launch_file_arguments,
         noninteractive=noninteractive,
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -153,7 +153,7 @@ def launch_a_launch_file(
 
     # If 'launch-prefix' launch file argument is also provided in the user input,
     # the 'launch-prefix' option is applied since the last duplicate argument is used
-    if args.launch_prefix is not None and len(args.launch_prefix):
+    if args and args.launch_prefix:
         launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
 
     if args.launch_prefix_filter is not None and len(args.launch_prefix_filter):
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -153,17 +153,11 @@ def launch_a_launch_file(
 
     # If 'launch-prefix' launch file argument is also provided in the user input,
     # the 'launch-prefix' option is applied since the last duplicate argument is used
-    if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
+    if args.launch_prefix is not None and len(args.launch_prefix):
         launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
 
-    # Pass launch prefix filter as a forward-slash separated list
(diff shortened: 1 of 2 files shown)


hidmic:
This needs a rebase.

camm73:
Should be good to go now @hidmic. Thanks for reviewing all these PRs!

ros2launch/ros2launch/api/api.py
```diff
@@ -153,9 +153,12 @@ def launch_a_launch_file(
 
     # If 'launch-prefix' launch file argument is also provided in the user input,
     # the 'launch-prefix' option is applied since the last duplicate argument is used
-    if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
+    if args.launch_prefix is not None and len(args.launch_prefix):
```
hidmic, line 156: @camm73 how about:
```suggestion
    if args and args.launch_prefix:
```
instead? `args` may still be `None` and using `len` instead of type coercion doesn't provide much of an advantage here IMHO. Same below.

ros2launch/ros2launch/command/launch.py
```diff
@@ -97,6 +97,12 @@ def add_arguments(self, parser, cli_name):
                  'Command must be wrapped in quotes if it contains spaces '
                  "(e.g. --launch-prefix 'xterm -e gdb -ex run --args')."
         )
+        parser.add_argument(
+            '--launch-prefix-filter',
+            type=str,
```
hidmic, line 102: @camm73 this is implied, we can drop it.

ros2launch/ros2launch/api/api.py
```diff
@@ -153,9 +153,12 @@ def launch_a_launch_file(
 
     # If 'launch-prefix' launch file argument is also provided in the user input,
     # the 'launch-prefix' option is applied since the last duplicate argument is used
-    if args is not None and args.launch_prefix is not None and len(args.launch_prefix) > 0:
+    if args and args.launch_prefix:
         launch_file_arguments.append(f'launch-prefix:={args.launch_prefix}')
 
+    if args.launch_prefix_filter is not None and len(args.launch_prefix_filter):
```
hidmic, line 159: @camm73 the same comment I made above applies here too:
```suggestion
    if args and args.launch_prefix_filter:
```

hidmic (approved):
CI up to `ros2launch`:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15157/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9887/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12848/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15341/)

ros2launch/ros2launch/api/api.py
camm73: Good catch—I missed that one. Fixing now

camm73:
> CI up to `ros2launch`:
> 
> * Linux [Build Status](http://ci.ros2.org/job/ci_linux/15157/)
> * Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9887/)
> * macOS [Build Status](http://ci.ros2.org/job/ci_osx/12848/)
> * Windows [Build Status](http://ci.ros2.org/job/ci_windows/15341/)

My flake8 didn't flag a quote issue in `launch.py`. Just pushed a fix

hidmic:
CI, take two:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15158/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/9888/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/12849/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15342/)

hidmic:
Alright, all green. Going in.

PR #268, opened 2021-09-21
Additional Tests for evaluate_parameters

DLu:
Related to https://github.com/ros2/launch/issues/508

I've added some crashing tests that fail when the output from a `Command` evaluates to something `: ` in it. 

`test_command_with_colon_output` results in 
```
E                           TypeError: Allowed value types are bytes, bool, int, float, str, Sequence[bool], Sequence[int], Sequence[float], Sequence[str]. Got <class 'dict'>.

/opt/ros/rolling/lib/python3.8/site-packages/launch_ros/utilities/evaluate_parameters.py:88: TypeError
```

`test_command_with_two_colons` results in
```
E                   yaml.scanner.ScannerError: mapping values are not allowed here
E                     in "<unicode string>", line 1, column 24:
E                       <robot><!-- two: colons: (with spaces)--></robot>
E                                              ^

/usr/lib/python3/dist-packages/yaml/scanner.py:577: ScannerError

```


DLu:
@hidmic The tests are failing here by design. Do you agree that these tests *should* pass?

hidmic:
> The tests are failing here by design. Do you agree that these tests should pass?

Agreed.

DLu:
Two notes: 
 * This PR could be massively simplified if I just modify `test_unallowed_yaml_types_in_substitutions`, which I hadn't seen before today. Commands need not necessarily be involved. 
 * I converted the first bit of `test_unallowed_yaml_types_in_substitutions` to be a different test `test_yaml_with_invalid_type_as_string`, but arguably, ALL of the tests in `test_unallowed_yaml_types_in_substitutions` should return a string instead of throwing an exception. I'd argue that in all cases where attempting to save the evaluated yaml would lead to an exception, it is just evaluated as a string instead.

hidmic:
> I'd argue that in all cases where attempting to save the evaluated yaml would lead to an exception, it is just evaluated as a string instead.

Hmm, that could hide issues.

---

I'm going to backtrack with my previous comment and note that you can wrap your substitutions that potentially expand into invalid YAML as follows:

```python
launch_ros.parameter_descriptions.ParameterVaue(..., value_type=str)
```

That'll ensure no YAML parsing is carried out. It is a bit verbose, though gets much cleaner whenever using XML or YAML launch descriptions e.g.:

```xml
<param name="..." value="..."  type="str"/>
```

Sorry for not bringing this up earlier @DLu. I just hit me we do have a code path for type coercion.

DLu:
So in [this launch file](https://github.com/hello-robot/stretch_ros2/blob/6fcfb2df5add17873d23ad622fa48a8c1605e9c0/stretch_description/launch/display.launch.py#L23), would this be the proper way to wrap it? 

```
    robot_state_publisher_node = Node(package='robot_state_publisher',
                                      executable='robot_state_publisher',
                                      output='both',
                                      parameters=[{'robot_description': ParameterVaue(robot_description_content, value_type=str),
                                                   'publish_frequency': 15.0}])
```

Issue #272, opened 2021-09-30
Feature request (``launch_testing``): Check if a topic is publishing messages, with a timeout


adityapande-1995:
#### Feature description
It would be nice to have a way to check if a topic is publishing messages (with a timeout). 
THis suggestion came up in https://github.com/ros2/launch_ros/pull/263

#### Implementation considerations
One way to do that can be :
```
from threading import Event, Thread
import unittest

import launch
import launch.actions
import launch_ros.actions
import launch_testing.actions
import pytest
import rclpy
from rclpy.node import Node
from std_msgs.msg import String


@pytest.mark.launch_test
def generate_test_description():
    return launch.LaunchDescription([
        launch_ros.actions.Node(
            package='demo_nodes_cpp',
            executable='talker',
            name='demo_node_1',
        ),

        launch_testing.actions.ReadyToTest()
    ])


class TestFixture(unittest.TestCase):

    def test_check_if_msgs_published(self, proc_output):
        rclpy.init()
        node = MakeTestNode('test_node')
        node.start_subscriber()
        msgs_received_flag = node.msg_event_object.wait(timeout=5.0)
        assert msgs_received_flag, 'Did not receive msgs !'
        rclpy.shutdown()


class MakeTestNode(Node):

    def __init__(self, name='test_node'):
        super().__init__(name)
        self.msg_event_object = Event()

    def start_subscriber(self):
        # Create a subscriber
        self.subscription = self.create_subscription(
            String,
            'chatter',
            self.subscriber_callback,
            10
        )

        # Add a spin thread
        self.ros_spin_thread = Thread(target=lambda node: rclpy.spin(node), args=(self,))
        self.ros_spin_thread.start()

    def subscriber_callback(self, data):
        self.msg_event_object.set()

```

jacobperron:
IMO, ideally syntax would look like this:
```
def test_foo(self):
    wait_for_topics = WaitForTopics(['foo', 'bar', 'baz'])
    assert wait_for_topics.wait(timeout=5.0)
    ...
```

However, we need an instance to a ROS node (and context). Maybe we can use the existing [`ROSAdapter`](https://github.com/ros2/launch_ros/blob/999b208a05998a83378f7ba273b8146ea3eb09d9/launch_ros/launch_ros/ros_adapters.py#L29) class, e.g.:

```
class WaitForTopics:
   def __init__(self, *, topics):
       self.__ros_adapter = ROSAdapter()
       self.__ros_adapter.start()

[...]

PR #274, opened 2021-10-04
launch testing : Wait for topics to publish

adityapande-1995:
This aims to close https://github.com/ros2/launch_ros/issues/272. It aims to provide an API in to be used with ``launch_testing`` where we provide a set of topics and we wait for at least a single message to appear on all of them.

## Sample usage : 
```
import unittest
import launch
import launch.actions
import launch_ros.actions
import launch_testing.actions
import launch_testing.markers
import pytest
import rclpy
from rclpy.node import Node
from std_msgs.msg import String

from launch_testing_ros import WaitForTopics

def generate_node(i):
    return launch_ros.actions.Node(
        executable='talker',
        package='demo_nodes_cpp',
        name='demo_node_' + str(i),
        remappings = [('chatter', 'chatter_' + str(i))]
    )

@pytest.mark.launch_test
@launch_testing.markers.keep_alive
def generate_test_description():
    # 'n' changes the number of nodes and topics
    n = 5
    description = [generate_node(i) for i in range(n)] + [launch_testing.actions.ReadyToTest()]
    return launch.LaunchDescription(description), {'count':n}

class TestFixture(unittest.TestCase):

    def test_check_if_msgs_published(self, proc_output, count):

        topic_list = [('chatter_' + str(i), String) for i in range(count)]
        with WaitForTopics(topic_list, timeout=5.0):
            print("The topics are alive !")


```

### DIFF
--- a/launch_testing_ros/launch_testing_ros/wait_for_topics.py
+++ b/launch_testing_ros/launch_testing_ros/wait_for_topics.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import random
+import string
 from threading import Event
 from threading import Lock
 from threading import Thread
@@ -22,21 +24,39 @@ from rclpy.node import Node
 
 
 class WaitForTopics:
-    """Wait to receive messages on supplied nodes."""
+    """
+    Wait to receive messages on supplied topics.
+
+    Example usage:
+    --------------
+
+    from std_msgs.msg import String
+
+    # Method 1, using the 'with' keyword
+    def method_1():
+        topic_list = [('topic_1', String), ('topic_2', String)]
+        with WaitForTopics(topic_list, timeout=5.0):
+            # 'topic_1' and 'topic_2' received at least one message each
+            print('Given topics are receiving messages !')
+
+    # Method 2, calling wait() and shutdown() manually
+    def method_2():
+        topic_list = [('topic_1', String), ('topic_2', String)]
+        wait_for_topics = WaitForTopics(topic_list, timeout=5.0)
+        assert wait_for_topics.wait()
+        print('Given topics are receiving messages !')
+        wait_for_topics.shutdown()
+    """
 
     def __init__(self, topic_tuples, timeout=5.0):
-        self.__ros_context = None
-        self.__ros_node = None
-        self.__ros_executor = None
         self.topic_tuples = topic_tuples
         self.timeout = timeout
-        self.start()
-
-    def start(self):
         self.__ros_context = rclpy.Context()
         rclpy.init(context=self.__ros_context)
         self.__ros_executor = SingleThreadedExecutor(context=self.__ros_context)
-        self.__ros_node = MakeTestNode(name='test_node', node_context=self.__ros_context)
+        node_name = '_test_node_' +\
+            ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
+        self.__ros_node = _WaitForTopicsNode(name=node_name, node_context=self.__ros_context)
         self.__ros_executor.add_node(self.__ros_node)
 
         # Start spinning
@@ -59,7 +79,7 @@ class WaitForTopics:
         rclpy.shutdown(context=self.__ros_context)
 
     def __enter__(self):
-        self.wait()
+        assert self.wait()
         return self
 
     def __exit__(self, exep_type, exep_value, trace):
@@ -68,19 +88,17 @@ class WaitForTopics:
         self.shutdown()
 
 
-class MakeTestNode(Node):
-    """Mock node to be used for listening on topics."""
+class _WaitForTopicsNode(Node):
+    """Internal dummy node to be used for listening on topics."""
 
     def __init__(self, name='test_node', node_context=None):
         super().__init__(node_name=name, context=node_context)
         self.msg_event_object = Event()
-        self.logger = rclpy.logging.get_logger('wait_for_topics')
 
     def start_subscribers(self, topic_tuples):
         self.subscriber_list = []
-        self.expected_topics = [name for name, _ in topic_tuples]
-        self.expected_topics.sort()
-        self.received_topics = []
+        self.expected_topics = {name for name, _ in topic_tuples}
+        self.received_topics = set()
         self.topic_mutex = Lock()
 
         for topic_name, topic_type in topic_tuples:
@@ -99,9 +117,8 @@ class MakeTestNode(Node):
         def topic_callback(data):
             self.topic_mutex.acquire()
             if topic_name not in self.received_topics:
-                self.logger.info('Message received for ' + topic_name)
-                self.received_topics.append(topic_name)
-                self.received_topics.sort()
+                self.get_logger().debug('Message received for ' + topic_name)
+                self.received_topics.add(topic_name)
             if self.received_topics == self.expected_topics:
                 self.msg_event_object.set()
 
--- a/launch_testing_ros/launch_testing_ros/wait_for_topics.py
+++ b/launch_testing_ros/launch_testing_ros/wait_for_topics.py
@@ -104,7 +104,7 @@ class WaitForTopics:
 
 
 class _WaitForTopicsNode(Node):
-    """Internal dummy node to be used for listening on topics."""
+    """Internal node used for subscribing to a set of topics."""
 
     def __init__(self, name='test_node', node_context=None):
         super().__init__(node_name=name, context=node_context)
--- a/launch_testing_ros/launch_testing_ros/wait_for_topics.py
+++ b/launch_testing_ros/launch_testing_ros/wait_for_topics.py
@@ -1,16 +1,32 @@
+# Copyright 2021 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
 from threading import Event
+from threading import Lock
 from threading import Thread
 
 import rclpy
 from rclpy.node import Node
 
+
 class WaitForTopics:
-    def __init__(self, topic_name, topic_type):
+    """Wait to receive messages on supplied nodes."""
+
+    def __init__(self, topic_tuples):
         self.__ros_context = None
         self.__ros_node = None
-        self.topic_name = topic_name
-        self.topic_type = topic_type
-
+        self.topic_tuples = topic_tuples
         self.start()
 
     def start(self):
@@ -19,7 +35,7 @@ class WaitForTopics:
         self.__ros_node = MakeTestNode(name='test_node')
 
     def wait(self, timeout=5.0):
-        self.__ros_node.start_subscriber(self.topic_name, self.topic_type)
+        self.__ros_node.start_subscribers(self.topic_tuples)
         return self.__ros_node.msg_event_object.wait(timeout)
 
     def shutdown(self):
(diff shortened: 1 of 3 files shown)


launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+        self.__ros_context = rclpy.Context()
+        rclpy.init(context=self.__ros_context)
+        self.__ros_executor = SingleThreadedExecutor(context=self.__ros_context)
+        self.__ros_node = MakeTestNode(name='test_node', node_context=self.__ros_context)
+        self.__ros_executor.add_node(self.__ros_node)
+
+        # Start spinning
+        self.__running = True
+        self.__ros_spin_thread = Thread(target=self.spin_function)
+        self.__ros_spin_thread.start()
```
jacobperron, line 45: We can simplify this assuming we don't want to expose `start()` to users (which I think we don't):
```suggestion
        self.topic_tuples = topic_tuples
        self.timeout = timeout
        self.__ros_context = rclpy.Context()
        rclpy.init(context=self.__ros_context)
        self.__ros_executor = SingleThreadedExecutor(context=self.__ros_context)
        self.__ros_node = MakeTestNode(name='test_node', node_context=self.__ros_context)
        self.__ros_executor.add_node(self.__ros_node)

        # Start spinning
        self.__running = True
        self.__ros_spin_thread = Thread(target=self.spin_function)
        self.__ros_spin_thread.start()
```

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+        self.__ros_executor = None
+        self.topic_tuples = topic_tuples
+        self.timeout = timeout
+        self.start()
+
+    def start(self):
+        self.__ros_context = rclpy.Context()
+        rclpy.init(context=self.__ros_context)
+        self.__ros_executor = SingleThreadedExecutor(context=self.__ros_context)
+        self.__ros_node = MakeTestNode(name='test_node', node_context=self.__ros_context)
```
jacobperron, line 39: It's possible we may want to create multiple instances of class in a test, so I think we should ensure the node name is unique by adding some random content to the string. It's also probably nice to hide the node from tooling by prefixing it with an underscore.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+        self.wait()
+        return self
+
+    def __exit__(self, exep_type, exep_value, trace):
+        if exep_type is not None:
+            raise Exception('Exception occured, value: ', exep_value)
+        self.shutdown()
+
+
+class MakeTestNode(Node):
```
jacobperron, line 71: nitpick: Avoid the generic name "MakeTestNode". I think something more descriptive would be beneficial, like "_WaitForTopicsNode". Note, I would prefix the name with some underscores to indicate this is an implementation detail.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+    """Mock node to be used for listening on topics."""
```
jacobperron, line 72: nitpick: Avoid the word "mock" here as it has certain connotations related to testing. E.g. it implies this is an imitation of a Node, however this is a real, fully functional Node.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def __init__(self, name='test_node', node_context=None):
+        super().__init__(node_name=name, context=node_context)
+        self.msg_event_object = Event()
+        self.logger = rclpy.logging.get_logger('wait_for_topics')
+
+    def start_subscribers(self, topic_tuples):
+        self.subscriber_list = []
+        self.expected_topics = [name for name, _ in topic_tuples]
+        self.expected_topics.sort()
```
jacobperron, line 82: Sorting doesn't seem necessary. Is it?

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+        self.shutdown()
+
+
+class MakeTestNode(Node):
+    """Mock node to be used for listening on topics."""
+
+    def __init__(self, name='test_node', node_context=None):
+        super().__init__(node_name=name, context=node_context)
+        self.msg_event_object = Event()
+        self.logger = rclpy.logging.get_logger('wait_for_topics')
```
jacobperron, line 77: We can use the Node's logger instead
```suggestion
        self.logger = self.get_logger()
```

or just replace any instance of `self.logger` with `self.get_logger()`.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+                    topic_name,
+                    self.callback_template(topic_name),
+                    10
+                )
+            )
+
+    def callback_template(self, topic_name):
+
+        def topic_callback(data):
+            self.topic_mutex.acquire()
```
jacobperron, line 100: Correct me if I'm wrong, but I don't think we need a mutex since we are using the SingleThreadedExecutor.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+        self.received_topics = []
+        self.topic_mutex = Lock()
+
+        for topic_name, topic_type in topic_tuples:
+            # Create a subscriber
+            self.subscriber_list.append(
+                self.create_subscription(
+                    topic_type,
+                    topic_name,
+                    self.callback_template(topic_name),
```
jacobperron, line 92: Alternatively, we can use [functools.partial](https://docs.python.org/3/library/functools.html#functools.partial).

```python
functools.partial(self.topic_callback, topic_name)
```

and then the callback would look like:

```python
def topic_callback(self, topic_name, data):
    ...
```

I don't think there's anything wrong with this, so feel free to leave it as-is.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+                    10
+                )
+            )
+
+    def callback_template(self, topic_name):
+
+        def topic_callback(data):
+            self.topic_mutex.acquire()
+            if topic_name not in self.received_topics:
+                self.logger.info('Message received for ' + topic_name)
```
jacobperron, line 102: super minor: IMO, this logging might be too verbose to add to the test output. I would reduce this to `debug` verbosity personally.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,110 @@
...
+from threading import Lock
+from threading import Thread
+
+import rclpy
+from rclpy.executors import SingleThreadedExecutor
+from rclpy.node import Node
+
+
+class WaitForTopics:
+    """Wait to receive messages on supplied nodes."""
```
jacobperron, line 25: It would be nice to show some example usage in the docstring, e.g. using in a `with` context and calling `wait()` / `shutdown()` manually.

jacobperron (commented):
I think this is a good start. I've left some feedback below.

Also, I'm not sure that creating a separate node class adds much value; I would have just put all the logic inside WaitForTopics. I'm okay with it either way though.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
jacobperron: ah, nevermind. I see that you need to in order to compare the two lists of topics.

I would use a set instead of a list to avoid having to sort.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Added underscore prefix and random characters at the end.

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Changed to use ``set()``

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Changed to ``debug`` verbosity

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Maybe ``Dummy node to be used for listening on topics`` ?

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Done

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Removed the mutex

launch_testing_ros/launch_testing_ros/wait_for_topics.py
jacobperron: Sure, or even "Node used for subscribing to a set of topics."

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,122 @@
...
+            ''.join(random.choices(string.ascii_uppercase + string.digits, k=10))
+        self.__ros_node = _WaitForTopicsNode(name=node_name, node_context=self.__ros_context)
+        self.__ros_executor.add_node(self.__ros_node)
+
+        # Start spinning
+        self.__running = True
+        self.__ros_spin_thread = Thread(target=self.spin_function)
+        self.__ros_spin_thread.start()
+
+    def spin_function(self):
```
jacobperron, line 66: If we don't expect the user to call this themselves, prefix it with an underscore.
```suggestion
    def _spin_function(self):
```

launch_testing_ros/launch_testing_ros/wait_for_topics.py
```diff
@@ -0,0 +1,122 @@
...
+        return self.__ros_node.msg_event_object.wait(self.timeout)
+
+    def shutdown(self):
+        self.__running = False
+        self.__ros_spin_thread.join()
+        self.__ros_node.destroy_node()
+        rclpy.shutdown(context=self.__ros_context)
+
+    def __enter__(self):
+        assert self.wait()
```
jacobperron, line 81: Asserting seems a little harsh, I recommend raising an exception instead. This gives the user the opportunity to catch the exception and do something else.

jacobperron (commented):
LGTM, two more comments below

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Added RuntimeError with info on which topics did not publish

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Done

launch_testing_ros/launch_testing_ros/wait_for_topics.py
adityapande-1995: Done

jacobperron (approved):

adityapande-1995:
CI : 
build : ``--packages-above-and-dependencies launch_testing_ros``
test : ``--packages-above launch_testing_ros``

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/15375/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10082/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/13059/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/15578/)

launch_testing_ros/test/examples/wait_for_topic_launch_test.py
```diff
@@ -0,0 +1,91 @@
...
+        # Add a topic that will never have anything published on it
+        topic_list.append(('invalid_topic', String))
+        expected_topics = {'chatter_' + str(i) for i in range(count)}
+
+        # Method 1
+        wait_for_node_object_1 = WaitForTopics(topic_list, timeout=2.0)
+        with pytest.raises(RuntimeError):
+            with wait_for_node_object_1:
+                assert wait_for_node_object_1.topics_received() == expected_topics
+                assert wait_for_node_object_1.topics_not_received() == {'invalid_topic'}
```
jacobperron, line 84: Since we expect the with-clause to raise, the assert on topics (not) received should not be executed:
```suggestion
        with pytest.raises(RuntimeError):
            with WaitForTopics(topic_list, timeout=2.0):
                pass
```

jacobperron (approved):
A couple suggestions for the test code. LGTM

launch_testing_ros/test/examples/wait_for_topic_launch_test.py
```diff
@@ -0,0 +1,91 @@
...
+class TestFixture(unittest.TestCase):
+
+    def test_topics_successful(self, count):
+        """All the supplied topics should be read successfully."""
+        topic_list = [('chatter_' + str(i), String) for i in range(count)]
+        expected_topics = {'chatter_' + str(i) for i in range(count)}
+
+        # Method 1 : Using the magic methods and 'with' keyword
+        wait_for_node_object_1 = WaitForTopics(topic_list, timeout=2.0)
+        with wait_for_node_object_1:
```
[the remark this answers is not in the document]
adityapande-1995: Done

jacobperron (approved):

clalancette:
It's weird that this caused a regression, particularly because CI seems to have been run properly and came back green.  But let's see if the revert PR fixes it.

Blast545:
I can't replicate the error in CI. I'll investigate a bit deeper before reaching any conclusion.

PR #276, opened 2021-10-07
Revert "launch testing : Wait for topics to publish (#274)"

Blast545:
This reverts commit bbcc0cc8c693e5c3009a6e7e269b45518c209fbb.

PR #277, opened 2021-10-08
Don't import WaitForTopics so that rclpy isn't imported

sloretz:
Might fix or workaround ros2/rclpy#831


clalancette:
FYI, we are likely to revert this in #276 .  If you think there is something else that can be done quickly, you may want to comment there :).

jacobperron:
Replaced by #278

PR #278, opened 2021-10-11
Move pytest entrypoints to own module

sloretz:
Alternative to #277 
Attempts to fix ros2/rclpy#831

This moves the pytest entrypoints to a separate module, and makes them import anything ROS specific as late as possible. This should prevent packages like `ament_package` importing downstream packages like rclpy.

### DIFF
--- a/launch_testing_ros/launch_testing_ros/pytest/hooks.py
+++ b/launch_testing_ros/launch_testing_ros/pytest/hooks.py
@@ -33,16 +33,3 @@ class LaunchROSTestModule(LaunchTestModule):
 
     def makeitem(self, *args, **kwargs):
         return LaunchROSTestItem.from_parent(*args, **kwargs)
-
-
-def pytest_launch_collect_makemodule(path, parent, entrypoint):
-    marks = getattr(entrypoint, 'pytestmark', [])
-    if marks and any(m.name == 'rostest' for m in marks):
-        return LaunchROSTestModule.from_parent(parent=parent, fspath=path)
-
-
-def pytest_configure(config):
-    config.addinivalue_line(
-        'markers',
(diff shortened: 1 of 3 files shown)


sloretz:
This branch is on the commit just before #276, so windows debug CI should show if the issue is fixed or not.

Windows debug: build / test all https://ci.ros2.org/job/ci_windows/15612/

sloretz:
The second commit also reverts #276 for testing the approach purposes

Windows debug CI all: [Build Status](https://ci.ros2.org/job/ci_windows/15613/)

jacobperron (approved):
Besides the test failure, the change LGTM.

Do you know if keeping isolating pytest hooks like this is a best practice documented anywhere?
Also, should we consider doing something similar for [launch_testing](https://github.com/ros2/launch/blob/ec481fb0a5d9882fa7ca1bc073d8c4f91278bd38/launch_testing/launch_testing/pytest/hooks.py)?

ivanpauno (approved):
LGTM!

The only weird thing is that the file called `hooks.py` does not contain the definition of any pytest hook now.
(IIRC `pytest_launch_collect_makemodule` and `pytest_configure` are called "hooks" in pytest parlance)

jacobperron:
I'm not sure why we are seeing those qt_dotgraph test failures. That package does not transitively depend on launch_ros...
I was not able to reproduce the failures locally and haven't seen them appear in any other job. Re-triggering to see if it's still an issue:
* Linux: [Build Status](https://ci.ros2.org/job/ci_linux/15588/)

ivanpauno:
macOS build failure seems unrelated, running it again:

* macOS [Build Status](https://ci.ros2.org/job/ci_osx/13273/)

PR #300, opened 2022-03-02
Handle empty strings when evaluating parameters

jacobperron:
If a substitution evaluates to the empty string, then yaml.safe_load() will return None, resulting
in an exception being raised. Instead, this commit allows for the empty string by first wrapping
it in an extra set of quotes before passing to yaml.safe_load().

In response to a reported regression https://github.com/ros2/launch_ros/pull/289#discussion_r818070166
if this is accepted it should be backported to Galactic and Foxy.

### DIFF
--- a/launch_ros/launch_ros/utilities/evaluate_parameters.py
+++ b/launch_ros/launch_ros/utilities/evaluate_parameters.py
@@ -71,6 +71,10 @@ def evaluate_parameter_dict(
                 # Value is a list of substitutions, so perform them to make a string
                 evaluated_value = perform_substitutions(context, list(value))
 
+                # Handle special case where yaml.safe_load will return None given an empty string
+                if len(evaluated_value) == 0:
+                    evaluated_value = "''"
+
                 try:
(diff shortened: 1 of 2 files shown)


adityapande-1995:

jacobperron:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16281/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10901/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16648/)

clalancette:
Now that we've gotten a more-or-less full rebuild on the buildfarm, let's see if the PR jobs are any happier:

adityapande-1995 (approved):
Looks good ! Not sure what to do about the Rpr CI

jacobperron:
I think we can ignore the Rpr failure and rely on ci.ros2.org for now, since it seems to be a specific issue with the build.ros2.org binaries.

PR #303, opened 2022-03-14
Use correct namespace when evaluating parameter files for composable nodes

jacobperron:
* When resolving parameters from a file, use the full node namespace, combined with any base
  namespace provided by launch (e.g. from PushRosNamespace).
* Handle edge case in 'to_parameters_list', when namespace equals '/'.
* Fix bug in test fixture returning invalid node names that contain consecutive forward slashes.
* Add tests to cover cases for loading parameters from file when namespace is provided by
  PushRosNamespace actions.

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -293,16 +293,17 @@ def get_composable_node_load_request(
     if parameters:
         request.parameters = [
             param.to_parameter_msg() for param in to_parameters_list(
-                context, request.node_name, expanded_ns,
+                context, request.node_name, combined_ns,
                 evaluate_parameters(
                     context, parameters
                 )
             )
         ]
+
     if composable_node_description.extra_arguments is not None:
         request.extra_arguments = [
             param.to_parameter_msg() for param in to_parameters_list(
-                context, request.node_name, expanded_ns,
+                context, request.node_name, combined_ns,
                 evaluate_parameters(
                     context, composable_node_description.extra_arguments
                 )
--- a/launch_ros/launch_ros/utilities/to_parameters_list.py
+++ b/launch_ros/launch_ros/utilities/to_parameters_list.py
@@ -75,7 +75,7 @@ def to_parameters_list(
     """
     parameters = []  # type: List[rclpy.parameter.Parameter]
     node_name = node_name.lstrip('/')
-    if namespace:
+    if namespace and namespace != '/':
         namespace = namespace.strip('/')
         node_name = f'{namespace}/{node_name}'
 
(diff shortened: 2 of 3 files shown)


jacobperron:
Testing packages above launch_ros: 
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16319/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/10939/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16689/)

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -293,16 +293,17 @@ def get_composable_node_load_request(
     if parameters:
         request.parameters = [
             param.to_parameter_msg() for param in to_parameters_list(
-                context, request.node_name, expanded_ns,
+                context, request.node_name, combined_ns,
```
jacobperron, line 296: This is the main bug.

launch_ros/launch_ros/utilities/to_parameters_list.py
```diff
@@ -75,7 +75,7 @@ def to_parameters_list(
     """
     parameters = []  # type: List[rclpy.parameter.Parameter]
     node_name = node_name.lstrip('/')
-    if namespace:
+    if namespace and namespace != '/':
```
jacobperron, line 78: This handles an edge case where `namespace == '/'`, which would result in a node name with a forward slash, e.g. `/test_node_name`. But, the rest of this function expects the slashes to be stripped.

test_launch_ros/test/test_launch_ros/actions/test_load_composable_nodes.py
```diff
@@ -71,7 +71,10 @@ def __init__(self):
     def load_node_callback(self, request, response):
         self.requests.append(request)
         response.success = True
-        response.full_node_name = f'{request.node_namespace}/{request.node_name}'
+        if request.node_namespace == '/':
+            response.full_node_name = f'/{request.node_name}'
+        else:
+            response.full_node_name = f'{request.node_namespace}/{request.node_name}'
```
jacobperron, line 77: This change is to avoid illegal node names (containing two forward slashes, e.g. `//test_node_name`).

adityapande-1995 (commented):
IIUC, the changes in ``test_launch_ros`` avoid illegal node names just for the test cases. What would happen if a user tries to pass illegal node names to launch_ros (outside test cases that is) ?

jacobperron:
> IIUC, the changes in test_launch_ros avoid illegal node names just for the test cases. What would happen if a user tries to pass illegal node names to launch_ros (other than the test cases that is) ?

Specifically, the change avoids a component container from reporting an illegal node name. If a real-world container does this, then I guess it will result in similar issues loading parameters (silently ignore parameters that don't match the node name). For example, if a container loads a node and reports an invalid name "//foo", then the following parameter will, silently, not be passed to the node:

```
/foo
  ros__parameters:
    bar: 1
```

_Maybe_ the following parameters file would work, but I didn't test it:

```
//foo
  ros__parameters:
    bar: 1
```

IMO, we shouldn't go out of our way to support unexpected node naming conventions. We could go so far as to validate node names reported by containers and emit a warning that there's a problem with the container being launched.

ivanpauno (approved):
LGTM!

PR #305, opened 2022-03-30
Increase test time tolerance

Blast545:
Addressing test regression on `nightly_win_deb` jobs.
https://ci.ros2.org/view/nightly/job/nightly_win_deb/2306/testReport/junit/test_launch_ros.test.test_launch_ros.actions/test_ros_timer/test_timer_action_sanity_check/

The time tolerance is currently set to 0.1 and test is failing because test_time = 1.12 > 1.1

AFAIK these type of timers make a guarantee to be triggered "at least after x amount of time". I rewrote a bit the test to reflect better this scenario. Please correct me if Iḿ wrong.

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
@@ -115,7 +115,7 @@ def test_timer_action_sanity_check():
     assert shutdown_reasons[0].reason == 'One second timeout'
 
     # Verify that at least 1 sec has passed between start of test and timeout
-    tolerance = 0.25
+    tolerance = 1.0
     assert (end_time - start_time) > 1
     assert (end_time - start_time) < 1 + tolerance
 
--- a/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
+++ b/test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
@@ -114,9 +114,9 @@ def test_timer_action_sanity_check():
     assert 0 == ls.run()
     assert shutdown_reasons[0].reason == 'One second timeout'
 
-    # Verify that 1 sec has passed between start of test and timeout
-    tolerance = 0.1
-    assert (end_time - start_time) > 1 - tolerance
+    # Verify that at least 1 sec has passed between start of test and timeout
+    tolerance = 0.25
+    assert (end_time - start_time) > 1
     assert (end_time - start_time) < 1 + tolerance
 
 


Blast545:
Exploratory CI:
CI windows without fix
[Build Status](https://ci.ros2.org/job/ci_windows/16842/)

CI windows with fix
[Build Status](https://ci.ros2.org/job/ci_windows/16843/)

test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
```diff
@@ -114,9 +114,9 @@ def test_timer_action_sanity_check():
     assert 0 == ls.run()
     assert shutdown_reasons[0].reason == 'One second timeout'
 
-    # Verify that 1 sec has passed between start of test and timeout
-    tolerance = 0.1
-    assert (end_time - start_time) > 1 - tolerance
+    # Verify that at least 1 sec has passed between start of test and timeout
+    tolerance = 0.25
```
clalancette, line 118: Honestly, I'd just make this 1.0.  Even a full second here is not *always* going to succeed (particularly on Windows), but it should be a lot less flaky that way.

Otherwise this refactor makes sense to me.

Blast545: Yeah, I've never seen this error before, so I imagine just increasing the tolerance was a good fix.

I don't have a strong opinion, so I can remove the test if you think that'd be better long term speaking.

clalancette: > Yeah, I've never seen this error before, so I imagine just increasing the tolerance was a good fix.

I'd say let's start with a tolerance of 1.0, and the rest of what you've done here.  If the test continues to be a problem, we can remove it later on.

Blast545:
Full CI after increasing tolerance: 

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16475/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11065/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/16846/)

test_launch_ros/test/test_launch_ros/actions/test_ros_timer.py
Blast545: Sure, added with: 1f02915

clalancette (approved):

clalancette:
With green CI, I'm going to go ahead and merge this.  Thanks for the fix @Blast545 !

PR #297, opened 2022-02-25
Add parameter substitution

kenji-miyake:
It is useful if we can use the parameter substitution.

```xml
<launch>
  <set_parameter name="name" value="value" />
  <let name="result" value="$(param name)" />
</launch>
```

```py
SetParameter('name', 'value').execute(context)
value = Parameter('name').perform(context)
```

Also, since I got this error while I'm testing this, fixed it.

```py
    @property
    def name(self) -> ParameterName:
        """Getter for name."""
>       return self.__param_dict.keys()[0]
E       TypeError: 'dict_keys' object is not subscriptable

build/launch_ros/launch_ros/actions/set_parameter.py:79: TypeError
```

PR #313, opened 2022-04-23
Support default values in parameter substitution

kenji-miyake:
I think I should have supported default values in https://github.com/ros2/launch_ros/pull/297 to be consistent with the behavior of `LaunchConfiguration`.

So I've fixed the code to be similar to https://github.com/ros2/launch/blob/d2935d6fd85636ace62699eec106cd56f5e9c6a4/launch/launch/substitutions/launch_configuration.py.

### DIFF
--- a/launch_ros/launch_ros/substitutions/parameter.py
+++ b/launch_ros/launch_ros/substitutions/parameter.py
@@ -14,9 +14,13 @@
 
 """Module for the Parameter substitution."""
 
+import collections.abc
+from typing import Any
 from typing import Iterable
 from typing import List
+from typing import Optional
 from typing import Text
+from typing import Union
 
 from launch.frontend import expose_substitution
 from launch.launch_context import LaunchContext
@@ -38,17 +42,41 @@ class Parameter(Substitution):
     def __init__(
         self,
         name: SomeSubstitutionsType,
+        *,
+        default: Optional[Union[Any, Iterable[Any]]] = None
     ) -> None:
         """Create a Parameter substitution."""
         super().__init__()
         self.__name = normalize_to_list_of_substitutions(name)
+        if default is None:
+            self.__default = default
+        else:
+            # convert any items in default that are not a Substitution or str to a str
+            str_normalized_default = []  # type: List[Union[Text, Substitution]]
+            definitely_iterable_default = ((),)  # type: Iterable[Any]
+            if isinstance(default, collections.abc.Iterable):
+                definitely_iterable_default = default
+            else:
+                definitely_iterable_default = (default,)
+            for item in definitely_iterable_default:
+                if isinstance(item, (str, Substitution)):
+                    str_normalized_default.append(item)
+                else:
+                    str_normalized_default.append(str(item))
+            # use normalize_to_list_of_substitutions to convert str to TextSubstitution's too
+            self.__default = \
+                normalize_to_list_of_substitutions(
+                    str_normalized_default)  # type: List[Substitution]
 
     @classmethod
     def parse(cls, data: Iterable[SomeSubstitutionsType]):
         """Parse a Parameter substitution."""
-        if not data or len(data) != 1:
-            raise AttributeError('param substitutions expect 1 argument')
-        kwargs = {'name': data[0]}
+        if len(data) < 1 or len(data) > 2:
+            raise TypeError('param substitution expects 1 or 2 arguments')
+        kwargs = {}
+        kwargs['name'] = data[0]
+        if len(data) == 2:
+            kwargs['default'] = data[1]
         return cls, kwargs
 
     @property
@@ -69,4 +97,8 @@ class Parameter(Substitution):
             if isinstance(param, tuple):
                 if param[0] == name:
                     return param[1]
-        raise SubstitutionFailure("parameter '{}' not found".format(name))
+
+        if self.__default is None:
+            raise SubstitutionFailure("parameter '{}' not found".format(name))
+        else:
+            return perform_substitutions(context, self.__default)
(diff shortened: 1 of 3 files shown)


kenji-miyake:
@jacobperron Hello, could you review this, please? :pray: 

Also, I know the code freeze deadline is over, but considering this feature isn't used by anyone, is there any possibility to backport this to Humble?
https://docs.ros.org/en/rolling/Releases/Release-Humble-Hawksbill.html#release-timeline

Thank you in advance. :bow:

launch_ros/launch_ros/substitutions/parameter.py
```diff
@@ -38,17 +42,41 @@ class Parameter(Substitution):
...
+                definitely_iterable_default = (default,)
+            for item in definitely_iterable_default:
+                if isinstance(item, (str, Substitution)):
+                    str_normalized_default.append(item)
+                else:
+                    str_normalized_default.append(str(item))
+            # use normalize_to_list_of_substitutions to convert str to TextSubstitution's too
+            self.__default = \
+                normalize_to_list_of_substitutions(
+                    str_normalized_default)  # type: List[Substitution]
```
jacobperron, line 69: Is this type annotation useful since `self.__default` may be `None` in some cases? Anyways, [`normalize_to_list_of_substitutions`](https://github.com/ros2/launch/blob/d2935d6fd85636ace62699eec106cd56f5e9c6a4/launch/launch/utilities/normalize_to_list_of_substitutions_impl.py#L26) already contains a return type annotation, so think the comment is not needed.

```suggestion
            self.__default = normalize_to_list_of_substitutions(str_normalized_default)
```

jacobperron (commented):
Could you also add a tests for valid and invalid parameter names with default values? Here:

https://github.com/ros2/launch_ros/blob/master/test_launch_ros/test/test_launch_ros/substitutions/test_parameter_substitution.py

Thanks!

jacobperron:
What is the use-case for default values? I don't think it's necessary to support this feature for the sake of being consistent with launch configurations (they are two different concepts).

kenji-miyake:
I'm sorry for my late reply. :bow: 

The use case is, for large launch systems, it's useful if we can override parameters.

```xml
// called launch
<launch>
  <node ...>
    <param name="foo" value="$(var global_param_foo bar)" />
  </node>
</launch>

// caller launch 1
<launch>
  <include file="{called launch}" />
</launch>
-> foo=bar

// caller launch 2
<launch>
  <set_parameter name="global_param_foo" value="override" />
  <include file="{called launch}" />
</launch>
-> foo=override
```

Or another case, we might want to change the behavior like this.

```
<launch>
  <set_parameter name="foo" value="bar" />

  <group unless="$(eval ' \'$(param x \'\')\'==\'\' ')" >
    <log message="defined" />
  </group>

  <group if="$(eval ' \'$(param x \'\')\'==\'\' ')" >
    <log message="not defined" />
  </group>
</launch>
```

Without this PR, it raises an error.

But they might be minor use cases and, as you say, not necessary.
Perhaps, it might be better to define something like the `has_param` substitution.

@jacobperron What do you think about this?

launch_ros/launch_ros/substitutions/parameter.py
kenji-miyake: Actually, I just copied it from `launch_configuration.py`.
Removed in 9e6478809d770e62cefd494a3317ea2168ec0ad6.

kenji-miyake:
> Could you also add tests for valid and invalid parameter names with default values? Here:

In e7e23e39cdaa0a36aae0768348c295a41c9ef203, I added a test case of `valid parameter name & default value`.

Regarding a invalid parameter name, I thought I've tested by `assert Parameter('name-invalid', default='default_value').perform(context) == 'default_value'`.
Did you mean something else? :thinking:

jacobperron:
In your examples, I don't see the difference with using launch configurations. E.g. this looks equivalent to your first example

```
// called launch
<launch>
  <arg name="global_param_foo" default="bar" />
  <node ...>
    <param name="foo" value="$(var global_param_foo)" />
  </node>
</launch>

// caller launch 1
<launch>
  <include file="{called launch}" />
</launch>
-> foo=bar

// caller launch 2
<launch>
  <let name="global_param_foo" value="override" />
  <include file="{called launch}" />
</launch>
-> foo=override
```

But, anyways, I think this is still okay to add to go along with the existing `SetParameterAction`.

jacobperron:
The tests look good :+1:

jacobperron:
Besides [this flake8 error](https://build.ros2.org/job/Rpr__launch_ros__ubuntu_jammy_amd64/53/testReport/junit/test_launch_ros.test/test_flake8/test_flake8/), the changes LGTM

kenji-miyake:
@jacobperron Thank you for your feedback. :smile: 

> In your examples, I don't see the difference with using launch configurations. E.g. this looks equivalent to your first example

I feel the issues with launch configurations are:
- Has more code lines.
- Can unintendedly affect other launch files.
  - (I think it can be avoided by the naming conventions.)

> But, anyways, I think this is still okay to add to go along with the existing SetParameter action.

I'll leave the decision to the maintainers!
This PR adds more flexibility, but I know it increases a bit of complexity.
I can wait to merge this until anyone will show another important use case.

jacobperron (approved):
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/16678/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11264/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17076/)

jacobperron:
I'm find with merging this, I just wanted to make sure I understood why it was wanted :)

Issue #291, opened 2021-11-29
Provide more meaningful output on error 


SteveMacenski:
#### Feature description

When launch cannot find a package, a launch file does not exist, a node doesn't exist, or generally speaking something goes wrong, the traceback provided really doesn't help in debugging the issue. I would really like these to be caught and categorized so that users receive a meaningful error message back about what they're doing wrong to be able to fix, ideally with the offending line.  

I'm finding it quite difficult when building new launch files and errors occur to figure out what's wrong. The terminal gets filled with internal launch tracebacks that infrequently help in diagnosing the underlying issue. Especially for common errors, we should have common error messages that tells people why it failed to execute and suppress the traceback. Surely this is an uphill battle to cover all situations, but if we simply cover the top 10, I think that would be a huge quality of life improvement

- Package doesn't exist on system
- Launch file or node doesn't exist on system
- Python syntax errors
- Field given to a `launch` API is invalid

SteveMacenski:
@hidmic I know this is probably low priority, but is this on anyone's radar for an upcoming release. This is a big quality of life issue in trying to debug modest robotic systems

jacobperron:
I'm not aware of anyone planning to tackle this issue, though I'm happy to review proposals if anyone in the community would like to work on it.

IIUC, errors are communicated via exceptions that [launch catches and logs](https://github.com/ros2/launch/blob/074cd2903ddccd61bce8f40a0f58da0b7c200481/launch/launch/launch_service.py#L344-L347). I guess there are situations where the errors are not being caught and we're getting undesired tracebacks.

For instance, if I launch a node action for a package that doesn't exist, the message looks _okay_ to me:

```
jacob@warner ~/tmp> ros2 launch talker.launch.py        
[INFO] [launch]: All log files can be found below /home/user/.ros/log/2022-05-31-14-41-11-720542-myhost-19000
[INFO] [launch]: Default logging verbosity is set to INFO
[ERROR] [launch]: Caught exception in launch (see debug for traceback): "package 'foo' not found, searching: ['/opt/ros/rolling']"
```
Of course it  would be better if it gave us a file and line number, but that might be something that needs to be added upstream in launch (I'm not sure). If there are particular messages that are cryptic, or a category of error that is causing a traceback, we can probably improve those on a case-by-case basis.

SteveMacenski:
An example is syntactical errors, they result in not super helpful things like below, which in no place give you the line number of the offending object

```
[...]

PR #320, opened 2022-06-27
Improve evaluate_paramenter_dict exceptions error message

FelipeGdM:
This PR updates the `evaluate_paramenter_dict()` possible exceptions adding the name of the parameter that caused the exception, thus helping find the root of the problem when it occurs

For example, if a `NoneType` argument is passed to the function, the thrown message would be

```
TypeError: Allowed value types are bytes, bool, int, float, str, Sequence[bool], Sequence[int], 
Sequence[float], Sequence[str]. Got <class 'NoneType'>. If the parameter is meant to be 
a string, try wrapping it in launch_ros.parameter_descriptions.ParameterValue(value, value_type=str)
```

Within a large project, with dozens of launch arguments, it is not trivial to discover which argument received the invalid value. As the name of the argument being evaluated is already available in the scope of the exception, adding it to the error message would be really helpful when dealing with this problem.

As discussed in some ROS 2 tickets (https://github.com/ros2/launch_ros/issues/291, https://github.com/ros2/launch/issues/412), one of the main issues of the launch system is the lack of helpful information when something goes wrong, this PR is a small step in the direction of solving this issue

### DIFF
--- a/launch_ros/launch_ros/utilities/evaluate_parameters.py
+++ b/launch_ros/launch_ros/utilities/evaluate_parameters.py
@@ -93,16 +93,20 @@ def evaluate_parameter_dict(
                     if not check_sequence_type_is_allowed(yaml_evaluated_value):
                         raise TypeError(
                             'Expected a non-empty sequence, with items of uniform type. '
-                            'Allowed sequence item types are bool, int, float, str.'
+                            'Allowed sequence item types are bool, int, float, str. '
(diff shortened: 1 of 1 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17782/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12327/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18425/)

PR #325, opened 2022-09-23
add SetROSLogDir action

wjwwood:
This action can be used to change the ROS log dir for any processes started after this action that contain ROS nodes.

It can be scoped when used in conjunction with the `launch.actions.GroupAction` action.

Requires https://github.com/ros2/launch/pull/652

PR #326, opened 2022-09-24
RosTimer -> ROSTimer and PushRosNamespace -> PushROSNamespace, to follow PEP8

wjwwood:
Follow up of https://github.com/ros2/launch_ros/pull/325#discussion_r979069795

I didn't actually see these being used anywhere else in the standard ROS repositories. But I did find a few places in the documentation: https://github.com/ros2/ros2_documentation/pull/3038

### DIFF
--- a/launch_ros/launch_ros/actions/__init__.py
+++ b/launch_ros/launch_ros/actions/__init__.py
@@ -18,8 +18,10 @@ from .composable_node_container import ComposableNodeContainer
 from .lifecycle_node import LifecycleNode
 from .load_composable_nodes import LoadComposableNodes
 from .node import Node
-from .push_ros_namespace import PushRosNamespace
-from .ros_timer import RosTimer
+from .push_ros_namespace import PushROSNamespace
+from .push_ros_namespace import PushROSNamespace as PushRosNamespace
+from .ros_timer import ROSTimer
+from .ros_timer import ROSTimer as RosTimer
 from .set_parameter import SetParameter
 from .set_parameters_from_file import SetParametersFromFile
 from .set_remap import SetRemap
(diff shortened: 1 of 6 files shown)


wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17370/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/11905/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/17873/)

jacobperron (approved):

ivanpauno (commented):
Should we deprecate the old names to make the transition easier?
Do we have any linter complaining about this?

wjwwood:
I was going to deprecate in a separate pr, since we could back port this one, but probably only want to deprecate in rolling. I just haven't had time.

wjwwood:
> Do we have any linter complaining about this?

If you're asking about the way the class is named, it's not possible for a linter to know one way or the other... I guess you could say something specific like `*Ros[A-Z].*` should be `ROS` instead. But no we don't have that.

clalancette:
> I was going to deprecate in a separate pr, since we could back port this one, but probably only want to deprecate in rolling. I just haven't had time.

Wait, I'm confused.  If we were to backport this, wouldn't we be breaking compatibility in old distributions?  That is, anything that called `PushRosNamespace` before would now break, right?

wjwwood:
This pr doesn't break API, it just adds new API (well rename and add an alias to the old name). No downstream code is required to change yet.

wjwwood:
I was thinking we could back port it as a convenience for downstream users that want to use the new class name in all branches (or have a single branch for Humble and Rolling and more).

ivanpauno (approved):

PR #327, opened 2022-10-04
Expose lifecycle_node

### LINKED ISSUE #236: Support LifecycleNode and EventHandler in frontend
#### Feature description

It's useful to support LifecycleNode in frontend(xml) as well.

As for LifecycleNode itself, if we only use simple launch files as below, it can be done by just exposing like this https://github.com/kenji-miyake/launch_ros/commit/8daecd810d65ebdea6df32ab145137faa343a80d.

```xml
<launch>
  <lifecycle_node pkg="lifecycle" exec="lifecycle_talker" name="lc_talker" namespace=""/>
  <node pkg="lifecycle" exec="lifecycle_listener" />
  <node pkg="lifecycle" exec="lifecycle_service_client" />
</launch>
```

However, I believe it's not enough useful because we usually would like to write state transitions as well like [this](https://github.com/autowarefoundation/ros2_socketcan/blob/a1f6cbf65e899deab01b0ccbc9084872580f9ae6/launch/socket_can_receiver.launch.py).

So I think it's better to support RegisterEventHandler and EventHandlers.

I think there are at least two problems.

1. How to expose EventHandlers?
Is it okay to add `parse_event_handler` and `instantiate_event_handler` as Action class does?

2. How to refer Python Object from frontend?
For example, `OnStateTransition` requires `LifecycleNode` object but I think it can't be given in frontend easily.
Probably this is what @ivanpauno mentioned here https://answers.ros.org/question/333404/launch-composed-nodes-using-the-launch-xml-frontend/, right?
> The topic of having references (i.e.: being able to identify) another action in the XML/YAML frontend is an interesting feature currently not supported. It would be useful, for example, to support event handlers.

So I guess some mechanism for finding Object from all entities.

Does anybody have any idea about this?
If somebody guides me, I'll try to work on this task!

esteve:
Original author: @kenji-miyake

Closes #236

### DIFF
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -23,6 +23,9 @@ from typing import Optional
 import launch
 from launch import SomeSubstitutionsType
 from launch.action import Action
+from launch.frontend import Entity
+from launch.frontend import expose_action
+from launch.frontend import Parser
 import launch.logging
 
 import lifecycle_msgs.msg
@@ -35,6 +38,7 @@ from ..events.lifecycle import StateTransition
 from ..ros_adapters import get_ros_node
 
 
+@expose_action('lifecycle_node')
 class LifecycleNode(Node):
     """Action that executes a ROS lifecycle node."""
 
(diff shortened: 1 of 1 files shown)


kenji-miyake:
@esteve Thank you for your work!
But actually, since it's still a limited part, I think it won't "close" the issue.

esteve:
@kenji-miyake thanks. I created https://github.com/ros2/launch/issues/658 to address event handlers

ivanpauno (approved):
LGTM!
It would be nice to add some tests before merging.

If we only expose this, and not event handlers, there's no much advantage of using `lifecycle_node` over `node` in XML launch files.
The only advantage is that the `StateTransition` and `ChangeState` events will be emited.
So if you include a python launch file from the XML launch file, you can take advantages of those using event handlers.
Though that seems fine IMO

ivanpauno:
@esteve @kenji-miyake could you add a test case for this?
Thanks!

SuperJappie08:
@ivanpauno @esteve @kenji-miyake, I could add some front-end tests, if you want?

SuperJappie08:
I've added the test in a new PR #482.

PR #334, opened 2022-11-29
Fix "imported but not used" linters

Blast545:
These linter errors started showing in the Galactic coverage jobs three days ago, reference job:
https://ci.ros2.org/job/nightly_linux_galactic_coverage/585/

As it was a quick one, I decided to open a PR to fix it. PTAL @Crola1702 @cottsay 

Signed-off-by: Jorge Perez <jjperez@ekumenlabs.com>

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -18,6 +18,7 @@ import threading
 
 from typing import List
 from typing import Optional
+from typing import Text # noqa
 from typing import Union
 
 import composition_interfaces.srv
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -18,7 +18,6 @@ import threading
 
 from typing import List
 from typing import Optional
-from typing import Text
 from typing import Union
 
 import composition_interfaces.srv
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -18,7 +18,7 @@ import threading
 
 from typing import List
 from typing import Optional
-from typing import Text # noqa
+from typing import Text  # noqa: F401
 from typing import Union
 
 import composition_interfaces.srv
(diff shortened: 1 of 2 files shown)


mjcarroll (approved):

Blast545:
I think it might be worth taking a look as to why this started appearing in the coverage jobs, and hasn't shown in the release/debug jobs. @Crola1702

Blast545:
Galactic up to `launch_ros`, Linux CI:

[Build Status](https://ci.ros2.org/job/ci_linux/17721/)

cottsay (changes requested):
Both of these are caused by typing annotations, and will appear with older Python interpreters. They should be marked with `#noqa`, not removed.

Here's where they're used:
https://github.com/ros2/launch_ros/blob/3c355ed49ba5e746d5e5ca454390bf6b6ac2fc2d/launch_ros/launch_ros/actions/load_composable_nodes.py#L82
https://github.com/ros2/launch_ros/blob/3c355ed49ba5e746d5e5ca454390bf6b6ac2fc2d/launch_ros/launch_ros/utilities/evaluate_parameters.py#L155

Crola1702:
Running CI with new change: [Build Status](https://ci.ros2.org/job/ci_linux/17726/)

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -18,7 +18,7 @@
 
 from typing import List
 from typing import Optional
-from typing import Text
+from typing import Text # noqa
```
cottsay, line 21:
```suggestion
from typing import Text  # noqa: F401
```

launch_ros/launch_ros/utilities/evaluate_parameters.py
```diff
@@ -17,7 +17,7 @@
 
 from collections.abc import Mapping
 from collections.abc import Sequence
-import pathlib
+import pathlib # noqa
```
cottsay, line 20:
```suggestion
import pathlib  # noqa: F401
```

cottsay (changes requested):
We can disable the specific check that's failing. Additionally, flake8 calls for two spaces before an inline comment, but the unqualified `#noqa` was actually squelching that violation as well.

Crola1702:
Running CI again with the latest commits:
[Build Status](https://ci.ros2.org/job/ci_linux/17759/)

cottsay (approved):

PR #311, opened 2022-04-14
Run condition for composable nodes

adityapande-1995:
Signed-off-by: Aditya <aditya050995@gmail.com>

## Summary 
This PR adds the ``condition`` parameter to composable nodes action. Resolves https://github.com/ros2/launch_ros/issues/294. This action is currently supported by ``ComposableNodeContainer`` but not ``ComposableNode`` itself.

## Example usage
One can use the following example file : 
```
import launch
import launch.actions
from launch.conditions import IfCondition, UnlessCondition
from launch.substitutions import LaunchConfiguration
import launch_ros.actions
from launch_ros.actions import ComposableNodeContainer
from launch_ros.descriptions import ComposableNode

def generate_launch_description():
    return launch.LaunchDescription([
        launch.actions.DeclareLaunchArgument(name='flag', default_value='False'),
        ComposableNodeContainer(
            name='my_container',
            namespace='',
            package='rclcpp_components',
            executable='component_container',

            composable_node_descriptions=[
                ComposableNode(
                    package='composition',
                    plugin='composition::Talker',
                    condition=IfCondition(LaunchConfiguration('flag')),
                    name='talker_1'),

                ComposableNode(
                    package='composition',
                    plugin='composition::Talker',
                    condition=UnlessCondition(LaunchConfiguration('flag')),
                    name='talker_2'),

                ComposableNode(
                    package='composition',
                    plugin='composition::Talker',
                    name='talker_3'),
            ],
            output='screen',
        ),
    ]) 
```
or alternatively as an xml, 

```
<launch>
        <arg name="flag" default="false"/>
        <node_container pkg="rclcpp_components" exec="component_container" name="my_container" namespace="">
                <composable_node pkg="composition" plugin="composition::Talker" name="talker_1" if="$(var flag)" />
                <composable_node pkg="composition" plugin="composition::Talker" name="talker_2" unless="$(var flag)" />
                <composable_node pkg="composition" plugin="composition::Talker" name="talker_3" />
        </node_container>
</launch>
```

Running this  as ``ros2 launch test.py  flag:=True`` will make it launch ``talker_1`` and ``talker_3``, and ``False`` will make it launch ``talker_2`` and ``talker_3``.

PR #341, opened 2022-12-08
Implement None check for ComposableNodeContainer

methylDragon:
This lets you pass in no list of nodes for ComposableNodeContainer so you can do this:

```
    container = ComposableNodeContainer(
        package='rclcpp_components',
        name=TEST_CONTAINER_NAME,
        namespace=TEST_CONTAINER_NAMESPACE
)
```

And so you can then pass that container to a subsequent LoadComposableNode call, if you wanted to.

Currently you would have to explicitly pass an empty list. But in that case, I'm not sure why the default argument in the constructor would be None as opposed to an empty list. (I think supporting both cases in this case would be preferable.)

### DIFF
--- a/launch_ros/launch_ros/actions/composable_node_container.py
+++ b/launch_ros/launch_ros/actions/composable_node_container.py
@@ -80,9 +80,11 @@ class ComposableNodeContainer(Node):
         """
         load_actions = None  # type: Optional[List[Action]]
         valid_composable_nodes = []
-        for node_object in self.__composable_node_descriptions:
-            if node_object.condition() is None or node_object.condition().evaluate(context):
-                valid_composable_nodes.append(node_object)
+        if self.__composable_node_descriptions:
+            for node_object in self.__composable_node_descriptions:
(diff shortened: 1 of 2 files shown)


methylDragon:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17787/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12333/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18431/)

adityapande-1995 (approved):
Looks good with green CI  !

HovorunB:
Can this be backported to humble please? As I see this comes from https://github.com/ros2/launch_ros/pull/311 which was [backported](https://github.com/ros2/launch_ros/pull/363) and for my case broke the compatibility since I didn't pass in an empty list for `composable_node_descriptions`. Backporting this PR would fix the compatibility, FYI: @adityapande-1995

clalancette:
@Mergifyio backport humble

PR #345, opened 2023-01-23
[ros2launch] Follow symlinks when walking.

mikepurvis:
Fixes an issue with using `ros2 launch` inside symlink-joined workspaces, such as when constructing them using systems like Nix or Bazel. Really these loop should just be rewritten to use `scandir` instead (which follows symlinks as the default behaviour), but this is a simpler change in the immediate.

Similar change to https://github.com/ros/ros/pull/225.

Credit for this find to @iwanders.

### DIFF
--- a/ros2launch/ros2launch/api/api.py
+++ b/ros2launch/ros2launch/api/api.py
@@ -46,7 +46,7 @@ def get_share_file_path_from_package(*, package_name, file_name):
     """
     package_share_directory = get_package_share_directory(package_name)
     matching_file_paths = []
-    for root, dirs, files in os.walk(package_share_directory):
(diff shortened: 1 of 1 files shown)


clalancette:
We discussed this, and we are a little concerned about the security implications of following symlinked directories.  This would seem to open users up to symlink swapping attacks.

That said, we also see how this could be useful.  So we think that we should leave the default to not follow symlinks, but add an option to `ros2launch` to allow following symlinks.  Does that make sense?

mikepurvis:
Hmm. We disagree with the reasoning on this— what is the scenario where an adversary can change a symlink on you and you aren't already owned in a dozen other ways?

In any case, if it's at least a config option then we can more trivially patch our version to change the default.

PR #358, opened 2023-03-06
Use SomeEntitiesType for type checking.

clalancette:
SomeActionsType has now been deprecated.

This was deprecated in https://github.com/ros2/launch/pull/679 , so update this code to match.  FYI @methylDragon @haudren-woven


haudren-woven:
Thank you !

methylDragon:
Merging with green CI but not passing RPr (since that is expected to fail since no release of launch has been made yet)

methylDragon (approved):

PR #365, opened 2023-06-14
cache lookup of importlib metadata in Node action

wjwwood:
The short version of this is that we use `importlib_metadata` to load extensions to the `Node` action via Python's `entry_point` pattern. This lets new packages extend `Node` without changing `launch_ros` itself. The problem is that we load these extensions every time a node needs them and that can take as long as 1 second on some machines. If you have a launch file with hundreds of nodes, like I did, then you get pauses of 30 seconds to a minute before the launch file starts launching things.

---

I used the profiler to find out that this `launch_ros.actions.Node.get_extensions()` was being called some 296 times in my launch file and each time it was taking ~0.4 seconds, leading to a delay of almost 30 seconds:

Screenshot_2023-06-15_14-58-26

After the change in this pr, the profiler shows what I experience on the command-line, which is much better behavior, taking almost no time to start launching things, since we only look this up once:

Screenshot_2023-06-15_15-00-10

Some potential issues with this pr:

- it will prevent extensions from changing after launch has started
- we could potentially cache more of the function, but this was the minimal change that made a large difference

For the first item, I don't think the rest of the code is written to assume this is possible in the first place, and I'm not sure what the use case for this would look like. I suppose you could have launch running, and while running install new python packages that add extensions, then cause launch to start a new node after this, and in that case it would not see new extensions. However, I don't know when you'd do that in practice. And I think this change benefits the existing users at the cost of a potential use case that I do not believe folks are using today, but I'm open to hear otherwise.

For the second, I don't think we necessarily need to cache more, based on what I see in the second profiler.

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -558,10 +558,15 @@ def instantiate_extension(
         extensions[extension_name] = extension_instance
     return extension_instance
 
+g_entry_points_impl = None
+
 
 def get_extensions(logger):
+    global g_entry_points_impl
     group_name = 'launch_ros.node_action'
-    entry_points_impl = importlib_metadata.entry_points()
+    if g_entry_points_impl is None:
+        g_entry_points_impl = importlib_metadata.entry_points()
+    entry_points_impl = g_entry_points_impl
     if hasattr(entry_points_impl, 'select'):
         groups = entry_points_impl.select(group=group_name)
     else:
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -558,6 +558,7 @@ def instantiate_extension(
         extensions[extension_name] = extension_instance
     return extension_instance
 
+
 g_entry_points_impl = None
 
 


adityapande-1995:
IIUC, this means that if one installs packages (that add extensions) while launch is still spinning up nodes, the nodes that start *after* the installation is done won't reflect the changes, right ? If yes, that seems like a weird use case to me, as when the installation completes is not predictable, and nor does launch guarantee the order in which nodes start.

So yes a +1 to this change, looks good !

wjwwood:
Right, the first time a node asks for the list of extensions it is cached, so if you install something new in parallel, it will be a race to see if it is included or not and with which nodes. I can't imagine a use case for that anyway, but I wanted to point it out.

adityapande-1995 (approved):
Looks good with green CI !

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18980/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13504/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/19724/)

fujitatomoya (approved):
> it will prevent extensions from changing after launch has started

i believe this is not the user expectation or assumption. lgtm.

PR #372, opened 2023-07-24
Remove create_future implementation.

clalancette:
This is available from Python directly, so just use that.


mjcarroll (approved):

clalancette:
CI for this is in https://github.com/ros2/launch/pull/723#issuecomment-1647903802

PR #356, opened 2023-03-01
`WaitForTopics`: let the user inject a callaback to be executed after starting the subscribers

LastStarDust:
I created this pull request as a proof of concept for the new feature discussed here: #348.
The ticket description is still correct and relevant, but for clarity's sake, let me summarize and expand on it here.

Let us assume there is a node ("NODE under test" in the diagram) which:

- takes a topic A as input
- does some calculation
- publishes another topic B as output

At the moment, there is no way to programmatically test the whole node end-to-end. For that, we need a way to:

1. inject a message of topic A into the node
2. wait for the calculation to be over
3. collect the resulting message of topic B coming out of the node

Currently, the `WaitForTopics` class provided by this package takes care of 2 and 3, but there is no mechanism to reliably accomplish 1, due to the asynchronous nature of ROS2 publishers and subscribers.

This PR is an attempt to enhance the `WaitForTopics` class so that it can take care of 1 as well, thus closing the loop. But it might well not be the only solution or the most effective. I am open to any kind of suggestion and guidance.

The main issue behind the implementation of such a feature is that in ROS2 the order of creation of publishers and subscribers does matter. They must either always be created in a specific order or be synchronized in some other way.

Below is a cartoon-style diagram illustrating what the end result should look like:

node-test-diagram

PR #375, opened 2023-08-16
Add async support for subscribing messages

Ryanf55:
This is an alternative to #356, used for discussion. It doesn't pass because the wait blocks.

### DIFF
--- a/launch_testing_ros/launch_testing_ros/repeater.py
+++ b/launch_testing_ros/launch_testing_ros/repeater.py
@@ -1,0 +1,59 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


Ryanf55:
@LastStarDust  FYI, here is another approach without using callbacks. Note how I also create a publisher in the test python code rather than calling the CI, which is more conducive to an application where you want to set specific fields in the test.

Ryanf55:
> @Ryanf55 Thank you for letting me know. I have some comments about the code, but before I invest any time in this I want to hear the opinion of the maintainers.

Looks like there are not really any active maintainers who have time to discuss this support. Are you still interested in getting better subscriber testing on `humble`?

LastStarDust:
@Ryanf55 for sure I am still interested in getter better subscriber testing on `humble`. 

@project owners
If there are no active maintainers or if they have no time for this feature, would you let us take care of this? In other words, would you give @Ryanf55 and @LastStarDust temporary maintainer role?

LastStarDust:
> > If there are no active maintainers or if they have no time for this feature, would you let us take care of this? In other words, would you give @Ryanf55 and @LastStarDust temporary maintainer role?
> 
> There are active maintainers of this repository. In general, this is very core ROS 2 stuff, so I don't think giving temporary maintainer role is suitable here.
> 
> That said, at the very least a new feature needs to target `rolling` first. Once this is targeted at `rolling`, I'll try to get a maintainer to look at it.

Fair enough. BTW [my PR](https://github.com/ros2/launch_ros/pull/356) is already targeting `rolling`.

Issue #398, opened 2024-06-13
feat: Adding a prefix option to node launch


wkaisertexas:
#### Feature description

Adding an extra parameter to the `Node` class to launch and executable with a specified prefix.

https://github.com/ros2/launch_ros/blob/8751ff4cd9060157831dcaeecd89611932aec608/launch_ros/launch_ros/actions/node.py#L112-L240

From what I can tell, this was a feature in `ros` which allowed a `prefix` to be specified which is launched ahead of the executable.

- [Ros Forums on Ros Launch and it use with GDB](https://wiki.ros.org/roslaunch/Tutorials/Roslaunch%20Nodes%20in%20Valgrind%20or%20GDB)

#### Implementation considerations

Potentially this feature could also be implemented with `ComposableNodes` API, however implementation and the use case there are a bit more vague.

Currently, I am using `Node(executable=f"{prefix} {executable}")` as a workaround, but this does not seem ideal.

clalancette:
I think we have this already in Launch; see http://docs.ros.org/en/jazzy/How-To-Guides/Getting-Backtraces-in-ROS-2.html#from-a-launch-file for an example.

Is that what you are asking about?

wkaisertexas:
Yes, I was a bit confused because it was lumped in as kwargs. Should there be a note about prefixing in the documentation for this method?

clalancette:
> Yes, I was a bit confused because it was lumped in as kwargs. Should there be a note about prefixing in the documentation for this method?

Yep, that would be great!  Feel free to open a PR doing that, we are happy to review it.

wkaisertexas:
@clalancette I am not sure if this is the correct way to do this with sphinx and the documentation system

PR #399, opened 2024-06-13
added notes on prefix

wkaisertexas:
`prefix` is part of kwargs, so it can be lost in the documentation

#398

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -196,6 +196,7 @@ class Node(ExecuteProcess):
         :param: remappings ordered list of 'to' and 'from' string pairs to be
             passed to the node as ROS remapping rules
         :param: ros_arguments list of ROS arguments for the node
+        :param: prefix command placed in front of executable on launch
         :param: arguments list of extra arguments for the node
         """
         if package is not None:


launch_ros/launch_ros/actions/node.py
```diff
@@ -196,6 +196,7 @@ def __init__(
         :param: remappings ordered list of 'to' and 'from' string pairs to be
             passed to the node as ROS remapping rules
         :param: ros_arguments list of ROS arguments for the node
+        :param: prefix command placed in front of executable on launch
```
wjwwood, line 199: We talked about this in the issue triage meeting, and the conclusion was that we should reference the base class that actually handles this argument, e.g. have a look at how Python does this for the subprocess API: https://docs.python.org/3/library/subprocess.html#subprocess.run

It's fine to give a brief description of the argument, but it should be separate from the other (actual) arguments and it should reference the place where it's actually documented, e.g. here: https://github.com/ros2/launch/blob/651f57451e4d0e93d055669764fcd588e2d9b1e6/launch/launch/actions/execute_process.py#L211-L216

It is indeed helpful to connect the dots here, but we don't want to duplicate too much documentation and risk them getting out of sync.

wjwwood (changes requested):

launch_ros/launch_ros/actions/node.py
wkaisertexas: Okay that makes total sense. Thank you for your reply

PR #409, opened 2024-08-30
Document names of frontend action arguments

### LINKED ISSUE #404: Document names of frontend action arguments
## Feature request

This applies to both `launch` and `launch_ros`, but it is more relevant for `launch_ros`, since it has the main user-facing/ROS-specific launch actions.

#### Feature description

When using the Python launch API directly, the names of the arguments are documented in the constructor docstring. For example, see `launch_ros.actions.Node.__init__()`. It makes sense to expect the user to rely on that, and this can easily be part of automatically-generated API docs (e.g., using `rosdoc2`).

However, when using the YAML or XML frontends, the names of the arguments might differ from the underlying Python class constructor. See: https://github.com/ros2/launch_ros/pull/253#issuecomment-2206862470. For example:

1. `launch_ros.actions.Node.__init__()` has `ros_arguments`: https://github.com/ros2/launch_ros/blob/13d6e1eb7c361f20706b709596496c16e0b5d1b1/launch_ros/launch_ros/actions/node.py#L128
2. but the corresponding frontend argument is `ros_args`: https://github.com/ros2/launch_ros/blob/13d6e1eb7c361f20706b709596496c16e0b5d1b1/launch_ros/launch_ros/actions/node.py#L313

This is not documented anywhere. While this might be part of some examples on docs.ros.org, it should be part of some sort of documentation.

Possible solutions:

1. Change frontend argument names to be the same as the corresponding Python class constructor arguments
    - I think it's just generally accepted, so I'll leave this out
2. Add some sort of Python annotation so that the names can be automatically parsed and included in the (`launch_ros`) API docs
    - Probably feasible, but uncertain; would require some work
3. Manually document the frontend argument names in the top-level Python class OR Python class constructor docstring
    - Feels weird to expect users to look at the Python action class docs to find the names of the frontend arguments
4. Manually document the frontend argument names of the `launch_ros` actions in a document under `launch_ros/doc/` and include in API docs generated by `rosdoc2`, similar to [what `launch` has](https://github.com/ros2/launch/tree/rolling/launch/doc/source)
    - API docs are accessible from [the index.ros.org page for the `launch_ros`](https://index.ros.org/p/launch_ros/github-ros2-launch_ros/#rolling): https://docs.ros.org/en/rolling/p/launch_ros/
    - This is the simplest solution
    - However, the docs would need to be updated (manually) if anything changes, and having this documentation a bit farther away from the actual code (compared to Python docstring) makes this slightly more of a hassle
    - The argument names concern the launch frontends (`launch_xml`/`launch_yaml`), but all the logic is in the underlying Python class (`launch_ros`), so it wouldn't make sense to document this in the actual frontend packages. One thing to point out: while the Python action class defines the names of the arguments for the frontends, the names are the same for all frontends, and the Python action implementation has no idea that there is a YAML frontend and an XML frontend. So these are all just "names for the frontends." The user-facing document could mention "the YAML and XML frontends" to make it simpler for users, but it doesn't have to.

christophebedard:
Resolves #404

This adds boilerplate for `launch_ros` docs along with a page to document the argument names for the frontend actions. The setup is similar to this page in the `launch` docs:

- https://github.com/ros2/launch/blob/rolling/launch/doc/source/architecture.rst
- https://docs.ros.org/en/rolling/p/launch/architecture.html

Tested with `rosdoc2`.


christophebedard:
This is a WIP just to get feedback on the execution.

launch_ros/doc/source/actions.rst
```diff
@@ -0,0 +1,50 @@
...
+   * - ``namespace``
+     - ``namespace``
+   * - ``node_name``
+     - ``node-name``
+   * - ``package``
+     - ``pkg``
+   * - ``parameters``
+     - ``param`` (``name``, ``value``)
+   * - ``remappings``
+     - ``remap`` (``from``, ``to``)
```
christophebedard, line 35: For arguments like this, it's a bit more complicated than just providing these names. What people really need are examples. I'm tempted to link to each action's frontend tests (at least for the ones that have frontend tests in `test_launch_ros`), like this one for the `Node` action: https://github.com/ros2/launch_ros/blob/38152b2d04c6b6f126d31726ac9e12888b2e2e1f/test_launch_ros/test/test_launch_ros/frontend/test_node_frontend.py. However, I'm not sure how to do it properly (i.e., how to get a distro-specific link to the source code, or maybe the test file itself can be included in the docs?), and this could get outdated, so I'd just overall like to get some feedback.

christophebedard:
Any feedback here? The more I think about it, the more I think we should just document the names of the frontend arguments in the Python class documentation. Then it just shows up in the API docs, and it's extremely close to where the argument names are defined (in the same class).

fujitatomoya (commented):
i say this is nice enhancement for user-experience, especially yaml and xml users to launch the node. and i also agree on adding examples to https://docs.ros.org/en/rolling/How-To-Guides/Launch-file-different-formats.html, this would be the 1st place for user to visit.

clalancette (commented):
So I agree with the sentiment that examples would also be good.

That said, I'm fine with adding in additional documentation.  I especially like that this page is sort of at the "front" of the generated API documentation, so easy to find.

One thought here is how the individual actions are documented.  It does seem like this is a bit of duplication of the pydoc strings that are in the class.  Is it possible/would it make sense to have this "front page" documentation link to deeper into the generated API documentation?  That way it will all be in sync.

christophebedard:
Adding examples to the ROS 2 docs is a good idea.

> One thought here is how the individual actions are documented. It does seem like this is a bit of duplication of the pydoc strings that are in the class. Is it possible/would it make sense to have this "front page" documentation link to deeper into the generated API documentation? That way it will all be in sync.

so I'd just list some actions on this page and link to their API doc page? The frontend argument names would have to be documented in the Python class or constructor docstring (because currently they're not documented), but yes I agree, this would keep the actual documentation in the source code itself.

rkent:
When you say that you tested this with rosdoc2, what exactly do you mean? That is, where did you obtain rosdoc2, and how exactly did you run rosdoc2. I'd like to make recommendations on how this could work better with rosdoc2 both present and future, but that is challenging because rosdoc2 is in flux.

christophebedard:
By "tested with rosdoc2," I meant that the two simple pages I added show up in the generated docs. I ignored the rest.

It's been a while, but:

```
$ pip install git+https://github.com/ros-infrastructure/rosdoc2.git@main  # Back in August 2024, probably
$ python3 -c 'import rosdoc2; print(rosdoc2.__version__)'
0.1.0
$ rosdoc2 build --package-path src/ros2/launch_ros/launch_ros/
```

Issue #408, opened 2024-07-29
:farmer: check_node_launch_test failing consistently in Iron and Humble Connext


Crola1702:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu Jammy
- Installation type:
  - Source
- Version or commit hash:
  - Iron and Humble
- DDS implementation:
  - RTI Connext

#### Steps to reproduce issue
1. Run a build in [Humble](https://build.ros2.org/job/Hci__nightly-connext_ubuntu_jammy_amd64/) or [Iron](https://build.ros2.org/job/Ici__nightly-connext_ubuntu_jammy_amd64/) connect
2 See check_node_launch_test fail

#### Additional information

Reference build: 
* [Ici__nightly-connext_ubuntu_jammy_amd64#151](https://build.ros2.org/job/Ici__nightly-connext_ubuntu_jammy_amd64/151) 
* [Hci__nightly-connext_ubuntu_jammy_amd64#258](https://build.ros2.org/job/Hci__nightly-connext_ubuntu_jammy_amd64/258)

Test regression:
* [launch_testing_ros.test.examples.check_node_launch_test.check_node_launch_test](https://build.ros2.org/job/Ici__nightly-connext_ubuntu_jammy_amd64/151/testReport/junit/launch_testing_ros.test.examples/check_node_launch_test/check_node_launch_test/)

<details>
<summary>
Log output:
</summary>

```
=================================== FAILURES ===================================
_____________________ launch tests: check_node_launch_test _____________________

======================================================================================================================================================================================================
FAIL: check_node_launch_test.TestFixture.test_node_start
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/ws/src/ros2/launch_ros/launch_testing_ros/test/examples/check_node_launch_test.py", line 56, in test_node_start
    assert node.wait_for_node('demo_node_1', 8.0), 'Node not found !'
AssertionError: Node not found !
assert False
 +  where False = <bound method MakeTestNode.wait_for_node of <check_node_launch_test.MakeTestNode object at 0x7facd297a770>>('demo_node_1', 8.0)
 +    where <bound method MakeTestNode.wait_for_node of <check_node_launch_test.MakeTestNode object at 0x7facd297a770>> = <check_node_launch_test.MakeTestNode object at 0x7facd297a770>.wait_for_node
----------------------------- Captured stdout call -----------------------------
[...]

PR #394, opened 2024-03-05
Make launch_testing_ros examples more robust.

clalancette:
The original goal here was to expand the timeouts on the launch_testing_ros examples so that they are more robust.  I did that, and in most cases these tests now wait at least 15 seconds for things to happen.

Along the way, I also somewhat rewrote some
of these tests so they properly clean up after themselves.

PR #419, opened 2024-10-15
Skip test_node_start test failing with Connext

Blast545:
As the title says, see: https://github.com/ros2/launch_ros/issues/408

### DIFF
--- a/launch_testing_ros/test/examples/check_node_launch_test.py
+++ b/launch_testing_ros/test/examples/check_node_launch_test.py
@@ -54,7 +54,6 @@ class TestFixture(unittest.TestCase):
     @unittest.skipIf(
         get_rmw_implementation_identifier() == 'rmw_connext_cpp',
         reason='Test failing with Connext')
-
     def test_node_start(self, proc_output):
         rclpy.init()
         try:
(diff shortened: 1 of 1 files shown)


Blast545:
Testing with Iron and Connext Only up-to `launch_testing_ros` to verify that the test is being skipped before running a full CI, see:

[Build Status](https://ci.ros2.org/job/ci_linux/21993/)

Blast545:
Test still enabled in first attempt, trying again with different string:

[Build Status](https://ci.ros2.org/job/ci_linux/21993/)

Blast545:
@clalancette PTAL, @Crola1702 fyi

fujitatomoya (commented):
instead of skipping the test, how about backporting https://github.com/ros2/launch_ros/pull/394 to make it robust? test fails with humble too, so fixing the problem would be better to support humble LTS?

Blast545:
@fujitatomoya Are you sure #394 will fix the issue after being backported? Just saw that you opened #420, happy to wait if you are taking a look into it.

fujitatomoya:
yeah i think so, that includes some cosmetic fixes, but main focus was to extend the timeout.

CC: @clalancette

clalancette:
Closing this in favor of #420, which was just merged.

PR #425, opened 2024-11-07
mock launch components causing rosdoc2 hang

### LINKED ISSUE #424: Issues preventing rosdoc2 from generating output (API docs)
## Bug report

I've been investigating why launch_ros does not generate any usable python API. That is, why does a url like http://docs.ros.org/en/humble/p/launch_ros/launch_ros.actions.composable_node_container.html show nothing useful.

**Required Info:**

- Operating System:
  - Ubuntu 24.04
- Installation type:
  - NA, running from rosdoc2, either locally or from the official build farm.
- Version or commit hash:
  - a13359b (which is current rolling)
- DDS implementation:
  - NA
- Client library (if applicable):
  - NA

#### Steps to reproduce issue
View http://docs.ros.org/en/humble/p/launch_ros/launch_ros.actions.node.html

#### Expected behavior
Useful docs

#### Actual behavior
Basically empty

#### Additional information

There are multiple reasons why this is failing, but all related to mocking of modules in sphinx autodoc. The failure with the stock code is:
```
  File "/home/kent/github/rkent/launch_ros/launch_ros/launch_ros/parameters_type.py", line 50, in <module>
    SomeSubstitutionsType_types_tuple +
TypeError: unsupported operand type(s) for +: 'SomeSubstitutionsType_types_tuple' and 'tuple'
 [autodoc.import_object]
```
```SomeSubstitutionsType_types_tuple``` is a mocked object, and it does not like to be added.

If you fix that (for example, surrounding it with a try block) there is a new failure. The run hangs with the displayed message:
```
reading sources... [  6%] launch_ros.actions
```

I've investigated this issue, and AFAICT it is a sphinx bug with mocked objects. There are actually two levels of this problem. Any attempt to use a mocked object as a decorator seems to generate a delay, then an error about a wrapped loop. But in ```node``` it generates an irrecoverable hang. I've developed a minimum test case to demonstrate this, and the offending code is:
```python
@expose_action('node')
class Node():
    """Action that executes a ROS node."""

    UNSPECIFIED_NODE_NAME = '<node_name_unspecified>'
```
where ```expose_action``` is a sphinx mocked object. Without the line ```UNSPECIFIED_NODE_NAME ...``` you get the delay and error, but with that line you get a hang. This is reproducible running sphinx directly without rosdoc2.

I'll probably be doing a PR that manually inserts appropriate mocked objects that work correctly, but I wanted to let anyone who cares see that I am investigating.

rkent:
Fixes #424 

There are multiple bug workarounds for upstream packages here.

- sphinx is slow with mocked decorators, because of a wrapping loop that has to error out
- sphinx with mock decorators hangs completely if there is an attribute defined on a class with a mocked decorator
- sphinx mocked objects as types do not support addition

Within rosdoc2, the supplied code has (relatively minor) workarounds for https://github.com/ros-infrastructure/rosdoc2/issues/156 (hence the supplied additional "Design" document), as well as a complex sys.path in conf.py that will not be needed after https://github.com/ros-infrastructure/rosdoc2/pull/155

But this PR as supplied should work with the existing rosdoc2.

### DIFF
--- a/launch_ros/doc/conf.py
+++ b/launch_ros/doc/conf.py
@@ -1,0 +1,57 @@
+# Copyright 2024 R. Kent James
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# -- Path setup --------------------------------------------------------------
+
+import sys
+
+from sphinx.ext.autodoc.mock import mock
+
+# -- Project information -----------------------------------------------------
+
+copyright = 'The <launch_ros> Contributors. License: Apache License 2.0'  # noqa A001
+
+# -- General configuration ---------------------------------------------------
+
+master_doc = 'index'
+
+source_suffix = {
+    '.rst': 'restructuredtext',
+    '.md': 'markdown',
+    '.markdown': 'markdown',
+}
+
+# Add mocks for decorators to prevent rosdoc2 hangs
+
+with mock(['launch']):
+    import launch  # noqa F401
+    module = sys.modules['launch']
+sys.modules['launch'] = module
+
+with mock(['launch.frontend']):
+    import launch.frontend  # noqa F401
+    module = sys.modules['launch.frontend']
+    # These are dummy decorators
+    module.expose_action = lambda _: lambda _: _
+    module.expose_substitution = lambda _: lambda _: _
+sys.modules['launch.frontend'] = module
+
+with mock(['launch.some_substitutions_type']):
+    import launch.some_substitutions_type  # noqa F401
+    module = sys.modules['launch.some_substitutions_type']
+    module.SomeSubstitutionsType_types_tuple = ()
+sys.modules['launch.some_substitutions_type'] = module
+
+# We still want autodoc to mock the base 'launch' module
+sys.modules.pop('launch')
--- a/launch_ros/doc/design.md
+++ b/launch_ros/doc/design.md
@@ -1,0 +1,4 @@
+Design Documentation
+====================
+
+See [the design document](https://design.ros2.org/articles/roslaunch.html) for more information on the ros2 launch system.


rkent:
This is now ready for review, I fixed the linting issues.

Just be clear what this is about. The existing launch_ros documentation in rosdoc2 does not show the python API. So if you go to https://docs.ros.org/en/rolling/p/launch_ros/launch_ros.actions.composable_node_container.html the documentation is empty.

I have a run of rosdoc2 available to view with this PR applied,  try https://prrosdoc2.rosdabbler.com/rolling/launch_ros/ Looking at that composable_node_container at https://prrosdoc2.rosdabbler.com/rolling/launch_ros/launch_ros.actions.composable_node_container.html you can see the python API documentation being shown.

tfoote:
This is great to resolve. Are there upstream issues that we could link from the code with comments such that people can know when/if these workarounds can be removed. Or know why they're there so that they don't get removed and accidentally break the documentation.

rkent:
I could not find existing upstream issues related to the problems I am trying to fix. Sphinx had an [earlier PR](https://github.com/sphinx-doc/sphinx/pull/6719) that supposedly supports mocked decorators, not sure why it is not working for us.

The other issue, with SomeSubstitutionsType_types_tuple, is a different issue. I think it is because mocked objects to not support addition.

With >1200 open issues in Sphinx, and my limited understanding of how all of this is supposed to work in Sphinx, I did not feel like I could contribute a meaningful issue in Sphinx that would be actionable. I could not find existing issues that seemed relevant.

At the moment I am not aware of other ros2 packages with similar problems, but finding these requires noting that a package is supposed to have documentation, but it is not actually appearing, which is not easy to automate over almost 2000 packages. I think I started focusing on this repo for an unrelated issue that was causing rosdoc2 to die completely, which turned out to be related to me running Sphinx 8 in my build farm.

The solution that this PR proposes essentially preloads the sys.modules cache with the specific mocked object that we want. It therefore is dependent on the implementation of sphinx.modules. I don't like that, it seems fragile and opaque, but I am not sure what else to do. Because this is confined to conf.py and therefore just the documentation, the fragility is not as risky as it would be in the active code of the package.

But I am open to any suggestions on how to improve this.

MichaelOrlov:
@tfoote Friendly ping for the follow-up review.

tfoote (approved):
Sorry this got lost in my queue.

ahcorde:
Pulls: ros2/launch_ros#425
Gist: https://gist.githubusercontent.com/ahcorde/feb1f71045f7130f58089d73ea870b38/raw/74b3b697bc463b78cf8c3ab7bad41a5f39628e48/ros2.repos
BUILD args:  --packages-above-and-dependencies launch_ros  --packages-above-and-dependencies launch_ros
TEST args:  --packages-above launch_ros  --packages-above launch_ros
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15111
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22441/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16693/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/1968/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23230/)

ahcorde (approved):

rkent:
This seems to work, http://docs.ros.org/en/rolling/p/launch_ros/launch_ros.html now shows the documentation.

Not sure what your procedure to backport this to iron and humble is, but might be worth doing.

christophebedard:
@Mergifyio backport jazzy humble

PR #430, opened 2024-12-20
Autostarting lifecycle nodes and example launch file demo

### LINKED ISSUE #418: Create option for automatic bringup of lifecycle nodes in launch_ros API
#### Feature description

Currently, transitioning up a lifecycle node can be done via an external node, lifecycle manager, or using Opaque functions with some rather nasty looking boilerplate with events. The event-based solution only works for `LifecycleNode` as that has the implementations for getting the states required. However, a `ComponentNode` can also be a lifecycle node and there is no way to my knowledge to perform lifecycle operations in launch_ros directly on the `ComponentNode` object. 

It would be nice if there was a `LifecycleNode` and `ComponentNode` field for `active_on_bringup` (or eq.) that incorporated this boilerplate to make this easier as many users of lifecycle nodes don't truly care to transition up and control the lifecycle of all their programs with such granularity.   

I don't think there are any unique implementation considerations.

SteveMacenski:
This resolves #418 which implements autostarting lifecycle nodes. This is complete and ready for a review 

You'll notice a couple of key changes worth explaining:

1. There is a new `is_lifecycle_node` attribute of the node and lifecycle nodes which is needed to remove a circular dependency on the LifecycleNode within the `LifecycleTransition` class which only is type checking. I replace this type check with checking if a non-None object (1) has the attribute at all and (2) has a lifecycle attribute with appropriate logging between the difference of a non-lifecycle node being attempted vs a non-node. This has been tested as well to function using the demos. 

2. You will also notice that I refactored lifecycle node to have a separate util `LifecycleEventManager`, who handles the event emitting, handling, and topic listening for lifecycle nodes. This way, this can be used in lifecycle-components as well! No changes were made here except removing the `self.__current_state` variable which was completely unused. The `LifecycleEventManager` is only created if the node requests autostart at Launch time and otherwise has no overhead nor exposes the interfaces/event handler/emitter if its a non-lifecycle node. 

3. The component nodes don't add a `/` before the node names with the namespaces. This messes with the node event matcher. I add in a leading `/` so that the `action.node_name == node_name` which has the `/` forcably applied in the node event matcher 

4. I tested `LifecycleNode`, `ComposableNodeContainer`, `LoadCompoableNodes` for all cases: `autostart=True` `autostart=False`, and `autostart` field not supplied. You can also run any experiments you like using the 2 extra launch files I provide in the examples directory which shows all the features at work/

### DIFF
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -14,9 +14,6 @@
 
 """Module for the LifecycleNode action."""
 
-import functools
-import threading
-from typing import cast
 from typing import List
 from typing import Optional
 
@@ -30,10 +27,7 @@ import lifecycle_msgs.srv
 
 from .lifecycle_transition import LifecycleTransition
 from .node import Node
-from ..events.lifecycle import ChangeState
-from ..events.lifecycle import StateTransition
 
-from ..ros_adapters import get_ros_node
 from ..utilities import LifecycleEventManager
 
 
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -86,7 +86,7 @@ class LifecycleNode(Node):
         if '<node_name_unspecified>' in self.node_name:
             raise RuntimeError('node_name unexpectedly incomplete for lifecycle node')
 
-        self.__lifecycle_event_manager = LifecycleEventManager(self.node_name)
+        self.__lifecycle_event_manager = LifecycleEventManager(self)
         self.__lifecycle_event_manager.setup_lifecycle_manager(context)
 
         # If autostart is enabled, transition to the 'active' state.
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -76,6 +76,7 @@ class LifecycleNode(Node):
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
+        self.__rclpy_change_state_client = None
 
     @property
     def is_lifecycle_node(self):
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -76,8 +76,6 @@ class LifecycleNode(Node):
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
         self.__rclpy_subscription = None
-        self.__current_state = \
-            ChangeState.valid_states[lifecycle_msgs.msg.State.PRIMARY_STATE_UNKNOWN]
 
     @property
     def is_lifecycle_node(self):
@@ -86,7 +84,6 @@ class LifecycleNode(Node):
     def _on_transition_event(self, context, msg):
         try:
             event = StateTransition(action=self, msg=msg)
-            self.__current_state = ChangeState.valid_states[msg.goal_state.id]
             context.asyncio_loop.call_soon_threadsafe(lambda: context.emit_event_sync(event))
         except Exception as exc:
             self.__logger.error(
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -34,6 +34,7 @@ from ..events.lifecycle import ChangeState
 from ..events.lifecycle import StateTransition
 
 from ..ros_adapters import get_ros_node
+from ..utilities import LifecycleEventManager
 
 
 class LifecycleNode(Node):
@@ -75,69 +76,11 @@ class LifecycleNode(Node):
         """
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
-        self.__rclpy_subscription = None
-        self.__rclpy_change_state_client = None
+        self.__lifecycle_event_manager = None
 
     @property
     def is_lifecycle_node(self):
         return True
-
-    def _on_transition_event(self, context, msg):
-        try:
-            event = StateTransition(action=self, msg=msg)
-            context.asyncio_loop.call_soon_threadsafe(lambda: context.emit_event_sync(event))
-        except Exception as exc:
-            self.__logger.error(
-                "Exception in handling of 'lifecycle.msg.TransitionEvent': {}".format(exc))
-
-    def _call_change_state(self, request, context: launch.LaunchContext):
-        while not self.__rclpy_change_state_client.wait_for_service(timeout_sec=1.0):
-            if context.is_shutdown:
-                self.__logger.warning(
-                    "Abandoning wait for the '{}' service, due to shutdown.".format(
-                        self.__rclpy_change_state_client.srv_name),
-                )
-                return
-
-        # Asynchronously wait so that we can periodically check for shutdown.
-        event = threading.Event()
-
-        def unblock(future):
-            nonlocal event
-            event.set()
-
-        response_future = self.__rclpy_change_state_client.call_async(request)
-        response_future.add_done_callback(unblock)
-
-        while not event.wait(1.0):
-            if context.is_shutdown:
-                self.__logger.warning(
-                    "Abandoning wait for the '{}' service response, due to shutdown.".format(
-                        self.__rclpy_change_state_client.srv_name),
-                )
-                response_future.cancel()
-                return
-
-        if response_future.exception() is not None:
-            raise response_future.exception()
-        response = response_future.result()
-
-        if not response.success:
-            self.__logger.error(
-                "Failed to make transition '{}' for LifecycleNode '{}'".format(
-                    ChangeState.valid_transitions[request.transition.id],
-                    self.node_name,
-                )
-            )
-
-    def _on_change_state_event(self, context: launch.LaunchContext) -> None:
-        typed_event = cast(ChangeState, context.locals.event)
-        if not typed_event.lifecycle_node_matcher(self):
-            return None
-        request = lifecycle_msgs.srv.ChangeState.Request()
-        request.transition.id = typed_event.transition_id
-        context.add_completion_future(
-            context.asyncio_loop.run_in_executor(None, self._call_change_state, request, context))
 
     def execute(self, context: launch.LaunchContext) -> Optional[List[Action]]:
         """
@@ -148,22 +91,9 @@ class LifecycleNode(Node):
         self._perform_substitutions(context)  # ensure self.node_name is expanded
         if '<node_name_unspecified>' in self.node_name:
             raise RuntimeError('node_name unexpectedly incomplete for lifecycle node')
-        node = get_ros_node(context)
-        # Create a subscription to monitor the state changes of the subprocess.
-        self.__rclpy_subscription = node.create_subscription(
-            lifecycle_msgs.msg.TransitionEvent,
-            '{}/transition_event'.format(self.node_name),
-            functools.partial(self._on_transition_event, context),
-            10)
-        # Create a service client to change state on demand.
-        self.__rclpy_change_state_client = node.create_client(
-            lifecycle_msgs.srv.ChangeState,
-            '{}/change_state'.format(self.node_name))
-        # Register an event handler to change states on a ChangeState lifecycle event.
-        context.register_event_handler(launch.EventHandler(
-            matcher=lambda event: isinstance(event, ChangeState),
-            entities=[launch.actions.OpaqueFunction(function=self._on_change_state_event)],
-        ))
+
+        self.__lifecycle_event_manager = LifecycleEventManager(self.node_name)
+        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
 
         # If autostart is enabled, transition to the 'active' state.
         autostart_actions = None
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -31,6 +31,7 @@ import lifecycle_msgs.srv
 from .node import Node
 from ..events.lifecycle import ChangeState
 from ..events.lifecycle import StateTransition
+from .lifecycle_transition import LifecycleTransition
 
 from ..ros_adapters import get_ros_node
 
@@ -77,6 +78,10 @@ class LifecycleNode(Node):
         self.__rclpy_subscription = None
         self.__current_state = \
             ChangeState.valid_states[lifecycle_msgs.msg.State.PRIMARY_STATE_UNKNOWN]
+
+    @property
+    def is_lifecycle_node(self):
+        return True
 
     def _on_transition_event(self, context, msg):
         try:
@@ -161,5 +166,27 @@ class LifecycleNode(Node):
             matcher=lambda event: isinstance(event, ChangeState),
             entities=[launch.actions.OpaqueFunction(function=self._on_change_state_event)],
         ))
+
+        # If autostart is enabled, transition to the 'active' state.
+        autostart_actions = None
+        if self.node_autostart:
+            autostart_actions = [
+                LifecycleTransition(
+                    lifecycle_node_names=[self.node_name],
+                    transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_CONFIGURE]
+                ),
+                LifecycleTransition(
+                    lifecycle_node_names=[self.node_name],
+                    transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE]
+                ),
+            ]
+
         # Delegate execution to Node and ExecuteProcess.
-        return super().execute(context)
+        node_actions = super().execute(context)  # type: Optional[List[Action]]
+        if node_actions is not None and autostart_actions is not None:
+            return node_actions + autostart_actions
+        if node_actions is not None:
+            return node_actions
+        if autostart_actions is not None:
+            return autostart_actions
+        return None
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -28,10 +28,10 @@ import launch.logging
 import lifecycle_msgs.msg
 import lifecycle_msgs.srv
 
+from .lifecycle_transition import LifecycleTransition
 from .node import Node
 from ..events.lifecycle import ChangeState
 from ..events.lifecycle import StateTransition
-from .lifecycle_transition import LifecycleTransition
 
 from ..ros_adapters import get_ros_node
 
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -39,6 +39,7 @@ class LifecycleNode(Node):
         *,
         name: SomeSubstitutionsType,
         namespace: SomeSubstitutionsType,
+        autostart: Optional[bool] = False,
         **kwargs
     ) -> None:
         """
@@ -70,7 +71,13 @@ class LifecycleNode(Node):
         """
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
+        self.__autostart = autostart
         self.__lifecycle_event_manager = None
+
+    @property
+    def node_autostart(self):
+        """Getter for autostart."""
+        return self.__autostart
 
     @property
     def is_lifecycle_node(self):
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -68,6 +68,8 @@ class LifecycleNode(Node):
         :param name: The name of the lifecycle node.
           Although it defaults to None it is a required parameter and the default will be removed
           in a future release.
+        :param namespace: The ROS namespace for this Node
+        :param autostart: Whether or not to automatically transition to the 'active' state.
         """
         super().__init__(name=name, namespace=namespace, **kwargs)
         self.__logger = launch.logging.get_logger(__name__)
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -39,7 +39,7 @@ class LifecycleNode(Node):
         *,
         name: SomeSubstitutionsType,
         namespace: SomeSubstitutionsType,
-        autostart: Optional[bool] = False,
+        autostart: bool = False,
         **kwargs
     ) -> None:
         """
--- a/launch_ros/launch_ros/actions/lifecycle_node.py
+++ b/launch_ros/launch_ros/actions/lifecycle_node.py
@@ -85,6 +85,9 @@ class LifecycleNode(Node):
     def is_lifecycle_node(self):
         return True
 
+    def get_lifecycle_event_manager(self):
+        return self.__lifecycle_event_manager
+
     def execute(self, context: launch.LaunchContext) -> Optional[List[Action]]:
         """
         Execute the action.
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -38,7 +38,10 @@ from launch.utilities import normalize_to_list_of_substitutions
 from launch.utilities import perform_substitutions
 from launch_ros.parameter_descriptions import ParameterFile
 
+import lifecycle_msgs.msg
+
 from .composable_node_container import ComposableNodeContainer
+from .lifecycle_transition import LifecycleTransition
 
 from ..descriptions import ComposableNode
 from ..ros_adapters import get_ros_node
@@ -230,11 +233,32 @@ class LoadComposableNodes(Action):
         # Generate load requests before execute() exits to avoid race with context changing
         # due to scope change (e.g. if loading nodes from within a GroupAction).
         load_node_requests = []
+        autostart_actions = []
         for node_description in self.__composable_node_descriptions:
             request = get_composable_node_load_request(node_description, context)
             # The request can be None if the node description's condition evaluates to False
             if request is not None:
                 load_node_requests.append(request)
+
+            # If autostart is enabled, transition to the 'active' state.
+            if node_description.node_autostart:
+                complete_node_name = request.node_namespace + request.node_name
+                if not complete_node_name.startswith('/'):
+                    complete_node_name = '/' + complete_node_name
+                self.__logger.info(
+                    'Autostart enabled for requested lifecycle node {}'.format(complete_node_name))
+                node_description.init_lifecycle_event_manager(complete_node_name, context)
+                autostart_actions.append(
+                    LifecycleTransition(
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_CONFIGURE]
+                    ))
+                autostart_actions.append(
+                    LifecycleTransition(
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE]
+                    ),
+                )
 
         if load_node_requests:
             context.add_completion_future(
@@ -242,6 +266,14 @@ class LoadComposableNodes(Action):
                     None, self._load_in_sequence, load_node_requests, context
                 )
             )
+
+        load_actions = super().execute(context)
+        if load_actions is not None and len(autostart_actions) != 0:
+            return load_actions + autostart_actions
+        if load_actions is not None:
+            return load_actions
+        if len(autostart_actions) != 0:
+            return autostart_actions
 
 
 def get_composable_node_load_request(
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -247,7 +247,7 @@ class LoadComposableNodes(Action):
                     complete_node_name = '/' + complete_node_name
                 self.__logger.info(
                     'Autostart enabled for requested lifecycle node {}'.format(complete_node_name))
-                node_description.init_lifecycle_event_manager(complete_node_name, context)
+                node_description.init_lifecycle_event_manager(node_description, context)
                 autostart_actions.append(
                     LifecycleTransition(
                         lifecycle_node_names=[complete_node_name],
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -247,7 +247,7 @@ class LoadComposableNodes(Action):
                     complete_node_name = '/' + complete_node_name
                 self.__logger.info(
                     'Autostart enabled for requested lifecycle node {}'.format(complete_node_name))
-                node_description.init_lifecycle_event_manager(node_description, context)
+                node_description.init_lifecycle_event_manager(context)
                 autostart_actions.append(
                     LifecycleTransition(
                         lifecycle_node_names=[complete_node_name],
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -241,7 +241,7 @@ class LoadComposableNodes(Action):
                 load_node_requests.append(request)
 
             # If autostart is enabled, transition to the 'active' state.
-            if node_description.node_autostart:
+            if hasattr(node_description, 'node_autostart') and node_description.node_autostart:
                 complete_node_name = request.node_namespace + request.node_name
                 if not complete_node_name.startswith('/'):
                     complete_node_name = '/' + complete_node_name
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -125,6 +125,7 @@ class Node(ExecuteProcess):
         exec_name: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
+        autostart: Optional[bool] = False,
         ros_arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         **kwargs
@@ -222,6 +223,7 @@ class Node(ExecuteProcess):
         self.__node_executable = executable
         self.__node_name = name
         self.__node_namespace = namespace
+        self.__autostart = autostart
         self.__parameters = [] if parameters is None else normalized_params
         self.__remappings = [] if remappings is None else list(normalize_remap_rules(remappings))
         self.__ros_arguments = ros_arguments
@@ -361,6 +363,15 @@ class Node(ExecuteProcess):
         if self.__final_node_name is None:
             raise RuntimeError("cannot access 'node_name' before executing action")
         return self.__final_node_name
+
+    @property
+    def node_autostart(self):
+        """Getter for autostart."""
+        return self.__autostart
+
+    @property
+    def is_lifecycle_node(self):
+        return False
 
     def is_node_name_fully_specified(self):
         keywords = (self.UNSPECIFIED_NODE_NAME, self.UNSPECIFIED_NODE_NAMESPACE)
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -125,7 +125,6 @@ class Node(ExecuteProcess):
         exec_name: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
-        autostart: Optional[bool] = False,
         ros_arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         arguments: Optional[Iterable[SomeSubstitutionsType]] = None,
         **kwargs
@@ -223,7 +222,6 @@ class Node(ExecuteProcess):
         self.__node_executable = executable
         self.__node_name = name
         self.__node_namespace = namespace
-        self.__autostart = autostart
         self.__parameters = [] if parameters is None else normalized_params
         self.__remappings = [] if remappings is None else list(normalize_remap_rules(remappings))
         self.__ros_arguments = ros_arguments
@@ -367,7 +365,7 @@ class Node(ExecuteProcess):
     @property
     def node_autostart(self):
         """Getter for autostart."""
-        return self.__autostart
+        return False
 
     @property
     def is_lifecycle_node(self):
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -362,15 +362,6 @@ class Node(ExecuteProcess):
             raise RuntimeError("cannot access 'node_name' before executing action")
         return self.__final_node_name
 
-    @property
-    def node_autostart(self):
-        """Getter for autostart."""
-        return False
-
-    @property
-    def is_lifecycle_node(self):
-        return False
-
     def is_node_name_fully_specified(self):
         keywords = (self.UNSPECIFIED_NODE_NAME, self.UNSPECIFIED_NODE_NAMESPACE)
         return all(x not in self.node_name for x in keywords)
--- a/launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
+++ b/launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
@@ -19,7 +19,6 @@ from typing import Optional
 
 import launch
 from launch.substitution import Substitution
-# from launch.utilities import ensure_argument_type
 from launch.utilities import perform_substitutions
 from launch_ros.parameters_type import Parameters
 from launch_ros.remap_rule_type import RemapRules
--- a/launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
+++ b/launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
@@ -1,0 +1,121 @@
+# Copyright 2025 Open Navigation LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Module for a description of a ComposableLifecycleNode."""
+
+from typing import List
+from typing import Optional
+
+import launch
+from launch.condition import Condition
+from launch.some_substitutions_type import SomeSubstitutionsType
+from launch.substitution import Substitution
+# from launch.utilities import ensure_argument_type
+from launch.utilities import perform_substitutions
+from launch_ros.parameters_type import Parameters
+from launch_ros.parameters_type import SomeParameters
+from launch_ros.remap_rule_type import RemapRules
+from launch_ros.remap_rule_type import SomeRemapRules
+from launch_ros.utilities import LifecycleEventManager
+
+from .composable_node import ComposableNode
+
+
+class ComposableLifecycleNode(ComposableNode):
+    """Describes a lifecycle node that can be loaded into a container with other nodes."""
+
+    def __init__(
+        self, *,
+        package: SomeSubstitutionsType,
+        plugin: SomeSubstitutionsType,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: Optional[SomeSubstitutionsType] = None,
+        parameters: Optional[SomeParameters] = None,
+        autostart: Optional[bool] = False,
+        remappings: Optional[SomeRemapRules] = None,
+        extra_arguments: Optional[SomeParameters] = None,
+        condition: Optional[Condition] = None,
+    ) -> None:
+        """
+        Initialize a ComposableNode description.
+
+        :param package: name of the ROS package the node plugin lives in
+        :param plugin: name of the plugin to be loaded
+        :param name: name to give to the ROS node
+        :param namespace: namespace to give to the ROS node
+        :param parameters: list of either paths to yaml files or dictionaries of parameters
+        :param autostart: Whether to autostart lifecycle node in the activated state
+        :param remappings: list of from/to pairs for remapping names
+        :param extra_arguments: container specific arguments to be passed to the loaded node
+        :param condition: action will be executed if the condition evaluates to true
+        """
+        super().__init__(
+            package=package,
+            plugin=plugin,
+            name=name,
+            namespace=namespace,
+            parameters=parameters,
+            remappings=remappings,
+            extra_arguments=extra_arguments,
+            condition=condition)
+
+        self.__autostart = autostart
+        self.__lifecycle_event_manager = None
+        self.__node_name = self._ComposableNode__node_name
+
+    def init_lifecycle_event_manager(self, node, context: launch.LaunchContext) -> None:
+        # LifecycleEventManager needs a pre-substitution node name
+        self.__node_name = perform_substitutions(context, node.node_name)
+        self.__lifecycle_event_manager = LifecycleEventManager(node)
+        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
+
+    @property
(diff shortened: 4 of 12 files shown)


SteveMacenski:
@mjcarroll can I get a review on this now that we're back from the holidays? :-)

doisyg:
That would be amazing to have this part of the standard launch!
(and I can already think about a next step: choose if you want to start in `Unconfigured`, `Inactive` or `Active`)

SteveMacenski:
@mjcarroll can I request a review? :heart:

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -145,21 +85,30 @@ def execute(self, context: launch.LaunchContext) -> Optional[List[Action]]:
...
+        if self.node_autostart:
+            autostart_actions = [
+                LifecycleTransition(
+                    lifecycle_node_names=[self.node_name],
+                    transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_CONFIGURE]
+                ),
+                LifecycleTransition(
+                    lifecycle_node_names=[self.node_name],
+                    transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE]
+                ),
```
wjwwood, line 115: Shouldn't you wait for the transition to configure to complete before sending the transition to activate?

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -230,19 +233,48 @@ def execute(
...
+                    LifecycleTransition(
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_CONFIGURE]
+                    ))
+                autostart_actions.append(
+                    LifecycleTransition(
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE]
+                    ),
+                )
```
wjwwood, line 261: Similar comment here about waiting before transitioning, rather than fire-forget with two subsequent transitions.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -230,19 +233,48 @@ def execute(
...
                 )
             )
 
+        load_actions = super().execute(context)
+        if load_actions is not None and len(autostart_actions) != 0:
+            return load_actions + autostart_actions
+        if load_actions is not None:
+            return load_actions
+        if len(autostart_actions) != 0:
+            return autostart_actions
```
wjwwood, line 276: I think this could just be:

```suggestion
        return autostart_actions
```

But if you want to call out to super I think that's ok. We just know (at least for now) that `launch.Action` always returns `None`:

https://github.com/ros2/launch/blob/a93ce70943c65b938d38eab78322dc27722b7a6f/launch/launch/action.py#L123-L129

But that being said, in case someone changes the base class of this class and doesn't notice this part of the code, calling super defensively isn't a bad idea.

Leave it or change it, doesn't matter to me, just wanted to point it out.

launch_ros/launch_ros/actions/node.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         exec_name: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
         remappings: Optional[SomeRemapRules] = None,
+        autostart: Optional[bool] = False,
```
wjwwood: Why does this need to be part of the Node action API? Doesn't it only need to be on a LifecycleNode action?

launch_ros/launch_ros/descriptions/composable_node.py
```diff
@@ -43,6 +45,7 @@ def __init__(
         name: Optional[SomeSubstitutionsType] = None,
         namespace: Optional[SomeSubstitutionsType] = None,
         parameters: Optional[SomeParameters] = None,
+        autostart: Optional[bool] = False,
```
wjwwood: Same comment as for Node action.

launch_ros/launch_ros/event_handlers/on_state_transition.py
```diff
@@ -34,7 +33,7 @@ def __init__(
         self,
         *,
         entities: SomeEntitiesType,
-        target_lifecycle_node: LifecycleNode = None,
+        target_lifecycle_node: Optional[SomeSubstitutionsType] = None,
```
wjwwood: Why does this need to be a `SomeSubstitutionsType`? Not saying it's wrong, just curious what use case you have in mind?

wjwwood: And further, I'm not sure this works because this would essentially be a type name as a string, but before this was a _value_ not a _type_, and we passed that instance of the object in to the event as `event.action`. So maybe I do think this isn't right.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
```diff
@@ -0,0 +1,142 @@
...
+
+        This utility also handles some events related to lifecycle:
+
+        - :class:`launch.events.lifecycle.ChangeState`
+
+          - this event can be targeted to a single lifecycle node, or more than
+            one, or even all lifecycle nodes, and it requests the targeted nodes
+            to change state, see its documentation for more details.
+
+        :param name: The name of the lifecycle node.
```
wjwwood: Might want to clarify here whether or not the name needs to be fully qualified or not.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
```diff
@@ -0,0 +1,142 @@
...
+        self.__rclpy_subscription = None
+        self.__rclpy_change_state_client = None
+
+    @property
+    def node_name(self):
+        return self.__node_name
+
+    def _on_transition_event(self, context, msg):
+        try:
+            event = StateTransition(action=self, msg=msg)
```
wjwwood: I don't think you can pass `action=self` here, see: https://github.com/ros2/launch_ros/blob/f9ab9371824d470bca47602bc7d277eb2217c90d/launch_ros/launch_ros/events/lifecycle/state_transition.py#L42

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
```diff
@@ -0,0 +1,142 @@
...
+                self.__logger.warning(
+                    "Abandoning wait for the '{}' service response, due to shutdown.".format(
+                        self.__rclpy_change_state_client.srv_name),
+                )
+                response_future.cancel()
+                return
+
+        if response_future.exception() is not None:
+            raise response_future.exception()
+        response = response_future.result()
```
wjwwood, line 104: `Future.result()` will throw for you, see:

https://github.com/ros2/rclpy/blob/2b38e662b3635fffe5b68309f75e39f9e7441602/rclpy/rclpy/task.py#L107-L118

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
```diff
@@ -0,0 +1,142 @@
...
+            self.__logger.error(
+                "Failed to make transition '{}' for LifecycleNode '{}'".format(
+                    ChangeState.valid_transitions[request.transition.id],
+                    self.__node_name,
+                )
+            )
+
+    def _on_change_state_event(self, context: launch.LaunchContext) -> None:
+        typed_event = cast(ChangeState, context.locals.event)
+        if not typed_event.lifecycle_node_matcher(self):
```
wjwwood: Again, if the user emits their own `ChangeState` pointing to their `LifecycleNode` action, then this will not match it right?

wjwwood (changes requested):
A few questions, but looking pretty good. Thanks for the contribution!

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: This is copy+pasted from the lifecycle node abstracted into a manager to share with composition node. I'd prefer not to make functional changes in this PR if that's OK with you. This has been the state of things previously and just reorganized to be shared among multiple node types

But w.r.t. this comment: No, actually! The subscription keeps them aligned: https://github.com/ros2/launch_ros/blob/84c44de2ab85eb03e55ddcd6491a79eceba38482/launch_ros/launch_ros/utilities/lifecycle_event_manager.py#L127-L131

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: This is copy+pasted from the lifecycle node abstracted into a manager to share with composition node. I'd prefer not to make functional changes in this PR if that's OK with you. This has been the state of things previously and just reorganized to be shared among multiple node types

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: Let me look into this one, you have a point

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: This is copy+pasted from the lifecycle node abstracted into a manager to share with composition node. I don't really know one way or another, I'm not the original author. I could try to find out, but I'd prefer if this got grandfathered in as 'this was already here'

launch_ros/launch_ros/event_handlers/on_state_transition.py
SteveMacenski: I needed this to not be specifically a `LifecycleNode` due to circular dependencies using this within LifecycleNode itself for the transitioning. So, to make sure that it is still a lifecycle-compliant node, I used the Optional type and checked that it was a proper instance of a lifecycle-compliant node using `is_lifecycle_node` (since components, lifecycle, and node in launch_ros can all be used to launch a rclcpp::LifecycleNode)

> Im not sure this works 

I can't comment confidently on the reasoning you have for it not working, but I can say from testing with the examples I provided that it does indeed work. I provided the testing samples here for that reason (and to augment the existing test samples of similar nature)

> but before this was a value not a type

Un-confidently: Isn't `LifecycleNode` a type and `Optional[SomeSubstitutionsType]` a type? Is there something else you'd suggest I try here instead? I just used `Optional[SomeSubstitutionsType]` based on what I saw elsewhere in the codebase for similar purposes, but I won't claim that this is _right_ -- only that it appears to _work_ and looks consistent to my naive eye with the rest of the codebase. My objective was to remove the circular dependencies, I don't know what non-LifecycleNode I should put here instead if not this if you think its wrong.

```
- target_lifecycle_node: LifecycleNode = None,
+ target_lifecycle_node: Optional[SomeSubstitutionsType] = None,
```

launch_ros/launch_ros/descriptions/composable_node.py
SteveMacenski: Composable nodes can be also lifecycle nodes -- see Nav2 :-) 

This is actually the main application for which I'm opening this PR. Its very difficult to work with lifecycle-components in launch_ros. There's no technical way (before this) to transition up a component that's a lifecycle node within launch. The other events that are used by the launch_ros.LifecycleNode reject updates from other node types.

launch_ros/launch_ros/actions/node.py
SteveMacenski: It is mostly so that it doesn't crash if someone uses it. `is_lifecycle_node` is set to false for this node type. The launch_ros.Node doesn't support the autostart usage. Its ignored. 

I could remove it if you preferred. I just didn't want people to get stuck with these kinds of traces when the other 2 node types support it:

```
Traceback (most recent call last):
  File "/root/jazzy_ws/./src/launch_ros/launch_ros/examples/lifecycle_autostart_pub_sub_launch.py", line 55, in <module>
    main()
  File "/root/jazzy_ws/./src/launch_ros/launch_ros/examples/lifecycle_autostart_pub_sub_launch.py", line 33, in main
    talker_node = launch_ros.actions.Node(
                  ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/jazzy_ws/./src/launch_ros/launch_ros/examples/../launch_ros/actions/node.py", line 221, in __init__
    super().__init__(cmd=cmd, **kwargs)
  File "/opt/ros/rolling/lib/python3.12/site-packages/launch/actions/execute_process.py", line 240, in __init__
    super().__init__(process_description=executable, **kwargs)
  File "/opt/ros/rolling/lib/python3.12/site-packages/launch/actions/execute_local.py", line 188, in __init__
    super().__init__(**kwargs)
TypeError: Action.__init__() got an unexpected keyword argument 'autostart'
```

launch_ros/launch_ros/actions/load_composable_nodes.py
SteveMacenski: I am copying the styling of the existing codebase here, this is how its done in all other locations I viewed, so I'd say we should leave it.

launch_ros/launch_ros/actions/lifecycle_node.py
SteveMacenski: LifecycleTransition handles this for us. 

1. We build a set of possible change events to emit at some point: https://github.com/ros2/launch_ros/blob/84c44de2ab85eb03e55ddcd6491a79eceba38482/launch_ros/launch_ros/actions/lifecycle_transition.py#L145-L155
2. Then we trigger the events when a transition occurs making it possible for an emit transition event to occur https://github.com/ros2/launch_ros/blob/84c44de2ab85eb03e55ddcd6491a79eceba38482/launch_ros/launch_ros/actions/lifecycle_transition.py#L167-L174
3. The actual event is registered and lying in wait to be performed  https://github.com/ros2/launch_ros/blob/84c44de2ab85eb03e55ddcd6491a79eceba38482/launch_ros/launch_ros/actions/lifecycle_transition.py#L195-L196

Thus, the events occur when they're possible and only when they're possible given the state already in the handy work done by the `lifecycle_transition.py` authors :smile: 

The first event is emited automatically in `lifecycle_transition.py` which then causes the rest of the cascading. If the configuration fails, it even enters the configured state, and thus activation is never triggered.

```
        for node_name in lifecycle_node_names:
            actions.append(emit_actions[node_name][0])
```

launch_ros/launch_ros/actions/load_composable_nodes.py
SteveMacenski: See above explanation

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: After looking into this in detail, you are right that this was non-functional. I tested node transitions up on Lifecycle/Composition, but didn't check if the transition_event callback was properly matching events.

I fixed it by making the `==` operator look at the lifecycle node rather than the lifecycle event management object & passing in the lifecycle node to the constructor. 

I tried also working on this by changing the matchers to use the `node_name` in the places that they directly compare the `event.actions == node`-style calls. It worked equally as well, but I felt it better to change less than more in the matchers in case some folks rely on it for very particular behavior.

SteveMacenski:
@wjwwood I addressed all of your comments and made a fix by your review requests

launch_ros/launch_ros/actions/node.py
wjwwood: I mean, it makes sense to me that a standard Node doesn't expect an autostart keyword. I would prefer it not to arguments it never uses.

launch_ros/launch_ros/event_handlers/on_state_transition.py
wjwwood: > Isn't LifecycleNode a type and Optional[SomeSubstitutionsType] a type? 

I mean, yes, but that's like saying replacing `rclcpp::Node` with `std::string` in c++ is ok because they're both types...

If it's just an issue with circular dependencies then for the Type annotations you can just string quote the type, e.g.:

```suggestion
        target_lifecycle_node: Optional['LifecycleNode'] = None,
```

Or I think you can do something like this at the top of the file:

```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
    from ... import LifecycleNode
```

See: https://stackoverflow.com/a/39757388

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
wjwwood: Well you may have copy-pasted it, but it's still important for you and the users to know what needs to be passed here. If you're only forwarding it to the other constructor, then maybe you can simply reference it from here rather than copy-pasting the explanation.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
wjwwood: > This is copy+pasted from the lifecycle node abstracted into a manager to share with composition node. I'd prefer not to make functional changes in this PR if that's OK with you. This has been the state of things previously and just reorganized to be shared among multiple node types

I get where you're coming from, but in this case your copy-paste is into a different class with a different type where you also use `self`, so you're essentially changing the behavior even though it's a copy-paste because of that.

I don't see  how the subscriptions keep them aligned? The `typed_event.lifecycle_node_matcher` is user defined and may not look at the subscription at all. Plus the matcher explicitly expects a `LifecycleNode` here not the type of self in your copy-pasted code:

https://github.com/ros2/launch_ros/blob/f9ab9371824d470bca47602bc7d277eb2217c90d/launch_ros/launch_ros/events/lifecycle/change_state.py#L92

Here's an example that won't likely work:

https://github.com/ros2/launch_ros/blob/f9ab9371824d470bca47602bc7d277eb2217c90d/launch_ros/examples/lifecycle_pub_sub_launch.py#L50

That is expecting the actions objects to match...

launch_ros/launch_ros/descriptions/composable_node.py
wjwwood: I guess I find it weird still that you'd pass an argument that may or may not be ignored. Wouldn't using the type system be good here, and having a composable lifecycle node type class to disambiguate it?

If we have a use case for a generic type that can be either a lifecycle node or something else, then maybe this is fine, but I believe you always know if a node is going to be lifecycle or not, otherwise you wouldn't pass autostart. Also, if you _think_ the node is lifecycle, and so you pass autostart, but it actually isn't, wouldn't you want to catch that? Seems like a bug waiting to be missed.

launch_ros/launch_ros/actions/lifecycle_node.py
wjwwood: Ok, makes sense, I missed the magic ✨ that `LifecycleTransition` was doing. That seems fine.

launch_ros/launch_ros/actions/node.py
wjwwood: I think it would be ok to check for the argument and raise a better error if you think it would help too.

launch_ros/launch_ros/event_handlers/on_state_transition.py
wjwwood: btw, didn't mean to come off snarky with the c++ comparison, but that is the sort of what's happening with the `SomeSubstitutionsType`, since that type essentially gets resolved into a string.

wjwwood:
I sent a few responses, resolved a few others. Sorry it took so long to reply, I'll try to be quicker on the next iteration. Thanks again!

launch_ros/launch_ros/actions/node.py
SteveMacenski: > I mean, it makes sense to me that a standard Node doesn't expect an autostart keyword. I would prefer it not to arguments it never uses.

A-OK, I think that's reasonable. Removing it. I think you're right and that error is acutally pretty semantically easy to understand 

> TypeError: Action.__init__() got an unexpected keyword argument 'autostart'

Completed in: a4ca018188b67fb7158bb559d6d904ac4b613c57

launch_ros/launch_ros/event_handlers/on_state_transition.py
SteveMacenski: Hey, no worries. I'm not overly familiar with this in Python3 so point taken

Fixed in 4c06f44e78c05ace536932a759ed4882aa636481

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: Actually, I have a nice cop-out for this. In the last changes, I actually removed the name argument to instead give the lifecycle node itself to handle the matcher issue. So that argument is no longer present :smile: 

I updated the outdated docstring entry for the lifecycle node object in  1ff9bacccca5b4ebe5e7493d61d6a8848073083b

launch_ros/launch_ros/descriptions/composable_node.py
SteveMacenski: >  having a composable lifecycle node type class to disambiguate it

I both agree and disagree - but I agree that having a lifecycle-component would make it absolutely clear. 

Completed in f445bd9bf0d24288daeacce02dd37176c730a802

SteveMacenski:
I appreciate it, I know this isn't the most active project in the ROS core ecosystem, so I appreciate the attention here :-) 

I think I've completed all of the tasks that need to be addressed

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: I override the `__eq__` operator so that the lifecycle node object is used for those comparisons :-) 

```
    def __eq__(self, other):
        return self.__lifecycle_node == other
```

After the changes in the other comments, I tried all of the launch examples again and they all work (including setting autostart to false, true, and removed): 

Node via `pub_sub_launch.py`:
* No autostart, gives error if one is given

Lifecycle node via `lifecycle_pub_sub_launch.py` with external triggers:
* The registered event handlers and event emiting works properly, nodes come up as expected 
* I think this was your concern, this does work. 

Lifecycle nodes via `lifecycle_autostart_pub_sub_launch.py` with internal autostart:
* Autostart = true, autostarts and pub/sub is logged. Autostart logging is seen.
* Autostart=false, system comes up, no pub/sub logging. Autostart logging is not seen.
* Autostart not set, system comes up, no pub/sub logging. Autostart logging is not seen.

Lifecycle component nodes via `lifecycle_component_autostart_pub_sub_launch.py` with internal autostart:
* Autostart=true, Autostart logging is seen.
* Autostart=false, Autostart logging is not seen.
* Autostart not set, Autostart logging is not seen.

Note for the last one `lifecycle_component_autostart_pub_sub_launch.py`, the components themselves are not lifecycle enabled, so the logging is used to validate that the autostart keyword is parsed and will setup the transition. They end up publishing / subscribing regardless of autostart since they're not actually lifecycle nodes at all. Unfortunately the [lifecycle listener / talker demos](https://github.com/ros2/demos/blob/rolling/lifecycle/src/lifecycle_listener.cpp) are not components for me to use, so I have to use the components that we have. I did test and do debug logging to make sure this is correct and I'm confident that it is.

wjwwood: > I override the __eq__ operator so that the lifecycle node object is used for those comparisons :-)

That's kind of my point, which is that the type annotations are indicating that you should expect something different. It just so happens that the matchers we use only do `==`, but a user sees it takes a `LifecycleNode` and may think they can call anything available on that class in their matcher, but that's not the case because you're passing a different class and in one case making it appear to be correct. It would be better to just pass the correct class, or have the new class actually inherit from `LifecycleNode`, or actually change the API of the callback for matchers so the false expectation is avoided.

My preference is for one of the first two options honestly. Trying to pass off this new class as if it were a `LifecycleNode` seems a bit too clever to me. It's gonna be surprising and confusing to someone at some point, in my opinion.

wjwwood:
Sorry, one more iteration on the matchers.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: I did this to avoid making other changes to matchers. I actually did this originally but backed that out because the `launch.events` matcher [1] compares the nodes rather than any specific attribute. The existing [lifecycle pub/sub launch example](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/examples/lifecycle_pub_sub_launch.py#L50) uses that matcher and would break the behavior if `==` didn't match. Since `launch.events` is pretty core, I didn't want to introduce any change that would break that behavior, even though the [`launch_ros.events` versions work perfectly for us](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/events/matchers/matches_node_name.py#L25-L29) [2]. The launch_ros version honestly seem better to me by being less fragile based on attributes rather than comparing full objects.  However anyone using the `launch.events` version rather than the `launch_ros.events` version would see a break, so I implemented the `__eq__` to make them both continue to work as expected. 

[1] https://github.com/ros2/launch/blob/rolling/launch/launch/events/matchers.py#L22-L24
[2] https://github.com/ros2/launch_ros/blob/rolling/launch_ros/launch_ros/events/matchers/matches_node_name.py#L25-L29

> It would be better to just pass the correct class, or have the new class actually inherit from LifecycleNode ... My preference is for one of the first two options honestly. 

I'm always open for critique - I don't really understand the suggestion though, can you clarify your vision for: "pass the right class"?  

I odn't think the new class is good since it would make users have to use a `LifecycleAutostartingNode` object just to obtain the `autostart` field, which is unnatural and seems confusing. Plus, it would require me to duplicate code. I'm on board with the semantic difference between `ComposableLifecycleNode` and `ComposableNode`, but having a second `LifecycleNode` for a feature that is intrinsically linked to lifecycle nodes is strange.   

> Trying to pass off this new class as if it were a LifecycleNode seems a bit too clever to me. It's gonna be surprising and confusing to someone at some point, in my opinion.

Pending on understanding your point about "pass the right class", I think what is currently here is a better solution. Its a little clever, but not really much so, I don't think. Especially if documented (which I can improve) and keeps all matchers working

SteveMacenski:
See updated comment :-)

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
wjwwood: >  I don't really understand the suggestion though, can you clarify your vision for: "pass the right class"?

You already have a new class, `LifecycleEventManager`, which you're passing on this line as `self`. My point was that you should either pass `self.__lifecycle_node` here or have `LifecycleEventManager` inherit from `LifecycleNode` since that is the type that the matcher is expecting. That way you don't need to override the `==` behavior and the matcher is getting what it expects and future changes, that might do more than `==`, will continue to work.

wjwwood: Naively, this is a concrete change that would satisfy me here:

```suggestion
        if not typed_event.lifecycle_node_matcher(self.__lifecycle_node):
```

Though that might not work for some other reason, in which case you might have to do the inheritance.

wjwwood: Well, I guess not, because if there's something in the `LifecycleEventManager` class that you need in the matcher, you could just do `lifecycle_node.__lifecycle_event_manager` (which btw there should probably be an accessor for the event manager from the lifecycle node).

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -43,6 +39,7 @@ def __init__(
         *,
         name: SomeSubstitutionsType,
         namespace: SomeSubstitutionsType,
+        autostart: Optional[bool] = False,
```
wjwwood: I think a docstring entry about this parameter here would be helpful. It's not clear to me why it's optional and what happens when it's `None`.

launch_ros/launch_ros/actions/lifecycle_node.py
```diff
@@ -145,21 +92,30 @@ def execute(self, context: launch.LaunchContext) -> Optional[List[Action]]:
...
-            matcher=lambda event: isinstance(event, ChangeState),
-            entities=[launch.actions.OpaqueFunction(function=self._on_change_state_event)],
-        ))
+
+        self.__lifecycle_event_manager = LifecycleEventManager(self)
+        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
+
+        # If autostart is enabled, transition to the 'active' state.
+        autostart_actions = None
+        if self.node_autostart:
```
wjwwood, line 106: Do you need to have a condition here for when autostart is `None`? Currently this condition is implicitly like `if self.node_autostart is not None and self.node_autostart is True`, which is maybe what you want, but also then why can it ever be `None`?

launch_ros/launch_ros/actions/node.py
```diff
@@ -362,6 +362,15 @@ def node_name(self):
...
 
+    @property
+    def node_autostart(self):
+        """Getter for autostart."""
+        return False
+
+    @property
+    def is_lifecycle_node(self):
+        return False
+
```
wjwwood: I don't think this should be added to `Node` either. It's a bad smell for lifecycle to be mentioned in the Node class at all. If we wanted to add another kind of node in the future we don't want to have to change the node class each time. Can we find a way to not need these two functions on node? This seems like a job for the type system.

launch_ros/launch_ros/descriptions/composable_node.py
```diff
@@ -17,6 +17,7 @@
 from typing import List
 from typing import Optional
 
+import launch
```
wjwwood: I don't think this is needed anymore.

launch_ros/launch_ros/descriptions/composable_node.py
```diff
@@ -29,6 +30,7 @@
 from launch_ros.parameters_type import SomeParameters
 from launch_ros.remap_rule_type import RemapRules
 from launch_ros.remap_rule_type import SomeRemapRules
+from launch_ros.utilities import LifecycleEventManager
```
wjwwood: I don't think this is needed anymore.

launch_ros/launch_ros/descriptions/composable_node.py
```diff
@@ -163,6 +165,11 @@ def node_namespace(self) -> Optional[List[Substitution]]:
         """Get node namespace as a sequence of substitutions to be performed."""
         return self.__node_namespace
 
+    @property
+    def node_autostart(self):
+        """Getter for autostart."""
+        return False
+
```
wjwwood: It would be great to avoid this addition too.

wjwwood (changes requested):
looks really close

I continued our conversation on the other thread, and pointed out a potential issue with how autostart is being typed and documented. Other than that, it's mostly cleanup requests after your most recent set of changes.

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
```diff
@@ -0,0 +1,121 @@
...
+        self, *,
+        package: SomeSubstitutionsType,
+        plugin: SomeSubstitutionsType,
+        name: Optional[SomeSubstitutionsType] = None,
+        namespace: Optional[SomeSubstitutionsType] = None,
+        parameters: Optional[SomeParameters] = None,
+        autostart: Optional[bool] = False,
+        remappings: Optional[SomeRemapRules] = None,
+        extra_arguments: Optional[SomeParameters] = None,
+        condition: Optional[Condition] = None,
```
wjwwood: You can just forward these arguments to the ComposableNode class, only adding `autostart`, instead of redeclaring them here. This is also more robust to future changes as new arguments to `ComposableNode` will automatically be accepted here too.

You can use `kwargs` to do this. Take a look at how `LifecycleNode` is done. It duplicates a few parameters that it needs to use, which is fine, but pass the rest along entirely in `kwargs`.

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
```diff
@@ -0,0 +1,121 @@
...
+    def init_lifecycle_event_manager(self, node, context: launch.LaunchContext) -> None:
+        # LifecycleEventManager needs a pre-substitution node name
+        self.__node_name = perform_substitutions(context, node.node_name)
+        self.__lifecycle_event_manager = LifecycleEventManager(node)
+        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
+
+    @property
+    def package(self) -> List[Substitution]:
+        """Get node package name as a sequence of substitutions to be performed."""
+        return self._ComposableNode__package
```
wjwwood: We should try to use accessors here. Rather than reach around the public/private name mangling.

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
```diff
@@ -0,0 +1,121 @@
...
+            name=name,
+            namespace=namespace,
+            parameters=parameters,
+            remappings=remappings,
+            extra_arguments=extra_arguments,
+            condition=condition)
+
+        self.__autostart = autostart
+        self.__lifecycle_event_manager = None
+        self.__node_name = self._ComposableNode__node_name
```
wjwwood: Same here, use an accessor. Make one if there's not one.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -230,19 +233,48 @@ def execute(
...
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_CONFIGURE]
+                    ))
+                autostart_actions.append(
+                    LifecycleTransition(
+                        lifecycle_node_names=[complete_node_name],
+                        transition_ids=[lifecycle_msgs.msg.Transition.TRANSITION_ACTIVATE]
+                    ),
+                )
+
```
wjwwood: I won't block my review on this, but I wanted to point out that adding this lifecycle specific logic here isn't ideal. Instead, it would be great to have something like `on_execute(context) -> List[Action]` on `ComposableNode`, calling that for each `node_description` and accumulating the actions returned to be later returned at the end of this function as you're doing with `autostart_actions` but generically.

Then the `on_execute` in `ComposableNode` could be empty for now, and all of this highlighted logic could be moved into `LifecycleComposableNode`'s override of `on_execute()`. That way this class never has to understand anything about lifecycle nodes.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -230,19 +233,48 @@ def execute(
...
                 load_node_requests.append(request)
 
+            # If autostart is enabled, transition to the 'active' state.
+            if node_description.node_autostart:
+                complete_node_name = request.node_namespace + request.node_name
+                if not complete_node_name.startswith('/'):
+                    complete_node_name = '/' + complete_node_name
+                self.__logger.info(
+                    'Autostart enabled for requested lifecycle node {}'.format(complete_node_name))
+                node_description.init_lifecycle_event_manager(node_description, context)
```
wjwwood: This is related to my comment about an `on_execute` "virtual" function on the composable node classes, but it's weird to be passing itself into this function. I think this could just be:

```suggestion
                node_description.init_lifecycle_event_manager(context)
```

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def init_lifecycle_event_manager(self, node, context: launch.LaunchContext) -> None:
+        # LifecycleEventManager needs a pre-substitution node name
+        self.__node_name = perform_substitutions(context, node.node_name)
+        self.__lifecycle_event_manager = LifecycleEventManager(node)
+        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
```
wjwwood: Following up on my previous comment, this could just be:

```suggestion
    def init_lifecycle_event_manager(self, context: launch.LaunchContext) -> None:
        # LifecycleEventManager needs a pre-substitution node name
        self.__lifecycle_event_manager = LifecycleEventManager(self)
        self.__lifecycle_event_manager.setup_lifecycle_manager(context)
```

Also, the `node_name` line seems completely unnecessary? Either that or the line in the constructor of this class where you do `self.__node_name = self._ComposableNode__node_name` isn't needed?

wjwwood (commented):
Sorry, after re-reviewing the new class you recently added I found some more things that should be looked into.

launch_ros/launch_ros/actions/lifecycle_node.py
SteveMacenski: Done for docstring, I remove None optionality

launch_ros/launch_ros/actions/lifecycle_node.py
SteveMacenski: I remove None optionality

launch_ros/launch_ros/actions/node.py
SteveMacenski: Done :+1:

launch_ros/launch_ros/descriptions/composable_node.py
SteveMacenski: Removed

launch_ros/launch_ros/descriptions/composable_node.py
SteveMacenski: Removed

launch_ros/launch_ros/descriptions/composable_node.py
SteveMacenski: Done :+1:

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
SteveMacenski: OK, done!

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
SteveMacenski: Totally agree :+1: Done

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
SteveMacenski: Totally agree :+1: Done

launch_ros/launch_ros/actions/load_composable_nodes.py
SteveMacenski: I don't disagree it would nicely bin the lifecycle logic in discrete places, but the composable node / lifecycle composable nodes are both descriptions not actions. I think it would be against the design intent of the object to have execution-related methods like that. 

So, I think this can be left as-is?

launch_ros/launch_ros/actions/load_composable_nodes.py
SteveMacenski: I agree, this is an artifact of moving the logic from lifecycle node to the composition node. 

Completed

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
SteveMacenski: Done - actually did this change from the previous comment before I got down to this one :laughing: 

> Also, the node_name line seems completely unnecessary? Either that or the line in the constructor of this class where you do self.__node_name = self._ComposableNode__node_name isn't needed?

It is actually necessary. The node name at the start is a substitution and we need toperform the substitution at this point to get the actual value to pass into the lifecycle event manager.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: > You already have a new class, LifecycleEventManager, which you're passing on this line as self. My point was that you should either pass self.__lifecycle_node here 

This is another obvious simplicity from heritage decisions where this started by only passing in the lifecycle node's name, not the object itself. Now that we have the object, that is exactly the right thing to do for the case of `_on_change_state_event`. I'll do that in a commit since there's absolutely no good reason not to do that. 

However, that doesn't address the issue of the `launch.event` matcher which uses the `==` operator - which can be defined externally by a user. [For example in the `lifecycle_pub_sub_launch.py` demo example](https://github.com/ros2/launch_ros/blob/rolling/launch_ros/examples/lifecycle_pub_sub_launch.py#L50). I don't actually think (?) that deriving from the lifecycle node class would solve that issue either, since the `LifecycleEventManager` would then be of type `LifecycleNode`, but not the same instance of it. 

Edit: Actually, I just found a way around that. I can just pass in the lifecycle node into the state transition too `StateTransition(action=self.__lifecycle_node, msg=msg)`. 

This is done :sunglasses: 

Personally, I think the `launch.event` matcher is extremely fragile. I feel like having the actions be required to have a `name` field which could be autogenerated as a hash if not provided would be better. 

 > which btw there should probably be an accessor for the event manager from the lifecycle node

ACKed and done

SteveMacenski:
@wjwwood all of your comments have been addressed, I believe fully now.

I see  I must have missed a DCO sign off on my commits, but I really don't want to mess with the git commit history bc I know that'll cause you more trouble re-reviewing not knowing what exactly I just changed. 

As the CEO of my own company, I trust that I can just say I give myself permissions to do this. I can also fix DCO after you review and approve if you like. I just don't want to make your life hard.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
wjwwood: > Personally, I think the launch.event matcher is extremely fragile. I feel like having the actions be required to have a name field which could be autogenerated as a hash if not provided would be better.

Not sure what you mean, but for XML having "named" instances of actions would be necessary/useful*. But what you're describing (with autogeneration or hashing), isn't that just `id(action)`? Which btw is what `==` is checking when you call it by default:

https://docs.python.org/3/library/functions.html#id

edit: * useful/necessary if you want to event handlers in XML, so you can do things like `<lifecycle_node action_name="foo" ...></> <event_handler event_type="ChangeState" target_action="foo" ...></>` where you do not have a code reference to the action object for comparison. This is something we've talked about doing for years, but haven't had the time to do. And so we don't have any kind of event handling in XML which is one of the few reasons you may have to use Python for launch files, off the top of my head.

launch_ros/launch_ros/event_handlers/on_state_transition.py
```diff
@@ -51,8 +50,19 @@ def __init__(
 
         If matcher is given, the other conditions are not considered.
         """
-        if not isinstance(target_lifecycle_node, (LifecycleNode, type(None))):
-            raise RuntimeError("OnStateTransition requires a 'LifecycleNode' action as the target")
+        lifecycle_property = type(target_lifecycle_node).__dict__.get('is_lifecycle_node', None)
```
wjwwood: Just curious, but is `getattr(target_lifecycle_node, 'is_lifecycle_node', None)` not sufficient? https://docs.python.org/3/library/functions.html#getattr

launch_ros/launch_ros/event_handlers/on_state_transition.py
```diff
@@ -51,8 +50,19 @@ def __init__(
...
+            raise RuntimeError('OnStateTransition requires a "LifecycleNode" action as the target,'
+                               ' target_lifecycle_node is not a node type.')
+
+        if (
+            target_lifecycle_node and
+            hasattr(target_lifecycle_node, 'is_lifecycle_node') and
+            not target_lifecycle_node.is_lifecycle_node
+        ):
+            raise RuntimeError('OnStateTransition requires a "LifecycleNode" action as the target,'
+                               ' target_lifecycle_node is not a lifecycle-enabled node.')
```
wjwwood: This diff is still a bit confusing to me, why is the original check not sufficient? Is it so we can support non-LifecycleNode based classes?

Also the why do we need the first new condition? It seems like it being either `None` or a `property` isn't important, but the second check definitely makes sense with `hasattr` and checking the value of `is_lifecycle_node`. Plus the two possible errors seem to be more or less the same (at least based on the error message).

wjwwood (approved):
lgtm

Had a few non-blocking questions, but I think this is good to go with passing CI.

wjwwood:
Rpr looks like it is failing due to a few small linter issues. After you fix that I'll run full CI with Windows and everything on ci.ros2.org unless someone else beats me to it.

launch_ros/launch_ros/utilities/lifecycle_event_manager.py
SteveMacenski: Huh, ok!

launch_ros/launch_ros/event_handlers/on_state_transition.py
SteveMacenski: The `__dict__` method checks the class definition if it has that attribute rather than the individual instance passed into the function. I thought this was a more strict type-checking method of handling it since someone could make an instance with that property instead.

launch_ros/launch_ros/event_handlers/on_state_transition.py
SteveMacenski: OnStateTransition is now used within the lifecycle node for the LifecycleTransition class, so we can't import `LifecycleNode` into the class or it creates circular dependencies. 

I'm going to fix these up to be more semantically meaningful. The first if is checking if we have the property at all in the class definition and the second is checking if its true. The second `hasattr(target_lifecycle_node, 'is_lifecycle_node')` is my defensive programming to make sure its there before calling it `target_lifecycle_node.is_lifecycle_node` , but I suppose at this point it shouldn't necessarily be needed. 

I think its nice to have 2 different error logging for either if the property doesn't exist (i.e. not a lifecycle node at all) and also if the value present is non-true.

SteveMacenski:
Should be good to go. I also resolved the DCO sign off

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
```diff
@@ -0,0 +1,94 @@
...
+# limitations under the License.
+
+"""Module for a description of a ComposableLifecycleNode."""
+
+from typing import List
+from typing import Optional
+
+import launch
+from launch.substitution import Substitution
+# from launch.utilities import ensure_argument_type
```
wjwwood: Looks like a left over.

wjwwood (approved):

launch_ros/launch_ros/event_handlers/on_state_transition.py
wjwwood: Oof, I thought the circular dependencies were only for the type annotations, I didn't realize we had actual circular dependencies. It's not great to have this kind of workaround for that, but maybe it's the best option atm. I would encourage you to make a comment about why we can't just use `isinstance(target_lifecycle_node, (LifecycleNode, type(None)))`, because it's not obvious why we're doing the type checking in this way.

I mean, it would be better to avoid it some other way and just check the type as we were before hand, but I don't want to hold this pr up more. I'll leave it there, but it doesn't leave me feeling awesome about the result.

One option, which I also don't love, is that you can import the class in the function, which should avoid any circular import issues, e.g.:

```python
        from ..actions import LifecycleNode
        if not isinstance(target_lifecycle_node, (LifecycleNode, type(None))):
            raise RuntimeError("OnStateTransition requires a 'LifecycleNode' action as the target")
```

It still feels like a hack, but at least it's more obvious what you're checking for than the indirect checks you have now.

wjwwood:
Looks like there's a test failure on the Rpr job, can you have a look and see if it's related/reproducible?

launch_ros/launch_ros/event_handlers/on_state_transition.py
wjwwood: You could, alternatively, import this class from the local function where `OnStateTransition` is used in `LifecycleNode`. That might make slightly more sense to me, but both are still kind of hacky.

SteveMacenski: > One option, which I also don't love, is that you can import the class in the function, which should avoid any circular import issues, e.g.:

Done. that is actually done elsewhere in the codebase, so I think that's a better option since it is at least consistent

launch_ros/launch_ros/descriptions/composable_lifecycle_node.py
SteveMacenski: Fixed. that was a copy+paste from the lifecycle node (removed from both)

SteveMacenski:
I don't think it relates to this - I just pushed a couple of new commits to fix your last 2 items - so we'll see :- )

SteveMacenski:
All passing now.

Any chance for a backport to Jazzy? This would be a really nice feature that could make some folks I'm talking to move to Jazzy if it were available

wjwwood:
CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/22559/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/16804/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2075/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/23352/)

wjwwood:
CI looks good, so I'm going to merge this now. Thanks again for the contribution!

> Any chance for a backport to Jazzy? This would be a really nice feature that could make some folks I'm talking to move to Jazzy if it were available

I think it should be fine to do so. The only changes to the API seem to be backwards compatible. Generally we don't backport features due to the risk of regression, but we can make exceptions when the risk is low. I'll ask for a second opinion on that though. I can at least open the pr for the backport and let another maintainer review/merge it as my second opinion.

wjwwood:
@mergifyio backport jazzy

mojomex:
Hi @wjwwood, is there any chance this could also be backported to Humble given that it still has two years until EOL?

emersonknapp:
@mergifyio backport humble

christophebedard:
@Mergifyio backport humble

PR #435, opened 2025-01-28
Add --remap argument to launch command

JoLichtenfeld:
This adds the option to specify remapping rules for launch files in the launch command.

Example: ros2 launch <pkg> <launch_file> --remap topic1:=new1 --remap topic2:=new2

### DIFF
--- a/ros2launch/ros2launch/command/launch.py
+++ b/ros2launch/ros2launch/command/launch.py
@@ -104,7 +104,7 @@ class LaunchCommand(CommandExtension):
         )
         parser.add_argument(
             '-r', '--remap', action='append', dest='remap_rules',
-            help=("Topics remapping rules, in the 'from:=to' form")
+            help=("Remapping rules, in the 'from:=to' form")
         )
         arg = parser.add_argument(
             'package_name',
--- a/ros2launch/ros2launch/command/launch.py
+++ b/ros2launch/ros2launch/command/launch.py
@@ -102,6 +102,10 @@ class LaunchCommand(CommandExtension):
             help=('Regex pattern for filtering which executables the --launch-prefix is applied '
                   'to by matching the executable name.')
         )
+        parser.add_argument(
+            '-r', '--remap', action='append', dest='remap_rules',
+            help=("Topics remapping rules, in the 'from:=to' form")
+        )
         arg = parser.add_argument(
             'package_name',
             help='Name of the ROS package which contains the launch file')
@@ -175,5 +179,6 @@ class LaunchCommand(CommandExtension):
                 noninteractive=args.noninteractive,
                 args=args,
                 option_extensions=self._option_extensions,
-                debug=args.debug
+                debug=args.debug,
+                remap_rules=args.remap_rules
             )
--- a/launch_ros/test/test_cli_remap.py
+++ b/launch_ros/test/test_cli_remap.py
@@ -28,20 +28,24 @@ class TestRemapArgument(unittest.TestCase):
         launch_proc_remapped = None
         try:
             # Start the talker_listener launch file with remapping
+            # Note: Using shell=True to allow sourcing the setup file.
+            # Ensure ROS_DISTRO is set in the environment where this test runs.
+            command_str = (
+                'source /opt/ros/$ROS_DISTRO/setup.bash && '
+                'ros2 launch demo_nodes_cpp talker_listener_launch.py '
+                '--remap /chatter:=/chatter_remapped'
+            )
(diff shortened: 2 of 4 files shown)


ros2launch/ros2launch/command/launch.py
```diff
@@ -102,6 +102,10 @@ def add_arguments(self, parser, cli_name):
             help=('Regex pattern for filtering which executables the --launch-prefix is applied '
                   'to by matching the executable name.')
         )
+        parser.add_argument(
+            '-r', '--remap', action='append', dest='remap_rules',
+            help=("Topics remapping rules, in the 'from:=to' form")
```
fujitatomoya: I am still learning the ROS 2 launch subsystem, but this does not seem to only applied to topics but also services and actions (https://github.com/ros2/rcl/pull/1170 merged in rolling recently).

fujitatomoya (commented):
i believe that would be really nice to have the ros2launch test aligns with this enhancement.

ros2launch/ros2launch/command/launch.py
JoLichtenfeld: That's right, services can be remapped, too, and since [ros2/rcl#1170](https://github.com/ros2/rcl/pull/1170) also topics. I will rephrase the help string.

fujitatomoya:
@JoLichtenfeld could you consider if we can add the test case for this?

JoLichtenfeld:
I will come up with a test case as soon as I find the time for it.

kramer-sim:
@fujitatomoya Hi, in which directory should the test be placed? Do you want that we write the test cases in /ros2launch/test?
Should it be similar to the test cases in [https://github.com/ros2/ros2cli/blob/f5e5a798d57c437410452f2b106f115d3a42c7a2/ros2cli/test/test_ros2cli_direct.py#L45C1-L48C34](ros2cli) or do you want to actually start a launch file on the command line and check whether the remapping works? 
You can find an example of that below:
```python
    def test_remap_argument(self):
        """Test that the --remap argument correctly remaps topics."""

        try:
            # Start the talker_listener launch file with remapping
            launch_proc_remapped = subprocess.Popen(
                ['ros2', 'launch', 'demo_nodes_cpp', 'talker_listener_launch.py', 
                 '--remap', '/chatter:=/chatter_remapped'],
                stdout=subprocess.PIPE,
                stderr=subprocess.PIPE
            )

            # Run ros2 topic list to get the remapped topics
            result = subprocess.run(['ros2', 'topic', 'list'], 
                                  stdout=subprocess.PIPE, 
                                  stderr=subprocess.PIPE, 
                                  text=True, 
                                  check=True)
            remapped_topics = result.stdout.strip().split('\n')

            # Verify /chatter is NOT in the list
            self.assertNotIn('/chatter', remapped_topics, 
                           f"Did not expect to find /chatter after remapping. Topics: {remapped_topics}")

            # Verify /chatter_remapped IS in the list
            self.assertIn('/chatter_remapped', remapped_topics, 
                        f"Expected to find /chatter_remapped after remapping. Topics: {remapped_topics}")

        finally:
            # Clean up
            if 'launch_proc_remapped' in locals():
                launch_proc_remapped.terminate()
                try:
                    launch_proc_remapped.wait(timeout=5)
                except subprocess.TimeoutExpired:
                    launch_proc_remapped.kill()
                    launch_proc_remapped.wait()


if __name__ == '__main__':
    unittest.main()
 ```

**Comment by fujitatomoya** [comment]
> do you want to actually start a launch file on the command line and check whether the remapping works?

probably this one? because it actually verifies that this arguments work with `launch`.
https://github.com/ros2/launch_ros/tree/rolling/ros2launch/test is suitable place to put the test codes for `ros2launch`.

**Comment by kramer-sim** [comment]
@fujitatomoya It appears that the test environment is missing the ROS 2 CLI extensions that provide the `topic` command. Only a limited set of commands are available, see this error message:
```
test/test_cli_remap.py:55: in test_remap_argument
    self.fail(f"'ros2 topic list' command failed with error: {result.stderr}")
E   AssertionError: 'ros2 topic list' command failed with error: usage: ros2 [-h] [--use-python-default-buffering]
E               Call `ros2 <command> -h` for more detailed usage. ...
E   ros2: error: argument Call `ros2 <command> -h` for more detailed usage.: invalid choice: 'topic' (choose from 'daemon', 'extension_points', 'extensions', 'pkg')
```
Do you have any recommendations how to proceed?

fujitatomoya:
@Mergifyio rebase

fujitatomoya:

PR #441, opened 2025-02-22
LoadComposableNodes: unload on shutdown

rayferric:
Added a new optional `unload_on_shutdown` argument to `LoadComposableNodes`.
If `True`, the action will unload the associated nodes from the container on exit.

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -237,12 +237,14 @@ class LoadComposableNodes(Action):
         :param request: service request to unload a node
         :param client: service client to call the service
         """
-        if not client.wait_for_service(timeout_sec=5.0):
-            self.__logger.warning(
-                "Abandoning wait for the '{}' service, due to timeout.".format(
+        if not client.service_is_ready():
+            self.__logger.info(
+                "Could not unload node '{}'. Is '{}' down?".format(
+                    self.__loaded_nodes[request.unique_id], self.__final_target_container_name,
                     client.srv_name
                 )
             )
+            return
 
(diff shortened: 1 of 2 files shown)


alsora:
@rayferric could you please provide some details about the use-case for this?
Shouldn't nodes always be unloaded when we shutdown? Why do we need a parameter for this?

rayferric:
> @rayferric could you please provide some details about the use-case for this?
> Shouldn't nodes always be unloaded when we shutdown? Why do we need a parameter for this?

Of course!

### The usecase:

This feature is useful when you split the development workflow into two launch files - one for simulation/rviz and another for the navigation stack. In such setup you can keep restarting the stack quickly without having to wait for gui programs to initialize.

Then if you'd want to have a high end multi-camera robot in the simulation, the obvious choice is to compose the simulator with your image processing pipeline. So in order to keep your simulation in a separate launch file, you'd start the container there and the simulator would be a ComposableNode. Then you could load the rest of your image processing stack from the second launch file using LoadComposableNodes.

The problem with vanilla LoadComposableNodes is that it exits immediately and doesn't allow you to unload the nodes in any way.

My PR adds an optional switch to the action that will make it keep the launch process alive and once the user sends a SIGINT, the action will unload all previously inserted nodes.

### Why not the default?

I wanted to keep this feature non-invasive, thus I created an optional flag for it.

alsora:
Ok, thanks for the explanation.
I have to admit that I haven't looked at the implementation yet and I had misunderstood the purpose.

This feature absolutely needs to be optional (i.e. controlled by a flag), as keeping the launch process alive is a major change.

I'll do a review in the next days, thanks for the context.

PR #470, opened 2025-06-17
use locals instead of launch configurations to store ros_remaps and ros_namespace

ijnek:
## Description
ros_remaps and ros_namespace are currently stored as launch configurations, but they don't behave like typical launch configurations — they're not declared using DeclareLaunchArgument, and their purpose is more aligned with contextual scoping.

They are better suited to be stored in the local stack (locals()) of the LaunchContext, rather than in the global launch configurations. This distinction aligns better with their use as scoped contextual variables rather than user-defined launch arguments.

This change is related to the discussion in [ros2/launch#801 (comment)](https://github.com/ros2/launch/issues/801#issuecomment-2975646031), where a workaround was proposed to distinguish between launch configurations that should be scoped versus those that should propagate globally.

This PR is marked as a draft because I haven't found definitive guidance in launch_ros or [ros2/design](https://github.com/ros2/design) regarding the intended semantics of the locals and globals stacks within LaunchContext. Based on my understanding, this usage seems appropriate, but I welcome feedback if this assumption is incorrect.

No

### DIFF
--- a/launch_ros/launch_ros/actions/node.py
+++ b/launch_ros/launch_ros/actions/node.py
@@ -400,7 +400,7 @@ class Node(ExecuteProcess):
             if self.__node_namespace is not None:
                 expanded_node_namespace = perform_substitutions(
                     context, normalize_to_list_of_substitutions(self.__node_namespace))
-            base_ns = context.launch_configurations.get('ros_namespace', None)
+            base_ns = context.get_locals_as_dict().get('ros_namespace', None)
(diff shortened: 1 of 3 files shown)


SuperJappie08:
As mentioned in [ros2/launch#801 (Comment)](https://github.com/ros2/launch/issues/801#issuecomment-2979461601), this does (currently) not work with normal group scoping.

SuperJappie08:
This should probably also effect `SetParameter` and `SetParametersFromFile`.

PR #516, opened 2026-02-12
Load composable nodes safely by optionally shutting down the launch process if it fails to load

### LINKED ISSUE #513: LoadComposableNodes should have an option to fail/shutdown if a node fails to load
Currently, when a `LoadComposableNodes` launch action is executed, it calls the `load_node` service of a container. If the service `returns success=False` (e.g., due to a missing plugin, invalid parameters, or a constructor failure), the action logs an error but the launch continues:

https://github.com/ros2/launch_ros/blob/80e61471e3776426e87e733a98ba914155f87011/launch_ros/launch_ros/actions/load_composable_nodes.py#L179-L185

In many production cases, if a component fails to load, the entire system is in an invalid state. I would like to propose adding an optional parameter (e.g., `on_failure_shutdown=True`) to `LoadComposableNodes` that raises an exception or emits a `Shutdown` event if any node fails to load.

This would allow users to "fail fast" and avoid having a partially loaded system running unnoticed.

xaru8145:
## Description

This PR introduces an optional `on_failure_shutdown` parameter to the `LoadComposableNodes` action. Currently, when a component fails to load (e.g., missing plugin, invalid parameters), the failure is logged but the launch continues. This change allows users to fail fast by raising a `RuntimeError`, triggering a system shutdown if a required node fails to load.

The `on_failure_shutdown` parameter defaults to **False**, strictly mimicking the previous behavior to ensure no breaking changes for existing launch files.

Changes:

- Added `on_failure_shutdown` to the `LoadComposableNodes` constructor. The default value of **False** ensures exact parity with the legacy loading logic.

- Updated `_load_node` to raise a `RuntimeError` if the service response success is False and `on_failure_shutdown` is enabled.

- Raising a `RuntimeError` allows the LaunchService to catch the exception in its main loop, set the return code to 1, and initiate an orderly shutdown.

- Updated specific tests and frontend tests for parsing the new parameter.

Fixes https://github.com/ros2/launch_ros/issues/513

Yes. Users can now optionally specify `on_failure_shutdown="true"` (XML/YAML) or `on_failure_shutdown=True` (Python) within a `LoadComposableNodes` action.

- Previous behavior: If a component failed to load into a container (e.g., due to a missing plugin or invalid parameters), the error was logged to the console, but the launch process would continue.

- New behavior: If the flag is set to True, a load failure will now trigger an immediate RuntimeError, halting the loading of any subsequent nodes in that action and initiating a system-wide shutdown with a non-zero exit code.

Yes, Gemini, for generating the tests.

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -65,6 +65,7 @@ class LoadComposableNodes(Action):
         *,
         composable_node_descriptions: List[ComposableNode],
         target_container: Union[SomeSubstitutionsType, ComposableNodeContainer],
+        on_failure_shutdown: bool = False,
         **kwargs,
     ) -> None:
         """
@@ -79,6 +80,7 @@ class LoadComposableNodes(Action):
 
         :param composable_node_descriptions: descriptions of composable nodes to be loaded
         :param target_container: the container to load the nodes into
+        :param on_failure_shutdown: if True, the launch will shutdown if a node fails to load
         """
         ensure_argument_type(
             target_container,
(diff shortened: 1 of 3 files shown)


kscottz:
Sick! Thank you. You just missed the cutoff for our weekly task assignment but we'll try and get someone to review shortly.

xaru8145:
Any updates on this? @kscottz @wjwwood

Issue #517, opened 2026-02-13
Feature: Combined Parameter Substitution


DLu:
### Description

I am proposing a new `launch_ros.Substitution` that takes multiple yaml paths as a parameter, and returns one unified (temporary) yaml path as the result.

### Motivation

YAML files are getting mighty unwieldy these days. [Check out all 600 lines of this navigation config](https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/params/nav2_params.yaml). It is easy enough to load multiple parameter files into a Node when you are writing the launch file directly, i.e. 

```python
    Node(...
        parameters=[yaml_path1, yaml_path2],
    )
```

However, there are scenarios where you might want to load those yaml files into one single `LaunchConfiguration`. Consider [`nav2_bringup / navigation_launch.py`](https://github.com/ros-navigation/navigation2/blob/main/nav2_bringup/launch/navigation_launch.py). The above navigation configuration is declared as a launch argument

```python
declare_params_file_cmd = DeclareLaunchArgument(
        'params_file',
        default_value=os.path.join(bringup_dir, 'params', 'nav2_params.yaml'),
        description='Full path to the ROS2 parameters file to use for all launched nodes',
    )
```
and then passed to multiple nodes (after some processing). 

I want to be able to specify multiple yaml files to replace the single `nav2_params.yaml` WITHOUT having to rewrite the entire launch file. 

### Design / Implementation Considerations

One option is to just copy/paste the launch file and rewrite the parameters as needed, which is valid, and [many people have done just that](https://github.com/search?q=path%3A**%2Fnavigation_launch.py+language%3APython+%222018+intel+corporation%22&type=code).

I have implemented a proof-of-concept approach (that I call `MultiYaml`) wherein multiple yaml files are passed in, and a new temporary yaml file is returned as the result of the substitution. The contents of the new temporary yaml file are merged into one large dictionary.

[`nav2_common / RewrittenYaml`](https://github.com/ros-navigation/navigation2/blob/064ca7585523f5205328eb186f43b65a52de428d/nav2_common/nav2_common/launch/rewritten_yaml.py#L118-L120) and [`nav2_common / ReplaceString`](https://github.com/ros-navigation/navigation2/blob/064ca7585523f5205328eb186f43b65a52de428d/nav2_common/nav2_common/launch/replace_string.py#L68-L78) already read a yaml file and write a temporary yaml file as a result. #258 exists to port that behavior to `launch_ros`. 

[...]

PR #518, opened 2026-02-17
Porting RewrittenYaml from Nav2

### LINKED ISSUE #258: consider porting `RewrittenYaml` from Nav2 to allow param overrides to YAML from launch
## Feature request
Consider porting `RewrittenYaml` [from Nav2](https://github.com/ros-planning/navigation2/pull/771/files#diff-ddc85dea3844fd855756b5c6040387f0c9bb0739ddf54eab519191d7021b9aefR19).

#### Feature description

The use case is using YAMLs for default params but allowing to override individual params from launch. The overwrites could be extended to an override yaml. 

```
launch_ros.actions.Node(
        package='nodes',
        executable='node_exe',
        parameters=[
            path_to_yaml,
            # overwrite parameters from yaml here
            {"param_name": value}
        ]
```
This seems to almost work but with the limitation that the params specified in the YAML must be for all nodes. This is not ideal in the event params have the same name between nodes and you wish to distinguish. 

[This](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/blob/master/src/tools/joystick_vehicle_interface_nodes/launch/joystick_vehicle_interface_node.launch.py#L64) is a pretty nice way to override individual parameters in a YAML during launch. However, it only seems to work if the YAML file is set to [for all nodes](https://gitlab.com/autowarefoundation/autoware.auto/AutowareAuto/-/blob/master/src/tools/joystick_vehicle_interface_nodes/param/logitech_f310.default.param.yaml#L1).

An alternative is Nav2's `RewrittenYaml`[[link](https://github.com/ros-planning/navigation2/pull/771/files#diff-ddc85dea3844fd855756b5c6040387f0c9bb0739ddf54eab519191d7021b9aefR19)] feature but that requires to clone/install all of Nav2.

DLu:
## Description

Straight port of RewrittenYaml from Nav2

Fixes #258 

Yes, adds a new substitution. 

No

### Additional Information

This is the exact implementation and test code from Nav2, with the exception of modifying the imports in the second two commits. No other improvements were attempted.

### DIFF
--- a/launch_ros/launch_ros/substitutions/__init__.py
+++ b/launch_ros/launch_ros/substitutions/__init__.py
@@ -19,6 +19,7 @@ from .find_package import FindPackage
 from .find_package import FindPackagePrefix
 from .find_package import FindPackageShare
 from .parameter import Parameter
+from .rewritten_yaml import RewrittenYaml
 
 
 __all__ = [
@@ -27,4 +28,5 @@ __all__ = [
     'FindPackagePrefix',
     'FindPackageShare',
     'Parameter',
+    'RewrittenYaml',
 ]
--- a/launch_ros/launch_ros/substitutions/rewritten_yaml.py
+++ b/launch_ros/launch_ros/substitutions/rewritten_yaml.py
@@ -1,0 +1,274 @@
+# Copyright (c) 2019 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from collections.abc import Generator
+import tempfile
+from typing import Optional, TypeAlias, Union
+
+import launch
+import yaml
+
+YamlValue: TypeAlias = Union[str, int, float, bool]
+
+
+class DictItemReference:
+
+    def __init__(self, dictionary: dict[str, YamlValue], key: str):
+        self.dictionary = dictionary
+        self.dictKey = key
+
+    def key(self) -> str:
+        return self.dictKey
+
+    def setValue(self, value: YamlValue) -> None:
+        self.dictionary[self.dictKey] = value
+
+
+class RewrittenYaml(launch.Substitution):
+    """
+    Substitution that modifies the given YAML file.
+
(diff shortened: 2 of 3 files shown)


DLu:
TBH: I feel like the multiple types of rewrites available here chafes a bit against the ["Distinct substitutions that can be composed make the most sense to me." that @mjcarroll articulated here](https://github.com/ros2/launch_ros/issues/517#issuecomment-3902027111), but I also haven't written a lot of uses for `RewrittenYaml`

izzaaaatanishq:
Related: https://github.com/ros-teleop/twist_mux/issues/65

emersonknapp:
This is sometimes useful, for sure. But I've noticed in the wild that a lot of uses of it aren't actually necessary, resulting in a no-op, and it's only there because the user copy-pasted a Nav2 python launchfile as their starting point.

We could consider having it raise a Warning when the result is a no-op just to make it clear to users that they've cargo-culted something unnecessary.

Also we should try _really_ hard to make sure that all Actions and Substitutions are exposed to launch frontend (so they can be used in YAML/XML). I had started on this one before parental leave but don't think I finished it. I think the interface is complex enough that the `$(rewritten-yaml ARGS ARGS ARGS)` gets complex enough it probably warrants some new patterns or architecture to support high-complexity substitutions. OR, we may be better off taking a closer look at the exact use cases this is smoothing over in the majority of uses and see how the core tooling can support that better.

Nothing I've said is a blocker, but food for thought.

peci1:
Any particular reason why this goes to launch_ros and not to launch? I don't see anything ROS-specific here...

mjcarroll:
@DLu when you get a moment, do you have any thoughts on these two comments?

DLu:
> This is sometimes useful, for sure. But I've noticed in the wild that a lot of uses of it aren't actually necessary, resulting in a no-op, and it's only there because the user copy-pasted a Nav2 python launchfile as their starting point.

This is a bigger question, and maybe something that people with more Nav2 experience than I should address.  I fall into the category of people who mostly copy-paste it. 

> Also we should try _really_ hard to make sure that all Actions and Substitutions are exposed to launch frontend (so they can be used in YAML/XML). 

I concur, but was starting by just porting what exists in the Nav2 repo. 

> I think the interface is complex enough that the `$(rewritten-yaml ARGS ARGS ARGS)` gets complex enough it probably warrants some new patterns or architecture to support high-complexity substitutions

That's basically what I was getting at with the "Yaml Operations" patterns I discussed in my original post here: #517 

> Any particular reason why this goes to launch_ros and not to launch? I don't see anything ROS-specific here...

A) The original ticket was here #258 in this repo. 
B) `ros2/launch/launch` has a fairly sparing use of `yaml` (except in the for-loops(?!)), whereas there's a lot here. 
C) All the parameter descriptions and utilities are in this repo. 

I'm happy to change it as needed.

PR #530, opened 2026-04-01
[rolling] Combined Parameter Files

### LINKED ISSUE #517: Feature: Combined Parameter Substitution

DLu:
## Description

Fixes #517 in Rolling - Combined parameter files. 

Yes, adds a new way to combine parameter files in launch files. 

No.

### Additional Information

After getting a little annoyed at #518, I decided to try another approach I mentioned in #517 where we add a new parameter type to `parameter_descriptions`. 

Current State:
 * ~I developed this in jazzy. It should eventually go in rolling but I wanted to get feedback before I committed to it.~ Now its rolling
 * New description: `CombinedParameterFiles` allows for combining multiple yaml files and writing into a temp yaml file.
   * **Issue**: I'm not sure if I have the typing correct on the `param_files` argument
 * `TempYamlFile` class consolidates the functionality for maybe writing to a temp yaml file between this and `ParameterFile`
 * Updated the typing in `parameters_type.py` and added `CombinedParameterFiles` as a valid type in a couple of spots. 
 * Added new parsing logic in `Node.parse`
 * Tests for core Python functionality in `test_combined_param_files.py`
 * Tests for parsing Yaml and XML in `test_node_with_combo_file_params.py`
   * ~**Issue**: I got the XML to work the way I wanted, but didn't get the YAML to parse the way I wanted. That test currently fails.~ Figured it out with @christophebedard

### DIFF
--- a/test_launch_ros/test/test_launch_ros/frontend/test_node_with_combo_file_params.py
+++ b/test_launch_ros/test/test_launch_ros/frontend/test_node_with_combo_file_params.py
@@ -58,10 +58,9 @@ def test_launch_file_params_yaml():
                 name: my_talker
                 args: '--number_of_cycles 1'
                 param:
-                    - file:
-                      - path: {}
-                    - file:
-                      - path: {}
+                - file:
+                    - path: {}
+                    - path: {}
         """.format(yaml_params1, yaml_params2))  # noqa: E501
 
     with io.StringIO(yaml_file) as f:
--- a/test_launch_ros/test/test_launch_ros/frontend/test_node_with_combo_file_params.py
+++ b/test_launch_ros/test/test_launch_ros/frontend/test_node_with_combo_file_params.py
@@ -1,0 +1,89 @@
+# Copyright 2019 Open Source Robotics Foundation, Inc.
+# Copyright 2020 Open Avatar Inc.
+# Copyright 2026 Metro Robots
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import io
+import pathlib
+import textwrap
+
+from launch import LaunchService
+from launch.frontend import Parser
+
+from launch_ros.utilities import evaluate_parameters
+
+import yaml
+
+parent_folder = pathlib.Path(__file__).parent
+# Escape backslashes if any to keep them after parsing takes place
+yaml_params1 = str(parent_folder / 'params.yaml').replace('\\', '\\\\')
+yaml_params2 = str(parent_folder / 'params2.yaml').replace('\\', '\\\\')
+
+
+def test_launch_file_params_xml():
+    xml_file = textwrap.dedent(
+        r"""
+        <launch>
+            <node pkg="demo_nodes_py" exec="talker_qos" name="my_talker" args="--number_of_cycles 1">
+                <param>
+                  <file path="{}"/>
+                  <file path="{}"/>
+                </param>
+            </node>
+        </launch>
+        """.format(yaml_params1, yaml_params2))  # noqa: E501
+
+    with io.StringIO(xml_file) as f:
+        check_launch_node(f)
+
+
+def test_launch_file_params_yaml():
+    yaml_file = textwrap.dedent(
+        r"""
+        launch:
+            - node:
+                pkg: demo_nodes_py
+                exec: talker_qos
+                name: my_talker
+                args: '--number_of_cycles 1'
+                param:
+                    - file:
+                      - path: {}
+                    - file:
+                      - path: {}
+        """.format(yaml_params1, yaml_params2))  # noqa: E501
+
+    with io.StringIO(yaml_file) as f:
+        check_launch_node(f)
+
+
+def check_launch_node(file):
+    root_entity, parser = Parser.load(file)
+    ld = parser.parse_description(root_entity)
+    ls = LaunchService()
+    ls.include_launch_description(ld)
+    assert 0 == ls.run()
+    evaluated_parameters = evaluate_parameters(
+        ls.context,
+        ld.describe_sub_entities()[0]._Node__parameters
+    )
+    assert len(evaluated_parameters) == 1
+    assert isinstance(evaluated_parameters[0], pathlib.Path)
+
+    param_path = evaluated_parameters[0]
+    file_params = yaml.safe_load(open(param_path))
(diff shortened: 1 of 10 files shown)


DLu:
I have updated this PR to be based on rolling, and removed the draft status. 

@christophebedard @emersonknapp I would welcome your feedback on the proper YAML format to mimic the XML test.

christophebedard:
Taking a look!

test_launch_ros/test/test_launch_ros/frontend/test_node_with_combo_file_params.py
```diff
@@ -0,0 +1,89 @@
...
+            - node:
+                pkg: demo_nodes_py
+                exec: talker_qos
+                name: my_talker
+                args: '--number_of_cycles 1'
+                param:
+                    - file:
+                      - path: {}
+                    - file:
+                      - path: {}
```
christophebedard: This works:

```suggestion
                param:
                    - file:
                        - path: {}
                        - path: {}
```

The key is that `param.get_attr('file', data_type=List[Entity], optional=True)` is expected to be:

1. In XML: a list of (i.e., _multiple_)`<file .../>` nested within `<param>...</param>`.
2. In YAML: a _single_ `file` with a list of dictionaries (`- ...`) under it.

For reference, look at how XML `<let>` and YAML `let:` are handled in `IncludeLaunchDescription`: https://github.com/ros2/launch/blob/a5a6dd7eb7865dbc64d0e18f0229b0ebad826082/launch/launch/actions/include_launch_description.py#L96-L116. I'm not entirely sure why it's

```yaml
        launch:
            - node:
                # ...
                param:
                    - file:
                        - path: {}
                        - path: {}
```

and not

```yaml
        launch:
            - node:
                # ...
                param:
                    file:
                        - path: {}
                        - path: {}
```

like `param:` is to `node:` here or how `let:` is to `include:` in `IncludeLaunchDescription`, but the parsing works 🤷 . Maybe because we have an extra nesting level here (all under `param` and not directly under `node`).

christophebedard: > Maybe because we have an extra nesting level here (all under param and not directly under node).

ah, that is probably because where you put your new parsing code is supposed to handle a single param out of a list of param. YAML `param:` expects a list under it (and we can have multiple XML `<param>` under a `<node>`), hence the `- file:`. In other words, you need `- file:` here because you can have multiple params :exploding_head: totally not confusing

PR #531, opened 2026-04-09
Add per composable node log_level support

penww:
## Description

Fixes # (https://github.com/ros2/rclcpp/issues/3091)

No, this change add new parameter for `ComposableNode`, not affect the existing functions.

Yes, partially with Claude Sonnet 4.6.

### Additional Information

Example for per component log level feature enabled:

```python
    return LaunchDescription([ComposableNodeContainer(
        name='component_demo_container',
        namespace='',
        package='rclcpp_components',
            ComposableNode(
                package='<PKG>',
                plugin='<PLUGIN>',
                log_level='DEBUG',  # component log level
            ),
        ],
        arguments=['--ros-args', '--log-level', 'INFO'],  # container log level
    )])
```

Related PR: https://github.com/ros2/rclcpp/pull/3092, https://github.com/ros2/ros2cli/pull/1212

### DIFF
--- a/launch_ros/launch_ros/actions/load_composable_nodes.py
+++ b/launch_ros/launch_ros/actions/load_composable_nodes.py
@@ -40,6 +40,7 @@ from launch.utilities import type_utils
 from launch_ros.parameter_descriptions import ParameterFile
 
 import lifecycle_msgs.msg
+import rclpy.logging
 
 from .composable_node_container import ComposableNodeContainer
 from .lifecycle_transition import LifecycleTransition
@@ -321,7 +322,6 @@ def get_composable_node_load_request(
     combined_ns = make_namespace_absolute(prefix_namespace(base_ns, expanded_ns))
     if combined_ns is not None:
         request.node_namespace = combined_ns
-    # request.log_level = perform_substitutions(context, node_description.log_level)
     remappings = []
     global_remaps = context.launch_configurations.get('ros_remaps', None)
     if global_remaps:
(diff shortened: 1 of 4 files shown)


launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -366,4 +366,18 @@ def get_composable_node_load_request(
...
+            context, composable_node_description.log_level).upper()
+        log_level_map = {
+            'DEBUG': rcl_interfaces.msg.Log.DEBUG,
+            'INFO': rcl_interfaces.msg.Log.INFO,
+            'WARN': rcl_interfaces.msg.Log.WARN,
+            'WARNING': rcl_interfaces.msg.Log.WARN,
+            'ERROR': rcl_interfaces.msg.Log.ERROR,
+            'FATAL': rcl_interfaces.msg.Log.FATAL,
+        }
+        request.log_level = log_level_map.get(log_level_str, 0)
```
fujitatomoya: this fallback is NOT great for the user application. if user sets typos `log_level='debig'`, it silently sets it to UNSET and the user will not be notified configuration gone wrong. i think that this behavior is actually worse then raising the exception, what do you think? i say, we should enforce the check and return error would be better. and again this should be done with rclpy `get_logging_severity_from_string`?

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -366,4 +366,18 @@ def get_composable_node_load_request(
                 )
             )
         ]
+
+    if composable_node_description.log_level is not None:
+        log_level_str = perform_substitutions(
+            context, composable_node_description.log_level).upper()
+        log_level_map = {
+            'DEBUG': rcl_interfaces.msg.Log.DEBUG,
```
fujitatomoya: should be rcl_interfaces.msg.LoggerLevel.

launch_ros/launch_ros/actions/load_composable_nodes.py
```diff
@@ -366,4 +366,18 @@ def get_composable_node_load_request(
...
+            'INFO': rcl_interfaces.msg.Log.INFO,
+            'WARN': rcl_interfaces.msg.Log.WARN,
+            'WARNING': rcl_interfaces.msg.Log.WARN,
```
fujitatomoya: docstring does not mention about this string though... we should add it to the help docstring?

fujitatomoya (changes requested):
test would be ideal too.

penww:
Hi @fujitatomoya, thank you for the review. I've fixed the reported issues and updated the PR accordingly. I'd appreciate it if you could take another look.

Issue #539, opened 2026-04-29
:farmer: `test_launch_ros` generating build regressions in nightly_win by a windows access violation


miguelgonrod:
### Operating System:

Windows 11

### ROS version or commit hash:

rolling

### Steps to reproduce issue

1. Go to https://ci.ros2.org/job/nightly_win_rel or https://ci.ros2.org/job/nightly_win_rep
2. Build a Job
3. See the job failing

### Expected behavior

To build correctly

### Actual behavior

Failing build

### Reference Build
- [nightly_win_rel#3780](https://ci.ros2.org/job/nightly_win_rel/3780)
- [nightly_win_rep#4087](https://ci.ros2.org/job/nightly_win_rep/4087)

### Log Output
```
--- stderr: test_launch_ros
Windows fatal exception: access violation

Thread 0x00002768 (most recent call first):
  File "C:\ci\ws\install\Lib\site-packages\rclpy\executors.py", line 872 in _wait_for_ready_callbacks
  File "C:\ci\ws\install\Lib\site-packages\rclpy\executors.py", line 970 in wait_for_ready_callbacks
  File "C:\ci\ws\install\Lib\site-packages\rclpy\executors.py", line 1000 in _spin_once_impl
  File "C:\ci\ws\install\Lib\site-packages\rclpy\executors.py", line 1020 in spin_once
  File "C:\ci\ws\install\Lib\site-packages\launch_ros\ros_adapters.py", line 77 in _run
  File "C:\pixi_ws\.pixi\envs\default\Lib\threading.py", line 1010 in run
  File "C:\pixi_ws\.pixi\envs\default\Lib\threading.py", line 1073 in _bootstrap_inner
  File "C:\pixi_ws\.pixi\envs\default\Lib\threading.py", line 1030 in _bootstrap

Thread 0x0000247c (most recent call first):
  File "C:\pixi_ws\.pixi\envs\default\Lib\selectors.py", line 314 in _select
  File "C:\pixi_ws\.pixi\envs\default\Lib\selectors.py", line 323 in select
  File "C:\pixi_ws\.pixi\envs\default\Lib\asyncio\base_events.py", line 1949 in _run_once
  File "C:\pixi_ws\.pixi\envs\default\Lib\asyncio\base_events.py", line 641 in run_forever
  File "C:\ci\ws\install\Lib\site-packages\launch\utilities\signal_management.py", line 149 in run_background_loop
[...]

PR #562, opened 2026-08-06
Disable Fast DDS shared memory transport on Windows

mjcarroll:
Opening a shared memory segment left inconsistent by an abruptly terminated peer faults inside `SharedMemGlobal::open_port_internal`, killing the test process with an access violation (0xC0000005).

`check_sanity()` walks the allocator red-black tree through pointers stored in the segment. The fault is not a C++ exception, so the surrounding `catch (std::exception&)` does not intercept it.

Workaround for #539; root cause is upstream in Fast DDS.

Yes I used AI, and it hates using the pull request template

### DIFF
--- a/test_launch_ros/test/conftest.py
+++ b/test_launch_ros/test/conftest.py
@@ -1,0 +1,24 @@
+# Copyright 2026 Open Source Robotics Foundation, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


mjcarroll:
Windows: [Build Status](https://ci.ros2.org/job/ci_windows/28929/)

knmcguire:
Just noting here that someone also made a PR with a similar fix but for system_test (which I think was also hitting the same issues)

https://github.com/ros2/system_tests/pull/605

mjcarroll:
I have filed an issue upstream (and will have a PR shortly) to make shared memory handling a little safer in this scenario: https://github.com/eProsima/Fast-DDS/issues/6501

